Issue #15, opened 2018-04-15
Naming and Layout/Style Proposal


meleu:
_From @SyrianBallaS on April 3, 2018 19:25_

<h1>Preface</h1>
<p>
  OK, so I noticed the .editorconfig file got accepted. That will definitely help a little bit, but ultimately not 
  solve the root issue of inconsistent style. It is understandable that inconsistencies are undesirable and wish 
  to propose a solution. Some of these come from the reviews in pull request #49, others from myself, and, isocpp.
  For some insight, I professionaly program in ASP.NET Core so some stylings may differ from isocpp.
</p>
<p>
	On another note (somewhat related), I propose using a package.config/project.json file for binaries to prevent
	changes to library files (plus it will save space in the repo). Had one a long time ago but deleted the repo that was in it because
	it was too far gone (incorrect).
</p>
<p>The lead engineer (I'm assuming @ScottFromDerby) will have the final say.</p>
<p>This is just a proposal, what's actually going to be used should be formalized in the wiki or a CONTRIBUTING file. This is just to give some ideas.</p>

<h2>Naming</h2>
<p>For a lot of these, you can easily change them using <code>ctrl+alt+f</code>.</p>

<h3>Data Members</h3>
<p>
	I would prefer data members be in snake-case with a leading underscore (i.e., <code>some_var_</code>) as encoding type information is redundant.
	However, if the majority strongly desire Hungarian Notation that's fine too.
</p>

<h3>Methods</h3>
<ul>
	<li>For this one I'm not too sure on what we should do but I'll propose something any way.</li>
</ul>

<h4>Event Handlers</h4>
<ul>
	<li>Given my background, I propose we use Pascal case for these.</li>
	<li>An event handler for us is like the Message Handlers in my fork.</li>
</ul>
<h4>Other Member functions, internal functions(helpers), and global functions</h4>
<ul>
	<li>These should be in snake-case</li>
</ul>

<h3>Classes/Structs</h3>
<ul>
	<li>These should be in snake-case, we could make the first letter capital.</li>
	<li>In my fork I wrapped all the new stuff in a namespace (ra) but put a macro <code>_RA </code> to make it more consistent. It's defined like this: <code>#define _RA ::ra::</code></li>
</ul>

<h3>Enums</h3>
<ul>
	<li>Should be changed to "enum class", this prevents conflicts and disables implicit conversion to int.</li>
	<li>snake-case</li>
[...]

PR #23, opened 2018-04-22
Safer string conversions and other helper functions.

SyrianBallaS:
Existing code had to be slightly changed in some parts in the files besides RA_Defs to resolve some errors.

The type_traits do indeed work, but demonstrating them in RA_Defs caused a circular dependency so they're commented out.


src/RA_Defs.cpp
```diff
@@ -6,57 +6,71 @@
...
-	return reinterpret_cast<char*>(stream.data());
-}
+inline namespace conversions{
+std::string DataStreamAsString(const DataStream& stream) {
 
-std::string Narrow(const wchar_t* wstr)
-{
-	static std::wstring_convert< std::codecvt_utf8_utf16< wchar_t >, wchar_t > converter;
-	return converter.to_bytes(wstr);
+    auto str{ convert_string<std::string>(stream) };
```
Jamiras: Make sure these conversions work with UTF-16 surrogates.  many of our rich presence scripts use unicode emojis:
http://docs.retroachievements.org/Rich-Presence/#unicode-standard-symbols

src/RA_Defs.cpp
```diff
@@ -6,57 +6,71 @@
...
-}
 
-std::string Narrow(const std::string& str)
-{
-	return str;
+
+    std::ostringstream oss;
+    oss.flags(std::ios::hex | std::ios::basefield | std::ios::showbase);
+    oss << nAddr;
+    return oss.str();
```
Jamiras: Will this put leading zeros in front of the value? i.e. `1` => `0x000001`?

src/RA_Defs.h
```diff
@@ -11,26 +14,67 @@
 #include <queue>
 #include <deque>
 #include <map>
+//#include <memory>
+//#include <thread>
+//#include <mutex>
```
Jamiras: Commented code should not be committed.

src/RA_Defs.h
```diff
@@ -11,26 +14,67 @@
...
 
 #ifndef RA_EXPORTS
 
 //	Version Information is integrated into tags
 
 #else
+// if you really want to use null instead of nullptr like .net
+#ifndef null
+#define null nullptr
+#endif // !null
```
Jamiras: We're not trying to be .NET. c++ has `NULL` and `nullptr`. We use those.

src/RA_Dlg_GameTitle.cpp
```diff
@@ -113,7 +113,7 @@ INT_PTR Dlg_GameTitle::GameTitleProc( HWND hDlg, UINT uMsg, WPARAM wParam, LPARA
 			{
 				//	Existing title
 				ASSERT( m_aGameTitles.find( std::string( sSelectedTitle ) ) != m_aGameTitles.end() );
-				nGameID = m_aGameTitles[ std::string( Narrow(sSelectedTitle) ) ];
+				nGameID = m_aGameTitles[ std::string(_RA Narrow(sSelectedTitle) ) ];
```
Jamiras: I don't like hiding the namespace behind a `#define`. Suggest explicit `ra::Narrow`

src/RA_Defs.cpp
SyrianBallaS: yes

src/RA_Defs.cpp
SyrianBallaS: utf16 is either wchar_t or char16_t, it will work.

src/RA_Defs.h
SyrianBallaS: ok

src/RA_Dlg_GameTitle.cpp
SyrianBallaS: ok

SyrianBallaS:
If you want some closure, I'll make a video later proving this stuff works.

src/RA_Dlg_GameTitle.cpp
SyrianBallaS: it's done

src/RA_Defs.h
SyrianBallaS: it's done

src/RA_Defs.cpp
SyrianBallaS: https://youtu.be/uT9Gz27NA3E

src/RA_Defs.cpp
SyrianBallaS: I'm not really sure about the narrow/widen anymore, thanks for bringing it up. If I can't figure it I'll put narrow/widen how they were and put more restrictions on convert_string.

SyrianBallaS:
Everything stated before has been resolved.

src/RA_Defs.cpp
SyrianBallaS: Those emojis appear now. They won't look exactly how they are on the docs unless you open the text file in a browser or something. But they appear the same using the release version of RA and from my code changes.

src/RA_Defs.h
```diff
@@ -12,28 +14,63 @@
...
+#define _DEPRECATED          [[deprecated]]
+#define _DEPRECATEDR(reason) [[deprecated(reason)]]
+#define _FALLTHROUGH         [[fallthrough]]//; you need ';' at the end
+#define _NODISCARD           [[nodiscard]]
+#define _NORETURN            [[noreturn]]
+#define _UNUSED              [[maybe_unused]]
+
+// Disables the use of const_casts, if you get an error, it's not a literal
+// type. You could use it on functions but they will need a deduction guide
+// That would probably be better with a forwarding function
```
Jamiras: comment doesn't seem to apply to anything

src/RA_Defs.h
```diff
@@ -12,28 +14,63 @@
...
 
 
+using namespace std::string_literals;
+//using namespace std::chrono_literals; we could use this later
+namespace ra {}
+using namespace ra;
+
+#endif	//RA_EXPORTS
+#define _CONSTANT_VAR inline constexpr auto
+#define _CONSTANT     inline constexpr
```
Jamiras: more seemingly unnecessary #defines

weren't you arguing against macros recently?

src/RA_Defs.h
```diff
@@ -62,160 +99,153 @@ extern GetParseErrorFunc GetJSONParseErrorStr;
...
+
+
+
+
+
+extern void RADebugLogNoFormat(const char* data);
+extern void RADebugLog(const char* sFormat, ...);
+extern BOOL DirectoryExists(const char* sPath);
+
+const int SERVER_PING_DURATION = 2 * 60;
```
Jamiras: i assume these are just whitespace changes. recommend leaving them as they were before to keep the review smaller. we can fix the whitespace in #15

src/RA_Defs.h
```diff
@@ -225,30 +255,382 @@ typedef DWORD			ARGB;
...
+_CONSTANT_VAR operator""_gameid(unsigned long long n) noexcept {
+    return static_cast<GameID>(n);
+} // end operator""_gameid
+
+} // inline namespace int_literals
+
+
+inline namespace conversions {
+
+// Noticed some weird uses of rvalues returning as lvalues, whatever they aren't used anywhere
```
Jamiras: this comment should be removed, it's more appropriate as a review comment wherever the problematic code was.

src/RA_Defs.h
```diff
@@ -225,30 +255,382 @@ typedef DWORD			ARGB;
...
+/// </typeparam>
+/// <returns><paramref name="str" /> as a <c>DataStream</c>.</returns>
+/// <remarks>
+///   <c>DataStream</c> is just an alias for an unsigned standard string.
+/// </remarks>
+template<
+    typename CharT,
+    class = std::enable_if_t<is_char_v<CharT> && std::is_signed_v<CharT>>
+>
+DataStream stodata_stream(const std::basic_string<CharT>& str) noexcept
```
Jamiras: This should just be `to_data_stream`.

Is it even used?

src/RA_Defs.h
```diff
@@ -225,30 +255,382 @@ typedef DWORD			ARGB;
...
-typedef std::basic_string<TCHAR> tstring;
 
 
+
+
+
+namespace ra {
+using cstring       = const char*;
+using cwstring      = const wchar_t*;
+using tstring       = std::basic_string<TCHAR>;
```
Jamiras: keep `tstring`, it was there before, and it makes sense to have it. the other two just make the code more obscure to save a couple characters.

src/RA_Defs.h
```diff
@@ -225,30 +255,382 @@ typedef DWORD			ARGB;
...
+
+template<typename CharT>
+using string_t = std::basic_string<CharT>;
```
Jamiras: again this just adds another layer of abstraction. you just end up saying "string_t<char>" instead of "std::basic_string<char>", which is already aliased as "std::string"

src/RA_Defs.h
```diff
@@ -225,30 +255,382 @@ typedef DWORD			ARGB;
...
+
+using DataStream     = std::basic_string<BYTE>;
+using ustring        = DataStream;
+using uofstream      = std::basic_ofstream<BYTE>;
+using uifstream      = std::basic_ifstream<BYTE>;
+using uostringstream = std::basic_ostringstream<BYTE>;
+using uistringstream = std::basic_istringstream<BYTE>;
+using uostream       = std::basic_ostream<BYTE>;
+using uistream       = std::basic_istream<BYTE>;
```
Jamiras: What are all of these `u` data types? Please don't define things until we need to use them.

src/RA_Defs.h
```diff
@@ -225,30 +255,382 @@ typedef DWORD			ARGB;
...
+
+using ByteAddress   = std::size_t;
+using DataPos       = std::size_t;
+using AchievementID = std::size_t;
+using LeaderboardID = std::size_t;
+using GameID        = std::size_t;
```
Jamiras: i don't think any of these will ever be more than 32-bits, but since we're targeting 32-bit, size_t is 32-bits.

src/RA_Defs.h
```diff
@@ -225,30 +255,382 @@ typedef DWORD			ARGB;
...
+
+// forgot to add this
+// if you are already in ra or type "using namespace ra" that's good enough
+inline namespace int_literals
+{
+
+// Don't feel like casting non-stop
```
Jamiras: These feel dirty. More "magic" code, and most of them will never be used.

src/RA_Defs.cpp
```diff
@@ -4,59 +4,79 @@
 #include <Windows.h>
 #include <locale>
 #include <codecvt>
```
Jamiras: `locale` and `codecvt` probably aren't used any more.

src/RA_Defs.h
SyrianBallaS: Well size_t has been used throughout the project before I even touched it, it felt appropriate.

src/RA_Defs.cpp
SyrianBallaS: Probably not

src/RA_Defs.h
SyrianBallaS: Well there's a few literals that don't exist. size_t, inptr_t/ptrdiff_t, DWORD (it has a variable size like size_t, it gave me warnings before somewhere), short, unsigned short. It would be a lot simpler to just use a literal instead of a static cast when that can already be done. A lot of them will be used. Implicit conversions are probably the the second if not first worst problem in programming.

src/RA_Defs.h
SyrianBallaS: These are so we don't need to convert back and forth between std::string and DataStream. We're not using std::u16string/u32string but it's still there.

src/RA_Defs.cpp
```diff
@@ -4,59 +4,79 @@
...
+#include <clocale>
 GetParseErrorFunc GetJSONParseErrorStr = GetParseError_En;
+namespace ra {
 
-static_assert(sizeof(BYTE*) == sizeof(char*), "dangerous cast ahead");
-char* DataStreamAsString(DataStream& stream)
-{
-	return reinterpret_cast<char*>(stream.data());
-}
+inline namespace conversions{
```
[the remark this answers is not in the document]
SyrianBallaS: OK, good to know. Thanks!

SyrianBallaS:
wstring_convert has been deprecated

src/RA_Defs.cpp
```diff
@@ -4,59 +4,79 @@
 #include <Windows.h>
 #include <locale>
 #include <codecvt>
-
+#include <iomanip>
+#include <clocale>
 GetParseErrorFunc GetJSONParseErrorStr = GetParseError_En;
```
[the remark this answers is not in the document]
SyrianBallaS: Sure

src/RA_Defs.cpp
```diff
@@ -4,59 +4,79 @@
...
-std::string Narrow(const wchar_t* wstr)
-{
-	static std::wstring_convert< std::codecvt_utf8_utf16< wchar_t >, wchar_t > converter;
-	return converter.to_bytes(wstr);
+    auto str{ convert_string<std::string>(stream) };
+
+    // pesky null character
+    if (!str.empty()) {
+        str.pop_back();
+    }
```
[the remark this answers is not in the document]
SyrianBallaS: I did do that before, I think there's a similar callback in RA_Interface. When I tried removing the extra null character from RA_httpthread the app threw an exceptions.

Oh boy RA_Interface itself needs a lot of work, but I'm gonna try holding off on it for now. I still need to find where that callback function is, I recall encountering it when I first started diving into this code 2 years ago.

src/RA_Defs.cpp
```diff
@@ -4,59 +4,79 @@
...
+	std::ostringstream oss;
+	oss << "0x" << std::setfill('0') << std::setw(6) << std::hex << nAddr;
+	return oss.str();
 }
 
+
+std::wstring Widen(const wchar_t* wstr){return std::wstring{ wstr }; }
+std::wstring Widen(const std::wstring& wstr) { return wstr; }
+std::string Narrow(const char* str) { return std::string{ str }; }
+std::string Narrow(const std::string& str) { return str; }
```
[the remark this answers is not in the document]
SyrianBallaS: You mean indent the namespace? Sure if that's what you want.
If you're wondering most of the styling I get is from here: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md

But w/e, the standards this project uses will be fine with me as well. We just need to have them documented so there are no ambiguities.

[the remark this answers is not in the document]
Jamiras: no, just between lines 73 and 74.

src/RA_Defs.h
Jamiras: if these are just for the conversion, they should go in the .cpp file where the conversion function exists.

src/RA_Defs.h
Jamiras: https://github.com/RetroAchievements/RAIntegration/blob/master/src/RA_Defs.h#L203-L207

Looks like ByteAddress was already 64-bit and this actually shortens it to 32-bit.

src/RA_Defs.h
Jamiras: >A lot of them will be used.

Can we not add them until they are?

SyrianBallaS: Sure

src/RA_Defs.h
SyrianBallaS: <code>unsigned long</code> isn't 64-bit, <code>unsigned long long</code> is. At least in VC++.

src/RA_Defs.h
SyrianBallaS: Sure, I see what what you mean.

src/RA_Defs.cpp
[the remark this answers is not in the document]
SyrianBallaS: ok

src/RA_Defs.h
SyrianBallaS: <code>inline constexpr auto</code> can sometimes be tedious to write out each time but if it's that important they can go away.

src/RA_Defs.h
SyrianBallaS: I was arguing against macros for constants, keywords aren't used in execution, they're just placeholders.

src/RA_Defs.h
SyrianBallaS: I really need to stop doing using that keyboard shortcut whenever the indentation gets uneven...

I got you.

src/RA_Defs.h
SyrianBallaS: if you say so.

src/RA_Defs.h
SyrianBallaS: sure

src/RA_Defs.h
SyrianBallaS: sure

src/RA_Defs.h
SyrianBallaS: It's preparation, I remember using it in my RASuite fork but it's not used here yet.

SyrianBallaS:
I think it should be ok now.

SyrianBallaS:
Alright, this should be fine. FIrst I made my master the same as RetroAchievements. Merged the branch in this PR with my master. Merged my master with the branch in this PR.

src/RA_Defs.cpp
```diff
@@ -1,68 +1,87 @@
...
+    return str; // ok it's definitly showing how it should be
 }
 
-std::string Narrow(const std::wstring& wstr)
-{
-	static std::wstring_convert< std::codecvt_utf8_utf16< wchar_t >, wchar_t > converter;
-	return converter.to_bytes(wstr);
+
+std::string Narrow(const std::wstring& wstr) {
+    return convert_string<std::string>(wstr);
```
Jamiras: This doesn't work. If passed L"TEST", it returns "84698384". The "oss << i" treats each wchar_t as an integer.

src/RA_Defs.cpp
```diff
@@ -1,68 +1,87 @@
...
 
-std::wstring Widen(const wchar_t* wstr)
+std::wstring Widen(const char* str)
 {
-	return std::wstring(wstr);
+    auto state = std::mbstate_t();
+    auto len = 1 + std::mbsrtowcs(nullptr, &str, 0, &state);
+    auto wstr{ L""s };
+    wstr.reserve(len);
+    MultiByteToWideChar(CP_UTF8, MB_PRECOMPOSED, str, len, wstr.data(),
```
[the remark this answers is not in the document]
SyrianBallaS: Alright didn't know.

src/RA_Defs.cpp
SyrianBallaS: OK, thanks for letting me know. It seemed to work for directories and web response parsing. I will look again.

SyrianBallaS:
I think it's ok. Visual Studio keeps on throwing exceptions at me, from stuff I did not change. Like in RA_Dlg_AchEditors WindowProc.

SyrianBallaS:
(reopen later)

SyrianBallaS:

SyrianBallaS:
Conflicts are resolved.

SyrianBallaS:
I'm certain the formatting was fine, need to check again. I might need to talk to GitHub about this...

SyrianBallaS:
(reopen later)

src/RA_Integration.vcxproj
```diff
@@ -74,6 +74,7 @@
     <ClInclude Include="RA_Dlg_Memory.h" />
     <ClInclude Include="RA_Dlg_RichPresence.h" />
     <ClInclude Include="RA_Dlg_RomChecksum.h" />
+    <ClInclude Include="ra_fwd" />
```
Jamiras: why do these new files not have an extension, and why is this one "ClInclude" whereas the other two are "None"?

src/ra_utility
```diff
@@ -0,0 +1,241 @@
...
+// function alias template for etoi (EnumToIntegral)
+template<typename Enum> _NODISCARD _CONSTANT_VAR to_integral = etoi<Enum>;
+
+namespace detail {
+
+/// <summary>
+///   Casts <typeparamref name="InputString" /> into an
+///   <typeparamref name="OutputString" />. This won't work with multi-byte
+///   strings, only byte strings. It can be used to convert <c>std::string</c> to
+///   and from <c>std::wstring</c> but that's only if they are byte strings.
```
Jamiras: any data sent to/from the server is assumed to be UTF-8. as such, any internal strings should also be assumed to be UTF-8. `Widen` and `Narrow` _must_ function with UTF-8 strings.

src/ra_utility
```diff
@@ -0,0 +1,241 @@
...
+>
+_NODISCARD OutputString string_cast(const InputString& str) noexcept
+{
+	if (str.empty())
+		return OutputString{}; // an alternative to throwing an exception
+
+	using out_ostringstream = std::basic_ostringstream<typename OutputString::value_type>;
+	out_ostringstream oss;
+	std::ios::sync_with_stdio(false);
+	for (auto& i : str)
```
Jamiras: it looks like this function is only used to convert `DataStream` (i.e. byte array) to/from `std::string`. As such, there will never be size converting and using a for loop to copy data seems quite inefficient.

Jamiras: Furthermore, `DataStream` is only used by RAWeb as a way to encapsulate the server response, which must support UTF-8. Nothing appears to be converting from std::string back to a `DataStream`. The parameters passed to RAWeb are UTF-8 std::strings.

I'd rather modify RAWeb to return a std::string (assumed to be UTF-8) and eliminate the `DataStream` data type entirely.

.gitignore
```diff
@@ -5,3 +5,5 @@
 *.user
 /RetroAchievementsKey
 test_app/
+src/*_REMOTE_*.cpp
+src/clipboard.txt
```
Jamiras: i realize these are to be ignored, but where do they come from such that they need to be ignored?

src/RA_Defs.h
```diff
@@ -62,19 +62,20 @@
...
-#include <sstream>
-#include <queue>
 #include <map>
 #include <array>
+#include <ciso646>
+#include <sstream>
+#include <queue>
+#include <iostream>
+#include <ra_utility>
+#include <ra_type_traits>
```
Jamiras: brackets are associated with system headers. prefer quotes for local files.

src/RA_Defs.h
```diff
@@ -89,15 +90,31 @@
...
+
 using namespace rapidjson;
 extern GetParseErrorFunc GetJSONParseErrorStr;
 #pragma warning(pop)
+
+
+using namespace std::string_literals;
+//using namespace std::chrono_literals; we could use this later
+namespace ra {}
+using namespace ra;
```
Jamiras: what does this do (declaring an empty namespace and using it)?

Jamiras: also, per CONTRIBUTING, you can use things in a namespace, but not the entire namespace.

src/RA_Defs.h
```diff
@@ -144,14 +165,12 @@ extern GetParseErrorFunc GetJSONParseErrorStr;
...
-typedef unsigned char	BYTE;
-typedef unsigned long	DWORD;
-typedef int				BOOL;
-typedef DWORD			ARGB;
+
 
 //namespace RA
 //{
 template<typename T>
+_DEPRECATEDR("Replaced with std::clamp")
```
Jamiras: this is only called in one place. suggest making the replacement and removing this code.

src/RA_Dlg_GameTitle.cpp
```diff
@@ -108,7 +108,7 @@ INT_PTR Dlg_GameTitle::GameTitleProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM
                     {
                         //	Existing title
                         ASSERT(m_aGameTitles.find(std::string(sSelectedTitle)) != m_aGameTitles.end());
-                        nGameID = m_aGameTitles[std::string(Narrow(sSelectedTitle))];
+                        nGameID = m_aGameTitles[std::string(ra::Narrow(sSelectedTitle))];
```
Jamiras: i realize the code was doing this before, but `Narrow` already returns a `std::string`, it's unnecessary to wrap it again.

src/RA_httpthread.cpp
```diff
@@ -245,7 +245,7 @@ BOOL RAWeb::DoBlockingRequest(RequestType nType, const PostArgs& PostData, Docum
     {
         if (response.size() > 0)
         {
-            JSONResponseOut.Parse(DataStreamAsString(response));
+			JSONResponseOut.Parse( DataStreamAsString( response ).c_str());
```
Jamiras: these are tabs. you can tell by highlighting in the browser - you can't select the individual spaces.

src/ra_fwd
```diff
@@ -0,0 +1,37 @@
...
+using ARGB          = DWORD;
+using ByteAddress   = std::size_t;
+using DataPos       = std::size_t;
+using AchievementID = std::size_t;
+using LeaderboardID = std::size_t;
+using GameID        = std::size_t;
+
+} // namespace ra
+
+#endif // !RA_FWD_H
```
Jamiras: how are any of these forward declarations? they all look like variations of typedefs.

src/ra_utility
SyrianBallaS: This will be gone if that's the case.

src/ra_utility
SyrianBallaS: Yeah it's possible to use <code>std::string</code> for utf-8, there's also a literal <code>u8</code> to force it to be UTF-8 in C++17 such as <code>u8"Hello world!"</code>.

I'm going to see if it will still work with just <code>std::string</code> as I go over this again.

.gitignore
SyrianBallaS: clipboard.txt is file I'm using to copy and paste stuff (not trying to sound smart-alecky). <code>*_REMOTE_*.cpp</code> came from KDiff3, a merging tool I use to resolve conflicts.

src/RA_Defs.h
SyrianBallaS: Sure thing.

src/RA_Defs.h
SyrianBallaS: These will be gone.

src/RA_Defs.h
SyrianBallaS: Sounds good to me.

src/RA_Dlg_GameTitle.cpp
SyrianBallaS: Sounds good to me.

src/RA_httpthread.cpp
SyrianBallaS: Thanks for pointing it out.

src/ra_fwd
SyrianBallaS: It's the C++11 way of doing a <code>typedef</code>. It's to prevent the compiler from from saying type mismatch (every type declared is treated as a different type, but using aliases a type, it just "uses" a different name for a type that already exists) since using has the same semantics as <code>auto</code>.

In another branch (the handle poc pr I think?). I actually did put forward declarations so I wouldn't have to include  <code>Windows.h</code> in the files I added. I got the idea from  <code>iosfwd</code> which was pretty much a bunch of <code>using</code>s.

SyrianBallaS:
As far as I can tell everything works with the recommendations. DataStream has been eliminated.

src/RA_Core.cpp
```diff
@@ -1545,37 +1548,24 @@ void _WriteBufferToFile(const std::string& sFileName, const Document& doc)
...
-    SetCurrentDirectory(NativeStr(g_sHomeDir).c_str());
-    FILE* pf = nullptr;
-    if (fopen_s(&pf, sFileName.c_str(), "wb") == 0)
-    {
-        fwrite(sData.data(), 1, sData.length(), pf);
-        fclose(pf);
-    }
+    using FileH = std::unique_ptr<FILE, decltype(&std::fclose)>;
+    FileH myFile{ std::fopen(sFileName.c_str(), "wb"), std::fclose };
+    std::fwrite(static_cast<const void*>(raw.c_str()), sizeof(char), raw.length(), myFile.get());
```
SyrianBallaS: I checked that the files were still being written to by deleting the files first.

src/RA_Core.cpp
```diff
@@ -1545,37 +1548,24 @@ void _WriteBufferToFile(const std::string& sFileName, const Document& doc)
...
-    if (fopen_s(&pf, sFile, "wb") == 0)
-    {
-        fwrite(sBuffer, 1, nBytes, pf);
-        fclose(pf);
-    }
+
+    using FileH = std::unique_ptr<FILE, decltype(&std::fclose)>;
+    FileH myFile{ std::fopen(sFile, "wb"), std::fclose };
+    auto sBuffer{ std::make_unique<char[]>(ra::to_unsigned(nBytes)) };
+    std::fwrite(static_cast<void* const>(sBuffer.get()), sizeof(char), std::strlen(sBuffer.get()), myFile.get());
```
SyrianBallaS: Wasn't sure if I should've removed this, even with original signature it was referenced anywhere.

src/RA_Core.cpp
```diff
@@ -1636,7 +1626,7 @@ std::string GetFolderFromDialog()
     std::string sRetVal;
 	CComPtr<IFileOpenDialog> pDlg;
 
-    HRESULT hr;
+    HRESULT hr = HRESULT{};
```
SyrianBallaS: Didn't want to guess what "0" was for HRESULT. It's like a lot of other languages but without the new.

src/RA_Defs.cpp
```diff
@@ -9,55 +9,74 @@ GetParseErrorFunc GetJSONParseErrorStr = GetParseError_En;
...
+std::string Narrow(const wchar_t* wstr)
 {
-    static std::wstring_convert< std::codecvt_utf8_utf16< wchar_t >, wchar_t > converter;
-    return converter.from_bytes(str);
+    auto state{ std::mbstate_t{} };
+    auto len{ 1 + std::wcsrtombs(nullptr, &wstr, 0_z, &state) };
+
+    // doesn't work with unique_ptr, though it works with sprintf and the functions in RAWeb
+    std::string str;
+    str.reserve(len);
```
SyrianBallaS: could have sworn I removed that comment. Yeah I tried using the array of unique_ptr before using a string but it didn't work. I don't like static arrays and I don't like using new directly.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -118,7 +118,7 @@ void Dlg_AchievementEditor::SetupColumns(HWND hList)
     {
         col.mask = LVCF_TEXT | LVCF_WIDTH | LVCF_SUBITEM | LVCF_FMT;
         col.cx = COLUMN_WIDTH[i];
-        tstring colTitle = COLUMN_TITLE[i];	//	Take non-const copy
+        ra::tstring colTitle = COLUMN_TITLE[i];	//	Take non-const copy
         col.pszText = const_cast<LPTSTR>(colTitle.c_str());
```
SyrianBallaS: Didn't really work on it since it's unrelated to this PR but in #61 I observed Spectre vulnerabilities when compiling with /Wall. Just thought I'd point it here. Spectre was that thing that reads kernel memory. Most of the places they popped up was with C arrays being assigned or compared.

src/RA_Core.cpp
Jamiras: It looks like it was intended to be called from a couple places in RA_Core.cpp that are currently commented out. I'm fine with removing it - less code to test. We have git history if we ever need to add it back.

src/RA_Defs.h
```diff
@@ -300,46 +300,46 @@ const int SERVER_PING_DURATION = 2 * 60;
...
 #ifndef RA_UTEST
 #undef ASSERT
 #define ASSERT( x ) assert( x )
 #else
 #undef ASSERT
 #define ASSERT( x ) {}
-#endif
-#else
-#undef ASSERT
-#define ASSERT( x ) {}
```
Jamiras: I currently get around 200 build errors in RELEASE mode. About half of them are "ASSERT is undefined"

Jamiras: the others fall into two categories:

not including in files that use unique_ptr
unit test code related to latest changes.

src/RA_Defs.cpp
```diff
@@ -9,55 +9,74 @@ GetParseErrorFunc GetJSONParseErrorStr = GetParseError_En;
...
+
+    ::WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, wstr, to_signed(len), str.data(),
+        to_signed(str.length()), nullptr, nullptr);
```
Jamiras: unit tests caught this one: `str.length()` is 0 at this point. it should be `wcslen(wstr) + 1`

i'm not sure what's going on with the second EARTH GLOBE unit test, `wcsrtombs` is returning -1, I suspect it doesn't support surrogate pairs.

Jamiras: what is the rationale behind changing this from `std::wstring_convert`?

SyrianBallaS: The JSON responses were getting parsed and I saw the Rich Presence stuff still getting printed correctly. 

Rather than explaining it you can read it.
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0618r0.html

At this point I don't really care too much, it needed to be capacity and not length since the size is 0.

Jamiras: Just make sure the unit tests pass. That's why they're there.

SyrianBallaS: Thanks for pointing it out. I checked it out using the memory bookmarker for saving files and it was able to write to it in tact.

src/RA_Defs.h
SyrianBallaS: I need to take a break for awhile after this. That is so strange. Don't know why these review comments aren't showing up when they should.

Well the ASSERT problem did exist, but I think the <memory> problem was a false error, I didn't get an error.

SyrianBallaS:
Gonna chill on the PRs for awhile.

SyrianBallaS:
Closing it for now, doing the Unit tests a few more times.

SyrianBallaS:
All unit tests pass, on Release mode and Debug mode. The "fix" was to make sure the buffer was big enough. A unicode character could be up to 4 bytes which the unit test had a 4 byte character. So basically the buffer had to be at least 4 times greater than the size of the wide string. I read somewhere that American unicode are 1 byte, and the rest are 2-4.

src/RA_Defs.h
Jamiras: this is what I get without manually adding the missing `#include <memory>`s:
```
1>e:\source\raintegration\src\ra_core.cpp(1555): error C2039: 'unique_ptr': is not a member of 'std'
```
```
1>e:\source\raintegration\src\ra_httpthread.cpp(343): error C2039: 'make_unique': is not a member of 'std'
```
Looks like it's just those two files.

src/RA_Defs.cpp
Jamiras: RA_Defs_Tests::TestNarrow is still failing:
```
Message: Assert failed. Expected:<Test> Actual:<TestÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍÍýýýý>
```
I had to add a +1 to the length to get it to copy the null terminator:
```
    ::WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, wstr, lstrlenW(wstr) + 1, str.data(),
        str.capacity(), nullptr, nullptr);
```
Also changed the `lstrlenW(wstr) * 4` to `str.capacity()` to avoid a second strlen call.

Jamiras:
UTF-8 is actually variable length. ASCII characters take 1-byte. As a general statement, Latin characters take 2, and Asian characters take 3. That covers the entire 16-bit unicode set. However, to handle surrogate pairs, the encoded character requires 4 or more bytes. The UTF-8 spec defines mapping for Unicode code points up to 31-bits, which would take 6 bytes.

The recommendation is to call ``::WideCharToMultiByte`` with an output buffer size of 0 to determine the actual size needed:
```
std::string Narrow(const wchar_t* wstr)
{
    size_t len = lstrlenW(wstr);
    int needed = ::WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, wstr, len + 1, nullptr, 0, nullptr, nullptr);

    std::string str;
    str.reserve(needed);

    ::WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, wstr, len + 1, str.data(), str.capacity(), nullptr, nullptr);
    return str.data(); // the .data() part is required
}
```

Jamiras:
And this implementation alleviates the need to build a new string from `str.data()`. The problem is that `reserve()` doesn't update the size, so despite having the data in the string, the size is still set to 0.
```
std::string Narrow(const wchar_t* wstr)
{
    size_t len = lstrlenW(wstr);
    int needed = ::WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, wstr, len + 1, nullptr, 0, nullptr, nullptr);

    std::string str(needed, 0); // allocate required space (including terminator)
    ::WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, wstr, len + 1, str.data(), str.capacity(), nullptr, nullptr);
    str.resize(needed - 1); // terminator is not actually part of the string
    return str;
}
```

SyrianBallaS:
GitHub is looking funky (assets won't load) so I'm replying by email. Those
are actually really good ideas. From what I've seen, "reserve" is required
is using C based string building since you can't use std::string. You can
update the size (might seem redundant) by assigning the string the pointer
to the string if you want the size updated.

Anyway, I think your ideas are really good.

SyrianBallaS:
Most of the lines `added` are for comparison operators for `enum classes` and a `tstrtoul` wrapper depending on the charset. Most of  lines _changed_ this time were to make the unicode part compatible without sacrificing the MBCS version. I was thinking of doing one of `strol` but never seen any pointer arithmetic in it so `stoi` should be fine.

PR #24, opened 2018-04-25
Additional changes to support rendering in RALibRetro

Jamiras:
Changes for https://github.com/RetroAchievements/RALibretro/issues/3

* exposed RA_IsOverlayFullyVisible so RALibRetro can not render the screen when the overlay is visible
* disabled rendering popups while overlay is visible
* fix updating highlighted address in Memory Inspector when RALibRetro is paused.

### DIFF
--- a/src/RA_AchievementOverlay.cpp
+++ b/src/RA_AchievementOverlay.cpp
@@ -1777,6 +1777,11 @@ API void _RA_RenderOverlay( HDC hDC, RECT* rcSize )
 	g_AchievementOverlay.Render( hDC, rcSize );
 }
 
+API bool _RA_IsOverlayFullyVisible()
+{
+    return g_AchievementOverlay.IsFullyVisible();
+}
+
 API void _RA_InitDirectX()
 {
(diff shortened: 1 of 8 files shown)


meleu:
@Jamiras could you attach here the compiled `.dll` file so I can make some tests with RALibretro?
(I think the `dll` needs to be zipped to allow you to attach the file)

Jamiras:
[RALibRetro.zip](https://github.com/RetroAchievements/RAIntegration/files/1947281/RALibRetro.zip)

Includes updated DLL and Libretro.exe

meleu:
@Jamiras when I try to launch RALibretro it complains about a missing `MSVCP140.dll`. Then I'm unable to test it.

Are your changes adding new dependencies?

meleu:
Yeah, you're right. I tried with a version compiled via Makefile and that problem doesn't happen.

But unfortunately I wasn't able to see the changes you implemented because the overlay doesn't work at all on a Virtual Machine. It's a known issue and IIRC it also happens with @leiradel .

PR #27, opened 2018-04-27
NULL to nullptr

SyrianBallaS:
Basically all this does is replace NULL with nullptr, for params that aren't pointers they were replaced with rvalues. Small PRs for now. The little things can sometimes lead to unexpected results.

### DIFF
--- a/src/RA_Core.cpp
+++ b/src/RA_Core.cpp
@@ -471,7 +471,7 @@ API void CCONV _RA_ClearMemoryBanks()
 //	const unsigned int nBufferSize = (3*1024*1024);	//	3mb enough?
 //
 //	char* buffer = new char[nBufferSize];	
-//	if( buffer != NULL )
+//	if( buffer != nullptr )
 //	{
 //		char sAddr[1024];
 //		sprintf_s( sAddr, 1024, "/files/%s", sFilename );
@@ -483,7 +483,7 @@ API void CCONV _RA_ClearMemoryBanks()
 //			_WriteBufferToFile( sOutput, buffer, nBytesRead );
 //
 //		delete[] ( buffer );
-//		buffer = NULL;
+//		buffer = nullptr;
 //	}
 //}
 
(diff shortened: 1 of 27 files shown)


src/RA_Core.cpp
```diff
@@ -525,18 +525,18 @@ API BOOL CCONV _RA_OfferNewRAUpdate(const char* sNewVer)
...
 		//	SW_SHOWNORMAL ); 
 
-		ShellExecute(NULL,
+		ShellExecute(nullptr,
 			TEXT("open"),
 			TEXT("http://www.retroachievements.org/download.php"),
-			NULL,
-			NULL,
+			nullptr,
+			nullptr,
```
Jamiras: Is it safe to use `nullptr` in the WINAPI calls?

src/RA_Core.cpp
```diff
@@ -752,32 +752,33 @@ API HMENU CCONV _RA_CreatePopupMenu()
 	HMENU hRA_LB = CreatePopupMenu();
 	if (RAUsers::LocalUser().IsLoggedIn())
 	{
+		// TODO: Use the literals once PR 23 is accepted
 		AppendMenu(hRA, MF_STRING, IDM_RA_FILES_LOGOUT, TEXT("Log&out"));
-		AppendMenu(hRA, MF_SEPARATOR, NULL, NULL);
+		AppendMenu(hRA, MF_SEPARATOR, UINT_PTR{}, nullptr);
```
Jamiras: I think it would be more standard/common to use `0` instead of `UINT_PTR{}`.

src/RA_Defs.h
```diff
@@ -11,6 +11,8 @@
 #include <queue>
 #include <deque>
 #include <map>
+#include <atlbase.h>
```
Jamiras: why was this necessary?

src/RA_Core.cpp
SyrianBallaS: Yes, "far* LP" or "near* P" stuff are pointers. I did notice some NULLs in the AppendMenu were actually supposed to be integers. Windows NT is compatible with C++ if you know how to use it, Microsoft themselves recommend using nullptr.

src/RA_Core.cpp
SyrianBallaS: Well I was going to use my literals in a different pull request but felt that I shouldn't copy and paste them. <code>UINT_PTR</code> is basically the same as <code>std::size_t</code>. Starting with C++11, rvalue implicit conversions are flagged.

But I know what you mean.

src/RA_Defs.h
SyrianBallaS: It's not. At least it wasn't here. It is needed for CComPtr though. I think I was going to that first but thought I should do something simpler.

SyrianBallaS:
@Jamiras Thanks for the review

src/RA_Core.cpp
SyrianBallaS: I was looking into it further and <code>UINT_PTR</code> is a variable/polymorphic type, it's either <code>uintptr_t</code>, <code>unsigned int</code>, or <code>unsigned __int64</code> depending on where you got it from. So I believe it's more appropriate to hold off on it for correctness. The other standard way to do it would be <code>UINT_PTR{0}</code> but that's pretty much the same thing. <code>type{}</code> just means <code>default</code>.

src/RA_Defs.h
SyrianBallaS: This is gone

PR #28, opened 2018-04-27
COM Interfaces in Smart Pointers and slight tweaks.

SyrianBallaS:
COM Interfaces wrapped in smart pointers. The only places I found that used COM was the ImagingFactory and OpenFileDialog. Strangely enough, I can actually see my Profile Picture in the Overlay but not on the website.

Tested the OpenFileDialog from the GameLibrary dialog, the ImageFactory was tested to see if the images would show up correctly. Game used to test was Contra.


src/RA_Defs.h
```diff
@@ -11,6 +11,8 @@
 #include <queue>
 #include <deque>
 #include <map>
+#include <memory>
+#include <atlbase.h>
```
Jamiras: prefer to have this in the files that need it, rather than global. `RA_Defs` is included by everything, not just the UI.

src/RA_Core.cpp
```diff
@@ -1590,35 +1590,41 @@ BOOL _FileExists(const std::string& sFileName)
...
-	//CoUninitialize();
+                    // https://msdn.microsoft.com/en-us/library/windows/desktop/bb761140(v=vs.85).aspx
+                    CoTaskMemFree(static_cast<LPVOID>(pStr));
+                    pStr = LPWSTR{}; // we didn't use new, so just reset it
+				} // end if
+				pItem.Release();
+			} // end if
+		} // end if
+		pDlg.Release(); // Ok, everythings nullified except the return
+	} // end if
```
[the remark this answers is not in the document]
SyrianBallaS: Sure.

src/RA_Defs.h
```diff
@@ -21,6 +23,10 @@
 //NB. These must NOT be accessible from the emulator!
 //#define RA_INTEGRATION_VERSION	"0.053"
 
+// Dammit, next pr is gonna be an update
```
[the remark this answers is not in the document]
SyrianBallaS: Sure, I understand.

src/RA_ImageFactory.cpp
```diff
@@ -21,118 +21,114 @@ UserImageFactoryVars g_UserImageFactoryInst;
...
+	} // end if
 
-	if( SUCCEEDED( hr ) )
+	if (SUCCEEDED(hr))
 	{
 		// Size of a scan line represented in bytes: 4 bytes each pixel
-		hr = UIntMult( nWidth, sizeof( ARGB ), &cbStride );
-	}
+		hr = UIntMult(nWidth, sizeof(ARGB), &cbStride);
+	} // end if
```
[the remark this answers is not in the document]
SyrianBallaS: Sure I understand

src/RA_ImageFactory.cpp
```diff
@@ -21,118 +21,114 @@ UserImageFactoryVars g_UserImageFactoryInst;
...
-	g_UserImageFactoryInst.m_pOriginalBitmapSource = NULL;
 
-	HeapSetInformation( NULL, HeapEnableTerminationOnCorruption, NULL, 0 );
+_Use_decl_annotations_
+BOOL InitializeUserImageFactory([[maybe_unused]] HINSTANCE hInst)
+{
+	HeapSetInformation(HANDLE{}, HeapEnableTerminationOnCorruption, PVOID{}, SIZE_T{});
 
-	hr = CoInitializeEx( NULL, COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE );
+	auto hr = CoInitializeEx(LPVOID{}, COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE);
```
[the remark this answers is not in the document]
SyrianBallaS: If it's that important, sure.

src/RA_Integration.vcxproj
```diff
@@ -189,6 +190,10 @@
     <ResourceCompile>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ResourceCompile>
+    <PostBuildEvent>
+      <Command>copy /y $(TargetPath) $(LocalDebuggerWorkingDirectory)$(TargetFileName)
+copy /y $(TargetPath) $(LocalDebuggerWorkingDirectory)$(TargetName).pdb</Command>
```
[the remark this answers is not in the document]
SyrianBallaS: OK, well I sometimes couldn't get the debug info I needed. Maybe I did something wrong.

src/RA_Defs.h
SyrianBallaS: Sure, I was gonna try to a precompiled header (it will work, had it working in RASuite fork) for faster compile and preventing double includes.

I agree with you though.

src/RA_ImageFactory.cpp
```diff
@@ -150,238 +146,243 @@ BOOL InitializeUserImageFactory( HINSTANCE hInst )
...
-	if( SUCCEEDED( hr ) )
-		hr = pDecoder->GetFrame( 0, &pFrame );
+	CComPtr<IWICBitmapDecoder> pDecoder;
+	HRESULT hr = g_UserImageFactoryInst.m_pIWICFactory->CreateDecoderFromFilename(Widen(sPath).c_str(),			// Image to be decoded
+		nullptr,							// Do not prefer a particular vendor
+		GENERIC_READ,					// Desired read access to the file
+		WICDecodeMetadataCacheOnDemand,	// Cache metadata when needed
+		&pDecoder);						// Pointer to the decoder
+
+// Retrieve the first frame of the image from the decoder
```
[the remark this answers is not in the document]
SyrianBallaS: Sure thing.

src/RA_Core.cpp
```diff
@@ -1590,35 +1590,41 @@ BOOL _FileExists(const std::string& sFileName)
...
-			IShellItem* pItem = nullptr;
-			hr = pDlg->GetResult(&pItem);
-			if (hr == S_OK)
+			CComPtr<IShellItem> pItem;
+			if (hr = pDlg->GetResult(&pItem); SUCCEEDED(hr))
 			{
-				LPWSTR pStr = nullptr;
-				hr = pItem->GetDisplayName(SIGDN_FILESYSPATH, &pStr);
-				if (hr == S_OK)
+				auto pStr = LPWSTR{};
```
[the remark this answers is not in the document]
SyrianBallaS: That uses the rvalue initializer, it prevents implicit conversions. But if it's that important, it can go back.

src/RA_Core.cpp
```diff
@@ -1590,35 +1590,41 @@ BOOL _FileExists(const std::string& sFileName)
 	}
 }
 
+
+// Todo: Remove comment later
+// Tested using the open folder dialog from game library
```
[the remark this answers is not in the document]
SyrianBallaS: Sure thing.

SyrianBallaS:
@Jamiras Thanks for reviewing, I'll get to your concerns later. I do think we need a formalized standard though.

src/RA_Core.cpp
```diff
@@ -1590,35 +1590,41 @@ BOOL _FileExists(const std::string& sFileName)
...
+
+	CComPtr<IFileOpenDialog> pDlg;
+	if (auto hr = CoCreateInstance(CLSID_FileOpenDialog, nullptr, CLSCTX_ALL, IID_IFileOpenDialog, 
+		reinterpret_cast<void**>(&pDlg)); SUCCEEDED(hr))
 	{
 		pDlg->SetOptions(FOS_PICKFOLDERS);
-		hr = pDlg->Show(nullptr);
-		if (hr == S_OK)
+
+		if (hr = pDlg->Show(nullptr); SUCCEEDED(hr))
```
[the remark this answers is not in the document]
SyrianBallaS: I prefer keeping the scope as local as possible

src/RA_ImageFactory.cpp
```diff
@@ -21,118 +21,114 @@ UserImageFactoryVars g_UserImageFactoryInst;
...
+_Success_(return == S_OK)
+HRESULT UserImageFactory_CreateDIBSectionFromBitmapSource(_In_ IWICBitmapSource *pToRenderBitmapSource,
+	_Out_ HBITMAP& hBitmapInOut)
 {
-	HRESULT hr = S_OK;
 
-	UINT nWidth = 0;
-	UINT nHeight = 0;
+	auto nWidth  = UINT{};
+	auto nHeight = UINT{};
```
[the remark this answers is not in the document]
SyrianBallaS: It prevents narrowing conversion errors

src/RA_Core.cpp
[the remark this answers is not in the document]
Jamiras: It's important to maintain a common style throughout the code. We don't do this anywhere else in the code, so we shouldn't here either. I have similar thoughts about several of the other changes you're proposing, but they're at least comprehensive (changing all references) and arguably beneficial.

Brace comments are similar to documenting math:
```x = a + b * 6; // multiply b by 6 and add it to a```
If the code is descriptive enough, you don't need to describe it to anyone else. 

The main reason for commenting braces is when they're too far apart from each other to line up visually. This is typically better handled by pulling a chunk of code out into a method so the braces are closer to each other.

File: src/RA_Dlg_MemBookmark.cpp [hunk 12]
```diff
@@ -706,50 +707,59 @@ void Dlg_MemBookmark::ExportJSON()
...
-			char defaultFileName[ 512 ];
-			sprintf_s ( defaultFileName, 512, "%s-Bookmarks.txt", std::to_string( g_pCurrentGameData->GetGameID() ).c_str() );
-			hr = pDlg->SetFileName( Widen( defaultFileName ).c_str() );
-			if ( hr == S_OK )
+            // Does this really need a limit?
+            std::string defaultFileName;
+            defaultFileName.reserve(512); // gives it a cap of 512+16 to prevent overflows but deallocates
+            // for now
+			sprintf_s ( defaultFileName.data(), 512, "%s-Bookmarks.txt",
+                std::to_string( g_pCurrentGameData->GetGameID() ).c_str() );
```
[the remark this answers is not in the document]
Jamiras: possibly all the way to `std::wstring` to avoid the `Widen` call on the next line.

src/RA_ImageFactory.cpp
```diff
@@ -21,118 +21,114 @@ UserImageFactoryVars g_UserImageFactoryInst;
...
-		while( GetParent( hWindow ) != NULL )
-			hWindow = GetParent( hWindow );
+
+		// TODO: Maybe replace some of these with literals
+		BITMAPINFOHEADER info_header{
+			sizeof(BITMAPINFOHEADER), static_cast<LONG>(nWidth),
+			-static_cast<LONG>(nHeight), WORD{1}, WORD{32},
+			static_cast<DWORD>(BI_RGB), DWORD{}, LONG{}, LONG{},
+			DWORD{}, DWORD{}
+		};
```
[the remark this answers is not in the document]
Jamiras: If you insist on building it this way, at least comment what the non-obvious values are:
```
BITMAPINFOHEADER info_header
{
    sizeof(BITMAPINFOHEADER),      // biSize
    static_cast<LONG>(nWidth),
    -static_cast<LONG>(nHeight),
    WORD{1},                       // biPlanes
    WORD{32},                      // biBitCount
...
```

src/RA_Dlg_MemBookmark.cpp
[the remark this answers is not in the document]
SyrianBallaS: Alright, seems good to me.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -706,50 +707,59 @@ void Dlg_MemBookmark::ExportJSON()
...
+    // First time noticing this
+	CComPtr<IFileSaveDialog> pDlg;
+	
+	if (auto hr = CoCreateInstance(CLSID_FileSaveDialog, nullptr, CLSCTX_ALL, IID_IFileSaveDialog,
+        reinterpret_cast<void**>(&pDlg)); SUCCEEDED(hr))
 	{
-		hr = pDlg->SetFileTypes( ARRAYSIZE( c_rgFileTypes ), c_rgFileTypes );
-		if ( hr == S_OK )
+		
+		if (hr = pDlg->SetFileTypes(ARRAYSIZE(c_rgFileTypes), c_rgFileTypes); hr == S_OK )
```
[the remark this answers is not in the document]
SyrianBallaS: Sure thing.

src/RA_Core.cpp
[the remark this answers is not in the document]
SyrianBallaS: Ok, it's fine. They will be gone. You are right that methods should be broken up if too far apart.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -706,50 +707,59 @@ void Dlg_MemBookmark::ExportJSON()
 	defaultDir.erase ( 0, 2 ); // Removes the characters (".\\")
 	defaultDir = g_sHomeDir + defaultDir;
 
-	IFileSaveDialog* pDlg = nullptr;
-	HRESULT hr = CoCreateInstance( CLSID_FileSaveDialog, NULL, CLSCTX_ALL, IID_IFileSaveDialog, reinterpret_cast<void**>( &pDlg ) );
-	if ( hr == S_OK )
+    // First time noticing this
```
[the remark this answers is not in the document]
SyrianBallaS: It's gone.

SyrianBallaS:
@Jamiras All your concerns are legit, I'll get to them when I can

src/RA_ImageFactory.cpp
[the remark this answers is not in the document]
SyrianBallaS: Alright I guess as long as the right size is narrower than the left side (or the same width) it should be fine.

src/RA_ImageFactory.cpp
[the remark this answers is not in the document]
SyrianBallaS: Sure. Seems fair.

src/RA_ImageFactory.cpp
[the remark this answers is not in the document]
SyrianBallaS: sure thing.

src/RA_Integration.vcxproj
```diff
@@ -175,6 +175,7 @@
       <DebugInformationFormat>EditAndContinue</DebugInformationFormat>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <DisableSpecificWarnings>4091</DisableSpecificWarnings>
+      <LanguageStandard>stdcpp17</LanguageStandard>
```
Jamiras: make sure this change gets made to `All Configurations`

SyrianBallaS:
Ok, that can be done

SyrianBallaS:
This one can stay open

SyrianBallaS:

SyrianBallaS:
Everything's good now.

src/RA_Core.cpp
```diff
@@ -1629,35 +1629,39 @@ BOOL _FileExists(const std::string& sFileName)
...
+
+	CComPtr<IFileOpenDialog> pDlg;
+	if (auto hr = CoCreateInstance(CLSID_FileOpenDialog, nullptr, CLSCTX_ALL, IID_IFileOpenDialog, 
+		reinterpret_cast<void**>(&pDlg)); SUCCEEDED(hr))
     {
         pDlg->SetOptions(FOS_PICKFOLDERS);
-        hr = pDlg->Show(nullptr);
-        if (hr == S_OK)
+
+		if (hr = pDlg->Show(nullptr); SUCCEEDED(hr))
```
Jamiras: Please separate this into two lines as it was previously. It's unusual to put a semicolon in the middle of a conditional clause.

Last time I suggested this you mentioned minimizing the scope. Having two separate lines doesn't change the scope of anything. `hr` is defined outside the if statement.

If you're that concerned about putting everything on one line, consider this:
```
if (SUCCEEDED(hr = pDlg->Show(nullptr)))
```

src/RA_Defs.h
```diff
@@ -1,3 +1,5 @@
+#ifndef RA_DEFS_H
+#define RA_DEFS_H
```
Jamiras: isn't this redundant with the `#pragma once` on the next line?

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -23,10 +24,9 @@ const int COLUMN_WIDTH[] = { 112, 64, 64, 64, 54 };
 static_assert(SIZEOF_ARRAY(COLUMN_TITLE) == SIZEOF_ARRAY(COLUMN_WIDTH), "Must match!");
 }
 
-const COMDLG_FILTERSPEC c_rgFileTypes[] =
-{
-    { L"Text Document (*.txt)",       L"*.txt" }
-};
+// If it's only going to be one we could make it a unique_ptr
```
Jamiras: Comments like this should be in the PR, not the code.

The method expects a pointer to a list. Just because the list only has one item doesn't mean that we should use an alternate data type and pretend it's a list.

Additionally, leaving it as a list makes it easier to add additional items in the future if desired.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -704,50 +704,74 @@ void Dlg_MemBookmark::ExportJSON()
...
-                            hr = pDlg->GetResult(&pItem);
-                            if (hr == S_OK)
-                            {
+                            // Exception got thrown here (didn't before), -sbs
+                            // Expression: p==0; where p is pItem.p
+                            // Solution: reset the owned pointer
+                            auto pItem2 = std::move(pItem); // let's see if that'll do it
+
+							if (hr = pDlg->GetResult(&pItem); SUCCEEDED(hr))
+                            {
```
Jamiras: It looks like this code could use a reformatting pass. That should be safe now that #39 has been merged.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -704,50 +704,74 @@ void Dlg_MemBookmark::ExportJSON()
...
-            if (hr == S_OK)
+            std::ostringstream oss;
+            oss.str().reserve(512); // storage on the heap really shouldn't matter but keeping relativelty the same
+            oss << g_pCurrentGameData->GetGameID() << "-Bookmarks.txt";
+
+            auto defaultFileName{ oss.str() };
+            // get in the habit of reseting the stringbuffer inside the
+            // stringstream if you wish to use it more than once (you will get
+            // garabage characters in the end)
+            oss.str("");
```
Jamiras: Comment may be valid (despite the spelling mistakes), but this isn't used in a loop, so it's not relevant here and should be removed. If it is being used below (i didn't see it), the reset should happen there.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -704,50 +704,74 @@ void Dlg_MemBookmark::ExportJSON()
...
+						if (hr = pDlg->Show(nullptr); SUCCEEDED(hr))
                         {
 
-                            hr = pDlg->GetResult(&pItem);
-                            if (hr == S_OK)
-                            {
+                            // Exception got thrown here (didn't before), -sbs
+                            // Expression: p==0; where p is pItem.p
+                            // Solution: reset the owned pointer
+                            auto pItem2 = std::move(pItem); // let's see if that'll do it
```
Jamiras: Comments like this worry me. If you don't understand why you need to work around functionality, maybe you shouldn't be doing it.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -704,50 +704,74 @@ void Dlg_MemBookmark::ExportJSON()
...
+                                    // We really shouldn't use pointers if we don't have too
+                                    // This will still use the pointer however, but that's for another time
+									for (auto bookmark : m_vBookmarks )
                                     {
                                         Value item(kObjectType);
-                                        char buffer[256];
-                                        sprintf_s(buffer, Narrow(bookmark->Description()).c_str(), sizeof(buffer));
-                                        Value s(buffer, allocator);
+										std::string buffer;
+                                        buffer.reserve(256);
```
Jamiras: I realize the old code allocated 256 bytes, but since you're changing this to `std::string`, suggest letting it do its own allocation, and just truncate it too long:
```
std::string buffer = Narrow(bookmark->Description());
if (buffer.length() > 256)
    buffer.resize(256);
Value s {buffer.c_str(), allocator);
```

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -825,42 +852,49 @@ std::string Dlg_MemBookmark::ImportDialog()
 {
     std::string str;
 
+	// With how repetive this is, it should be it's own function
```
Jamiras: more commentary better suited to the PR

src/RA_ImageFactory.cpp
```diff
@@ -54,32 +53,35 @@ HRESULT UserImageFactory_CreateDIBSectionFromBitmapSource(IWICBitmapSource *pToR
...
+		// TODO: Maybe replace some of these with literals
+		BITMAPINFOHEADER info_header{
+			sizeof(BITMAPINFOHEADER),    // biSize
+            static_cast<LONG>(nWidth),
+			-static_cast<LONG>(nHeight),
+            WORD{1},                     // biPlanes
+            WORD{32},                    // biBitCount
+			static_cast<DWORD>(BI_RGB)
+		};
+		BITMAPINFO bminfo{ info_header };
```
Jamiras: what happened here? please fix alignment.

src/RA_ImageFactory.cpp
```diff
@@ -54,32 +53,35 @@ HRESULT UserImageFactory_CreateDIBSectionFromBitmapSource(IWICBitmapSource *pToR
...
 
         // Get a DC for the full screen
-        hdcScreen = GetDC(hWindow);
-        hr = hdcScreen ? S_OK : E_FAIL;
+		auto hdcScreen = GetDC(hWindow);
+
 
         // Release the previously allocated bitmap 
-        if (SUCCEEDED(hr))
+		if (hr = hdcScreen ? S_OK : E_FAIL; SUCCEEDED(hr))
```
Jamiras: I realize this is how it was before, but suggest simplifying it to:
```
if (hdcScreen)
```
there's no reason to get `hr` involved.

src/RA_ImageFactory.cpp
```diff
@@ -150,30 +149,31 @@ BOOL InitializeUserImageFactory(HINSTANCE hInst)
...
 
     pToRenderBitmapSource = nullptr;
 
     // Get the client Rect
     //RECT rcClient = rcDest;
     //hr = GetClientRect(hWnd, &rcClient) ? S_OK: E_FAIL;
 
-    if (SUCCEEDED(hr))
-    {
+	// This top if statement seems redundant
```
Jamiras: Not only did you add a comment here, but you removed the logic it applied to. Please remove.

SyrianBallaS:
Thank you. I realized that some comments should be in a pr. I'll just save
them somewhere else. Raw pointers don't check for anything so it was good
it was caught. What you suggested vs selection statements seems like a good
middle ground. Gonna get the vs spellchecker.

Thank you for the feedback.

src/RA_ImageFactory.cpp
SyrianBallaS: Got it.

SyrianBallaS:
Closing this until I fix the stuff discussed in the reviews.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: OK, I'll save that stuff in another file (not PR) so I don't forget.

src/RA_ImageFactory.cpp
SyrianBallaS: I could have sworn it was aligned right, it looks aligned in Visual Studio. I guess I need to check if it's aligned on GitHub.

src/RA_ImageFactory.cpp
SyrianBallaS: Thanks for the suggestion.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: OK, I'll just save those notes for myself and get the VS Spell Checker while I'm at it (it's an extension has nothing to do with code).

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: I'll merge RetroAchievements/master with this branch preferring the remote branch as far as whitespace is concerned. That should fix the whitespace issue.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: Thanks for pointing that out, I'll study it more. I'm pretty sure this was correct but need to make sure.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: Thank you. From what you're saying, it should be std::array since it's a global. We could always add more later if needed.

src/RA_Defs.h
SyrianBallaS: I was taught to always use include guards. It might seem redundant, it can go away.

src/RA_Core.cpp
SyrianBallaS: The selection statement keep scopes tight. It's not unusual you just don't know about it, everything you don't know is strange at first. People who don't know programming at all think code looks like gibberish.
http://en.cppreference.com/w/cpp/language/if

What you suggested seems like a good middle ground though.

src/RA_Defs.h
```diff
@@ -19,9 +25,32 @@
...
+
+// Maybe an extra check just in-case
+
+#define _NORETURN            [[noreturn]]
+
+#if _HAS_CXX17
+#define _DEPRECATED          [[deprecated]]
+#define _DEPRECATEDR(reason) [[deprecated(reason)]]
+#define _FALLTHROUGH         [[fallthrough]]//; you need ';' at the end
+#define _UNUSED              [[maybe_unused]]
```
SyrianBallaS: _NODISCARD is already defined somewhere else in the STL and [[noreturn]] doesn't require C++17

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -704,50 +704,43 @@ void Dlg_MemBookmark::ExportJSON()
...
-        hr = pDlg->SetFileTypes(ARRAYSIZE(c_rgFileTypes), c_rgFileTypes);
-        if (hr == S_OK)
+        if (SUCCEEDED(hr = pDlg->SetFileTypes(c_rgFileTypes.size(), &c_rgFileTypes.front())))
         {
-            char defaultFileName[512];
-            sprintf_s(defaultFileName, 512, "%s-Bookmarks.txt", std::to_string(g_pCurrentGameData->GetGameID()).c_str());
-            hr = pDlg->SetFileName(Widen(defaultFileName).c_str());
-            if (hr == S_OK)
+            std::ostringstream oss;
+            oss << g_pCurrentGameData->GetGameID() << "-Bookmarks.txt";
```
SyrianBallaS: I was looking into it and I really doubt it will be longer than 512 bytes. Unless you allow the game IDs to be longer than 498 bytes (498 byte/char characters)

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -704,50 +704,43 @@ void Dlg_MemBookmark::ExportJSON()
...
                     {
-                        pDlg->SetDefaultExtension(L"txt");
-                        hr = pDlg->Show(nullptr);
-                        if (hr == S_OK)
+                        if (pDlg->SetDefaultExtension(L"txt"); SUCCEEDED(hr = pDlg->Show(nullptr)))
                         {
-
-                            hr = pDlg->GetResult(&pItem);
-                            if (hr == S_OK)
+                            if (SUCCEEDED(hr = pDlg->GetResult(&pItem.p)))
```
SyrianBallaS: https://docs.microsoft.com/en-us/cpp/atl/reference/ccomptrbase-class#operator_amp
It seems the `CComPtr `needs to be `null `to use the `address-of operator` on it. 
Since it was already used the framework thinks we have buggy code, if
we truly need the address then we need store the address explicit. Which appears to be the case.
We would never have known with just the raw interface pointer
		   
Straight from` atlcomcli.h: CComPtrBase T** operator&()`
The `assert `on `operator&` usually indicates a bug. 
If this is really what is needed, however, take the address of the `p `member explicitly.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -758,17 +751,19 @@ void Dlg_MemBookmark::ExportJSON()
                                     doc.AddMember("Bookmarks", bookmarks, allocator);
```
SyrianBallaS: I did do a format pass on this using the provided vssettings file, GitHub skips a few lines, click "View whole File" if you need to see it for yourself.

src/RA_ImageFactory.cpp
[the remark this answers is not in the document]
SyrianBallaS: image

PR #29, opened 2018-04-27
C Arrays to std::array

SyrianBallaS:
Converted all the "global" C arrays (skimmed the top of each file) with std::array. No risk of memory leak (as far as what was changed). Also missed a few COM stuff and changed those.

You can use the methods std::array has to offer, with the same, if not faster speed. A few globals were turned in to constant expressions to disable accidental const_cast on them.

### DIFF
--- a/src/RA_AchievementOverlay.cpp
+++ b/src/RA_AchievementOverlay.cpp
@@ -16,12 +16,13 @@
 
 namespace
 {
-	const float PAGE_TRANSITION_IN = (-0.2f);
-	const float PAGE_TRANSITION_OUT = ( 0.2f);
-	const int NUM_MESSAGES_TO_DRAW = 4;
-	const char* FONT_TO_USE = "Tahoma";
 
-	const char* PAGE_TITLES[] = { 
+	inline constexpr auto PAGE_TRANSITION_IN   = -0.2f;
+	inline constexpr auto PAGE_TRANSITION_OUT  = 0.2f;
+	inline constexpr auto NUM_MESSAGES_TO_DRAW = 4;
+	inline constexpr auto FONT_TO_USE          = "Tahoma";
+
+	inline constexpr std::array<const char*, NumOverlayPages> PAGE_TITLES {
 		" Achievements ", 
 		" Friends ", 
 		" Messages ",
@@ -35,7 +36,6 @@ namespace
 		" Message Viewer "
 
 		};
-	static_assert( SIZEOF_ARRAY( PAGE_TITLES ) == NumOverlayPages, "Must match!" );
 
 }
 
--- a/src/RA_AchievementOverlay.cpp
+++ b/src/RA_AchievementOverlay.cpp
@@ -1777,6 +1777,11 @@ API void _RA_RenderOverlay( HDC hDC, RECT* rcSize )
 	g_AchievementOverlay.Render( hDC, rcSize );
 }
 
+API bool _RA_IsOverlayFullyVisible()
+{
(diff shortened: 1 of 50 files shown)


src/RA_MemManager.h
```diff
@@ -19,6 +19,8 @@ class MemCandidate
 	bool m_bHasChanged;
 };
 
+// TODO: We should try to make these function objects, or a function type, not sure the intention here
+
```
Jamiras: These are pointers provided by the emulator to access the memory. They're passed across the DLL boundary as `void*` and cast to `_RAMByteReadFn` and `_RAMByteWriteFn` in `_RA_InstallMemoryBank`.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -827,42 +826,50 @@ std::string Dlg_MemBookmark::ImportDialog()
...
-	IFileOpenDialog* pDlg = nullptr;
-	HRESULT hr = CoCreateInstance( CLSID_FileOpenDialog, NULL, CLSCTX_ALL, IID_IFileOpenDialog, reinterpret_cast<void**>( &pDlg ) );
-	if ( hr == S_OK )
+	// Never noticed this, nearly identical to the on in RA_Core
+	CComPtr<IFileOpenDialog> pDlg;
+
+	// Keep the scope as local as possible
+	
+	if (auto hr = CoCreateInstance(CLSID_FileOpenDialog, nullptr, CLSCTX_ALL, IID_IFileOpenDialog, 
+		reinterpret_cast<void**>(&pDlg)); SUCCEEDED(hr))
```
Jamiras: these changes should be part of #28, not here.

Jamiras: It would be preferable. The changes fit that PR, not this one. If for any reason either of these PR has to be reverted, it keeps the related changes separate.

src/RA_MemManager.h
Jamiras: I'm fine with making them `std::function<>` as long as it works with the DLL contract. It shouldn't be part of this PR.

SyrianBallaS: I know, that's what the <code>TODO:</code> was for. That'll be for another time.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: Well, it would be kinda funky. I could just copy and paste the original here and copy and paste what's here in #28.

SyrianBallaS: OK the original is mostly the way it was.

SyrianBallaS:
Previous concerns have been addressed

src/RA_AchievementOverlay.cpp
```diff
@@ -16,12 +16,14 @@
 
 namespace
 {
-	const float PAGE_TRANSITION_IN = (-0.2f);
-	const float PAGE_TRANSITION_OUT = ( 0.2f);
-	const int NUM_MESSAGES_TO_DRAW = 4;
-	const char* FONT_TO_USE = "Tahoma";
+    // I'm going to guess you never want these changed, simple types can be auto deduced
```
Jamiras: comment is not relevant to code. please remove

SyrianBallaS:
Conflicts are resolved.

SyrianBallaS:
Weird, lets try again.

SyrianBallaS:
Everything seems OK from my end.

SyrianBallaS:
I'll reopen this later.

PR #30, opened 2018-04-28
RapidJSON Update

SyrianBallaS:
This was probably the most involved of all my PRs. The submodule has been updated to match 1.1.0 tag of the RapidJSON repo.


Jamiras:
Please avoid reformatting changes. I scrolled through several hundred lines of code without seeing anything related to RapidJSON. Please resubmit with only the pertinent changes. Not only does it make doing the review easier, it avoids conflicts when merging other PRs, and there's several open right now - mostly yours.

SyrianBallaS:
Got you, I'll reopen this later.

SyrianBallaS:
I think everything should be ok now.

SyrianBallaS:

SyrianBallaS:
Everything's back in order.

SyrianBallaS:
Was so focus on not formatting anything that I forgot to fix some alignements, reopening later.

src/RA_AchievementOverlay.cpp
```diff
@@ -1583,41 +1584,38 @@ void AchievementOverlay::InstallNewsArticlesFromFile()
...
-                nNewsItem.m_sTitle = NextNewsArticle["Title"].GetString();
-                nNewsItem.m_sPayload = NextNewsArticle["Payload"].GetString();
-                nNewsItem.m_sAuthor = NextNewsArticle["Author"].GetString();
-                nNewsItem.m_sLink = NextNewsArticle["Link"].GetString();
-                nNewsItem.m_sImage = NextNewsArticle["Image"].GetString();
-                nNewsItem.m_nPostedAt = NextNewsArticle["TimePosted"].GetUint();
 
-                tm destTime;
-                localtime_s(&destTime, &nNewsItem.m_nPostedAt);
+    std::ifstream ifile{ RA_NEWS_FILENAME, std::ios::binary };
```
SyrianBallaS: lots of these files from the macros look like json files when I open them, is there a reason why they use the .txt extension?

src/RA_AchievementOverlay.cpp
```diff
@@ -1583,41 +1584,38 @@ void AchievementOverlay::InstallNewsArticlesFromFile()
...
+                i["Title"].GetString(),
+                i["Payload"].GetString(),
+                i["TimePosted"].GetUint(),
+                std::to_string(i["TimePosted"].GetUint()), // does the format really matter?
+                i["Author"].GetString(),
+                i["Link"].GetString(),
+                i["Image"].GetString()
+            };
+
+            //tm destTime;
```
SyrianBallaS: Commented out just in case the format does matter, the `to_string `gives the same value as the json response. The timestamp that is.

src/RA_AchievementSet.cpp
```diff
@@ -310,59 +312,62 @@ BOOL AchievementSet::SaveToFile()
...
 
-BOOL AchievementSet::Serialize(FileStream& Stream)
-{
-    //	Why not submit each ach straight to cloud?
-    return FALSE;
 
-    //FILE* pf = nullptr;
-    //const std::string sFilename = GetAchievementSetFilename( m_nGameID );
-    //if( fopen_s( &pf, sFilename.c_str(), "wb" ) == 0 )
+//BOOL AchievementSet::Serialize( FileStream& Stream )
```
SyrianBallaS: Should this function be gone or just marked as deprecated? Since I heard before commented code should not be committed. Should the currently commented out code be gone?

src/RA_AchievementSet.cpp
```diff
@@ -420,148 +423,142 @@ BOOL AchievementSet::LoadFromFile(GameID nGameID)
...
-
-            while (!feof(pFile))
+    if (m_nSetType == Local)
+    {
+        std::string buffer;
+        buffer.reserve(4096); 
+
+        //	Get min ver:
+        ifile.getline(buffer.data(), buffer.capacity());
+        auto nCharsRead{ ifile.gcount() };  //	UNUSED at this point? TBD
```
SyrianBallaS: The read count does appear to be to 0. Another thing I noticed is that a lot of the "ReadTil" functions in RA_Core works exactly like `std::getline`.

src/RA_AchievementSet.cpp
```diff
@@ -420,148 +423,142 @@ BOOL AchievementSet::LoadFromFile(GameID nGameID)
...
 
-                //	Loading Local from file...?
-                g_pCurrentGameData->SetGameTitle(buffer);
-            }
-
-            while (!feof(pFile))
+    if (m_nSetType == Local)
+    {
+        std::string buffer;
+        buffer.reserve(4096);
```
SyrianBallaS: This is required. If you don't think it is try it for yourself. Since we're doing a binary operation on the file we need a pointer, the "capacity" is the same thing as memory in bytes except we don't have to worry about deallocating since `std::string` is managing the pointer for us.

src/RA_AchievementSet.cpp
```diff
@@ -420,148 +423,142 @@ BOOL AchievementSet::LoadFromFile(GameID nGameID)
...
+
+        while (!ifile.eof())
+        {
+            ifile.getline(buffer.data(), buffer.capacity());
+            if (nCharsRead = ifile.gcount(); nCharsRead > 0)
             {
-                ZeroMemory(buffer, 4096);
-                _ReadTil(EndLine, buffer, 4096, &nCharsRead, pFile);
-                if (nCharsRead > 0)
+                buffer = buffer.data();
```
SyrianBallaS: the `buffer = buffer.data(); ` part is required. The string in memory has characters written to it but has not been saved to the object. We get the best of both worlds this way.  The size is 0 despite the capacity not being 0. It's also not nested as a selection statement because it's used by more than one scope, should be obvious but some people just don't know.

If you don't think it's not required try it for yourself.

src/RA_AchievementSet.cpp
```diff
@@ -420,148 +423,142 @@ BOOL AchievementSet::LoadFromFile(GameID nGameID)
...
-                        Achievement& newAch = g_pLocalAchievements->AddAchievement();
-                        buffer[nCharsRead - 1] = '\0';	//	Turn that endline into an end-string
-
-                        newAch.ParseLine(buffer);
-                    }
+                    //g_LeaderboardManager.AddLeaderboard(lb);
+                }
+                else if (isdigit(buffer.front()))
+                {
+                    Achievement& newAch = g_pLocalAchievements->AddAchievement();
```
SyrianBallaS: If achievements are ID'd copying should be deleted (also meaning they can't be used as pointers but can be cheaply passed by value), we should consider using `unordered_map `if you prefer speed over order or `map `if you prefer order over speed. We can easily use` quick sort` from `std::algorithm` to so sort it if needed. This type of copying technically works even with deleted copying since you are copying a reference (which the object at that address is always const), otherwise you should just using move semantics.

src/RA_CodeNotes.cpp
```diff
@@ -20,35 +22,39 @@ size_t CodeNotes::Load(const std::string& sFile)
...
-                const std::string& sAuthor = NextNote["User"].GetString();	//	Author?
+            const std::string& sAddr = i["Address"].GetString();
+            auto nAddr = static_cast<ByteAddress>(std::strtoul(sAddr.c_str(), nullptr, 16));
 
-                m_CodeNotes.insert(std::map<ByteAddress, CodeNoteObj>::value_type(nAddr, CodeNoteObj(sAuthor, sNote)));
-            }
+            // Author?
+            const std::string& sAuthor = i["User"].GetString();
+
+            auto code_pair{ std::make_pair(nAddr, CodeNoteObj{ sAuthor, sNote }) };
```
SyrianBallaS: Your CodeNoteObj would need a move constructor and assignment for this to work properly. Instead of making the "ID" const just delete copying, that way you enforce uniqueness making duplication impossible. If you tried to define moving explicitly the program would not compile. Any way if you define and/or delete any default operation you need to define and/or delete them all.

src/RA_CodeNotes.cpp
```diff
@@ -20,35 +22,39 @@ size_t CodeNotes::Load(const std::string& sFile)
...
+        const auto& NoteArray = doc["CodeNotes"];
 
-            for (SizeType i = 0; i < NoteArray.Size(); ++i)
-            {
-                const Value& NextNote = NoteArray[i];
-                if (NextNote["Note"].IsNull())
-                    continue;
+        for (auto& i : NoteArray.GetArray())
+        {
+            if (i["Note"].IsNull())
```
SyrianBallaS: It might be more appropriate to use to an Expects/Ensures Contract to make sure it's not null. It's not in the standard yet but Microsoft has done a pretty good job at implemented Bjarne Stroustrup's C++ Core Guidelines.

src/RA_CodeNotes.cpp
```diff
@@ -20,35 +22,39 @@ size_t CodeNotes::Load(const std::string& sFile)
...
+
+    if (!doc.HasParseError())
+
     {
-        Document doc;
-        doc.ParseStream(FileStream(pf));
-        if (!doc.HasParseError())
-        {
-            ASSERT(doc["CodeNotes"].IsArray());
+        ASSERT(doc["CodeNotes"].IsArray());
```
SyrianBallaS: You know that assert gets compiled out in release mode right? It might be more appropriate to throw an exception or use a conditional depending on whether it's debug mode or not

src/RA_Core.cpp
```diff
@@ -872,115 +872,117 @@ API void CCONV _RA_LoadPreferences()
     RA_LOG(__FUNCTION__ " - loading preferences...\n");
 
     SetCurrentDirectory(NativeStr(g_sHomeDir).c_str());
-    FILE* pf = nullptr;
-    fopen_s(&pf, std::string(std::string(RA_PREFERENCES_FILENAME_PREFIX) + g_sClientName + ".cfg").c_str(), "rb");
-    if (pf == nullptr)
+    std::ifstream ifile{ ra::PrefsFilename(), std::ios::binary };
+
+#pragma region Comments
+    /*if (pf == nullptr)
```
SyrianBallaS: This situation isn't possible, most of it was commented out already. Deleted my prefs file and it came back.

src/RA_Core.cpp
```diff
@@ -994,52 +996,49 @@ API void CCONV _RA_SavePreferences()
...
+    doc.AddMember("Leaderboard Counter Display", g_bLBDisplayCounter, a);
+    doc.AddMember("Leaderboard Scoreboard Display", g_bLBDisplayScoreboard, a);
+    doc.AddMember("Num Background Threads", g_nNumHTTPThreads, a);
+    doc.AddMember("ROM Directory", StringRef(g_sROMDirLocation.c_str()), a);
 
-        fclose(pf);
+    Value positions(kObjectType);
+
+    for (auto& i : g_mWindowPositions)
+    {
```
SyrianBallaS: We should be using ranged fors as much as possible (existed since C++11) if you don't need to advance or reverse the iterators. There's also a VS Extension `for each` which works exactly like the .NET `foreach `but has the same semantics as a ranged for loop. there's also `std::for_each` but not totally sure if you guys would be comfortable using `lambdas`. Of course sometimes it's not possible.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -774,50 +776,47 @@ void Dlg_MemBookmark::ExportJSON()
...
-                ClearAllBookmarks();
+                wchar_t buffer[256];
+                swprintf_s(buffer, 256, L"%s", Widen(i["Description"].GetString()).c_str());
 
-                const Value& BookmarksData = doc["Bookmarks"];
-                for (SizeType i = 0; i < BookmarksData.Size(); ++i)
-                {
-                    MemBookmark* NewBookmark = new MemBookmark();
+                auto NewBookmark = std::make_unique<MemBookmark>();
+                NewBookmark->SetDescription(buffer);
```
SyrianBallaS: Originally did not use delete with new, replaced it with `unique_ptr`. `unique_ptr` follows RAII exactly so it destroys the pointer it owns when it goes out of scope. Unlike CComPtr `release `does not deallocate, it just releases ownership. Its much better to use smart pointers because they have many checks to let you know about any potential problems at compile time (the standard ones at least). Raw pointers just makes apps fail silently or hang in release mode most of the time.

SyrianBallaS:
(reopen later)

RA_Integration.sln
```diff
@@ -38,4 +38,7 @@ Global
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {220AB67D-0EDC-4B55-B737-54BDB6CDACDD}
 	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {DD1A0623-724F-4AA3-B594-07F2ACDCBA51}
+	EndGlobalSection
```
SyrianBallaS: I have no idea where this came from.

src/RA_AchievementSet.cpp
```diff
@@ -311,7 +312,7 @@ BOOL AchievementSet::SaveToFile()
     //}
 }
 
-BOOL AchievementSet::Serialize(FileStream& Stream)
+BOOL AchievementSet::Serialize(_UNUSED std::iostream& Stream)
```
SyrianBallaS: New rapidjson has FileRead/WriteStream, the iostream wrappers should be fast if you don't use `std::endl`. 
`std::ios::sync_with_stdio(false)` might help if using stuff like `cout`.

I just put iostream and marked the function as deprecated since it's not used anymore.

src/RA_AchievementSet.cpp
```diff
@@ -421,38 +421,43 @@ BOOL AchievementSet::LoadFromFile(GameID nGameID)
...
-    errno_t nErr = fopen_s(&pFile, sFilename.c_str(), "r");
-    if (pFile != nullptr)
+
+    std::ios::sync_with_stdio(false);
+
+    if (std::ifstream ifile{ sFilename }; !ifile.is_open())
+    {
+        //	Cannot open file
+        RA_LOG("Cannot open file %s\n", sFilename.c_str());
+        RA_LOG("Error %s\n", ra::GetLastErrorMsg());
```
SyrianBallaS: Windows kept on telling me me "Cannot create a file that already exists". This had no affect on core achievements, once I put a local achievement in it stopped giving an error.

src/RA_AchievementSet.cpp
```diff
@@ -421,38 +421,43 @@ BOOL AchievementSet::LoadFromFile(GameID nGameID)
...
     {
         //	Store this: we are now assuming this is the correct checksum if we have a file for it
         g_pCurrentGameData->SetGameID(nGameID);
 
         if (m_nSetType == Local)
         {
-            const char EndLine = '\n';
-            char buffer[4096];
-            unsigned long nCharsRead = 0;
+            auto buffer{ std::make_unique<char[]>(4096) };
```
SyrianBallaS: If you care about size (which is the only reason I can think of for using C arrays) then prefer this approach.
If you don't just just `std::basic_string `and use `data()` when you want to write to it.

src/RA_AchievementSet.cpp
```diff
@@ -421,38 +421,43 @@ BOOL AchievementSet::LoadFromFile(GameID nGameID)
...
 
             //	Get min ver:
-            _ReadTil(EndLine, buffer, 4096, &nCharsRead, pFile);
+            ifile.getline(buffer.get(), 4096);
+            auto nCharsRead{ ifile.gcount() };
```
SyrianBallaS: this came out as 0 for me as well.

src/RA_AchievementSet.cpp
```diff
@@ -421,38 +421,43 @@ BOOL AchievementSet::LoadFromFile(GameID nGameID)
...
+                g_pCurrentGameData->SetGameTitle(buffer.get());
 
-            while (!feof(pFile))
+            while (!ifile.eof())
             {
-                ZeroMemory(buffer, 4096);
-                _ReadTil(EndLine, buffer, 4096, &nCharsRead, pFile);
-                if (nCharsRead > 0)
+                buffer.reset();
+                buffer = std::make_unique<char[]>(4096);
```
SyrianBallaS: A safer way of doing `ZeroMemory`

src/RA_AchievementSet.h
```diff
@@ -27,7 +27,8 @@ class AchievementSet
     void Clear();
     void Test();
 
-    BOOL Serialize(FileStream& Stream);
+    _DEPRECATEDR("Why not submit each ach straight to cloud?")
+    BOOL Serialize(_UNUSED std::iostream& Stream);
```
SyrianBallaS: Presently, the compiler won't give you warning. If for some reason someone did use a function marked as deprecated, the compiler will warning that you are using a deprecated function.

src/RA_CodeNotes.cpp
```diff
@@ -17,35 +19,39 @@ size_t CodeNotes::Load(const std::string& sFile)
...
-                const std::string& sAddr = NextNote["Address"].GetString();
-                ByteAddress nAddr = static_cast<ByteAddress>(std::strtoul(sAddr.c_str(), nullptr, 16));
-                const std::string& sAuthor = NextNote["User"].GetString();	//	Author?
+            // Author?
+            const std::string& sAuthor = i["User"].GetString();
 
-                m_CodeNotes.insert(std::map<ByteAddress, CodeNoteObj>::value_type(nAddr, CodeNoteObj(sAuthor, sNote)));
-            }
+            auto code_pair{ std::make_pair(nAddr, CodeNoteObj{ sAuthor, sNote }) };
+            m_CodeNotes.insert(code_pair);
```
SyrianBallaS: TODO: Make proper move semantics for rvalue refs. Using const members with rvalues dangles the ref.

src/RA_Core.cpp
```diff
@@ -874,115 +875,117 @@ API void CCONV _RA_LoadPreferences()
...
-                        if (iter->value.HasMember("Height"))
-                            pos.nHeight = iter->value["Height"].GetInt();
-                    }
+                    auto& pos = g_mWindowPositions[iter->name.GetString()];
+
+                    pos.nLeft = pos.nTop = pos.nWidth = pos.nHeight = WindowPosition::nUnset;
+                    pos.bLoaded = false;
+
+                    if (iter->value.HasMember("X"))
+                        pos.nLeft = iter->value["X"].GetInt();
```
SyrianBallaS: Because of the need for `name `and `value`, this was not made as a ranged for.

src/RA_Core.cpp
```diff
@@ -1690,3 +1688,23 @@ BOOL CanCausePause()
...
     return (_RA_CausePause != nullptr);
 }
+
+namespace ra {
+_Use_decl_annotations_
+std::string GameJSONFilename(GameID game_id) noexcept
+{
+    std::ostringstream oss;
+    oss << RA_DIR_DATA << game_id << ".txt";
+    return oss.str();
```
SyrianBallaS: With tiny functions like this, `stringstream `doesn't seem that bad but I've been a little hesitant to do it for larger functions.

src/ra_errors.cpp
```diff
@@ -0,0 +1,65 @@
...
+bool CALLBACK InvalidModule() noexcept { return !g_hThisDLLInst; }
+
+int CALLBACK ShowLastError() noexcept
+{
+    // real quick workaround I got something in another branch -Samer
+#if _MBCS
+    auto myErrorCode{ std::to_string(::GetLastError()) };
+#elif _UNICODE
+    auto myErrorCode{ std::to_wstring(::GetLastError()) };
+#else
```
SyrianBallaS: That other header was `ra_utility` but it had stuff that we didn't need here

src/ra_errors.h
```diff
@@ -0,0 +1,42 @@
...
+
+_EXTERN_C
+
+_Success_(return != 0)
+int CALLBACK ShowError(_In_ const tstring& message, _In_ HWND hwnd = ::GetActiveWindow()) noexcept;
+_NODISCARD bool CALLBACK InvalidModule() noexcept;
+
+// Shows the last system error with a message without throwing
+int CALLBACK ShowLastError() noexcept;
+_END_EXTERN_C
```
SyrianBallaS: Some of these aren't marked with `[[nodiscard]]` because I felt it wasn't necessary to absolutely use them.
Also a lot of these functions I put use `__stdcall `mostly because of calls to the Windows API and it's similar to `__thiscall `(calling convention for classes)

src/RA_Condition.h
```diff
@@ -26,7 +26,7 @@ extern const char* COMPARISONVARIABLESIZE_STR[];
 
 enum ComparisonVariableType
 {
-    Address,			//	compare to the value of a live address in RAM
+    Address,			//	compare to the  of a live address in RAM
```
Jamiras: please put this back.

src/RA_Core.cpp
Jamiras: There are several things that generate the prefs file. A successful login is one. Regardless, this functionality is changed with the IConfiguration service in PR 66.

src/RA_Core.cpp
Jamiras: why do you think `stringstream` is bad?

src/RA_Dlg_MemBookmark.cpp
Jamiras: I don't think this is going to work without additional changes. `AddBookmark` and `AddBookmarkMap` both store the raw pointer, which will be deallocated when `NewBookmark` goes out of scope.

`m_vBookmarks` should be a vector of `unique_ptr` - it's the authoritative collection of bookmarks. `m_BookmarkMap` is just a convenience mapper for accessing items in `m_vBookmarks`. In my opinion, the number of bookmarks will never be large enough to require a convenience mapper.

src/RA_httpthread.cpp
```diff
@@ -388,6 +388,8 @@ BOOL RAWeb::DoBlockingHttpGet(const std::string& sRequestedPage, std::string& Re
 
                     while (nBytesToRead > 0)
                     {
+
+                        // TODO: Replace pData's type with DataStream after PR #23 has been merged
```
Jamiras: PR 23 has been merged. please address or remove this comment.

SyrianBallaS:
reopening later

src/RA_Condition.h
SyrianBallaS: Don't know how that happened, will do.

.gitignore
```diff
@@ -4,6 +4,10 @@
 /src/RA_BuildVer.h
 *.user
 /RetroAchievementsKey
-test_app/
 src/*_REMOTE_*.cpp
+src/*_BACKUP_*.cpp
+src/*_BASE_*.cpp
+src/*_REMOTE_*.cpp
```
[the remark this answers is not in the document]
SyrianBallaS: sure thing. It will go away.

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -92,99 +93,91 @@ Dlg_GameLibrary::~Dlg_GameLibrary()
...
-                    sstr << bufPct;
-                }
+            std::stringstream sstr;
+            sstr << nEarned;
+            if (nEarnedHardcore > 0)
+                sstr << " (" << std::to_string(nEarnedHardcore) << ")";
 
-                GameProgressOut[nID] = sstr.str();
+
+            if (sstr << " / " << nNumAchievements; nNumAchievements > 0)
```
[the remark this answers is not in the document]
SyrianBallaS: OK, that's fine. You make a good point.

src/RA_Core.cpp
SyrianBallaS: Sure, I need to merge the new stuff anyway.

src/RA_Core.cpp
SyrianBallaS: This is really strange, it must have been a bug in VC++ because I can't seem to reproduce what I had encountered last time. It's fine I guess.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: I would have to disagree, the pointer lives long enough to be inserted in both collections and is deallocated when no longer needed. 
For now I'll put it back and use delete and probably refactor this class entirely in a different PR. While debugging it gives nothing but garbage. I don't think the mapper is needed either.
I can make it work without it being a pointer, but some other work would have to be done to ensure uniqueness.

src/RA_httpthread.cpp
SyrianBallaS: Sure.

src/RA_AchievementOverlay.cpp
SyrianBallaS: Probably

SyrianBallaS:
Implemented feedback.

Very strange how freeing dynamic memory in the Memory Bookmarker causes an access violation. STL containers weren't meant to be used with pointers as value types anyway. The best approach I think would be to delete copying, then there would be no need for pointers Well that's for another PR.

src/RA_AchievementSet.cpp
Jamiras: ifstream will set errno, not GetLastErrror, the "file already exists" error must be from something else.

src/RA_AchievementSet.cpp
Jamiras: buffer should only be allocated once outside the loop.

Jamiras: The `ZeroMemory` is probably unnecessary. The `nCharsRead` will prevent using stale data.

src/RA_AchievementSet.cpp
```diff
@@ -474,79 +411,72 @@ BOOL AchievementSet::LoadFromFile(ra::GameID nGameID)
...
                 g_pCurrentGameData->ParseData(doc);
-                ra::GameID nGameID = g_pCurrentGameData->GetGameID();
+                auto nGameID = g_pCurrentGameData->GetGameID();
 
                 RA_RichPresenceInterpretter::PersistAndParseScript(nGameID, g_pCurrentGameData->RichPresencePatch());
 
-                const Value& AchievementsData = doc["Achievements"];
-                for (SizeType i = 0; i < AchievementsData.Size(); ++i)
+                const auto& AchievementsData = doc["Achievements"];
+                for (auto& i : AchievementsData.GetArray())
```
Jamiras: `i` is no longer an appropriate variable name. Suggest `achievement` or `ach`

src/RA_CodeNotes.cpp
```diff
@@ -17,45 +19,44 @@ size_t CodeNotes::Load(const std::string& sFile)
...
 
-            const Value& NoteArray = doc["CodeNotes"];
+        const auto& NoteArray = doc["CodeNotes"];
 
-            for (SizeType i = 0; i < NoteArray.Size(); ++i)
-            {
-                const Value& NextNote = NoteArray[i];
-                if (NextNote["Note"].IsNull())
-                    continue;
+        for (auto& i : NoteArray.GetArray())
```
Jamiras: suggest `note` instead of `i`

src/RA_AchievementSet.cpp
Jamiras: also, `%s` requires you `.c_str()` the `tstring`.

rzumer:
To guarantee some measure of stability, I would prefer sticking to the latest stable version of the library. However, version 1.1 was released on 2016. Have you looked at that version? Is it compatible with the API changes made here? It is possible that the current master version is less buggy than 1.1 considering the hundreds of open issues, but despite being in active development the 1.2 Beta milestone has yet to be reached. Whatever version you settle with must be as stable as possible.

SyrianBallaS:
This uses the master branch of the rapidjson repo as a submodule. I was
going to go the nuget route but most of the people I talked about it with
were against it. It's the same library the original version used here was a
beta.

rzumer:
Using different versions for development and release (e.g. master and v1.1) does not seem like a good idea to me, and updating the submodule to a bleeding edge version constantly is also not sustainable, which is why I suggest setting the submodule version to the latest release tag.

The current master version sits between v1.1 and v1.2 beta. If you plan to release with v1.1, then it makes more sense to roll back to f54b0e4 and leave it at that state until version 1.2 is stable.

SyrianBallaS:
sure that's fine I can do that.

src/RA_Core.cpp
```diff
@@ -1238,10 +1239,18 @@ API void CCONV _RA_InvokeDialog(LPARAM nID)
             break;
 
         case IDM_RA_FILES_ACHIEVEMENTEDITOR:
-            if (g_AchievementEditorDialog.GetHWND() == nullptr)
+            if (::FindWindow(nullptr, _T("Achievement Editor")) == nullptr)
```
Jamiras: why was this change made? does it need to be applied to line 1252 as well?

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -654,14 +654,17 @@ void Dlg_MemBookmark::WriteFrozenValue(const MemBookmark & Bookmark)
     char buffer[32];
     sprintf_s(buffer, sizeof(buffer), "%0*x", width, Bookmark.Value());
 
-    for (unsigned int i = 0; i < strlen(buffer); i++)
+    if(std::strlen(buffer) > 0)
```
Jamiras: why would buffer be empty? if this is necessary, suggest "if (*buffer)", we don't need to iterate the entire string.

src/RA_AchievementSet.cpp
```diff
@@ -431,38 +364,39 @@ BOOL AchievementSet::LoadFromFile(ra::GameID nGameID)
...
     SetCurrentDirectory(NativeStr(g_sHomeDir).c_str());
-    FILE* pFile = nullptr;
-    errno_t nErr = fopen_s(&pFile, sFilename.c_str(), "r");
-    if (pFile != nullptr)
+
+    if (std::ifstream ifile{ sFilename }; !ifile.is_open())
+    {
+        //	Cannot open file
+        RA_LOG("Cannot open file %s\n", sFilename.c_str());
+        RA_LOG("Error %s\n", NativeStr(ra::GetLastErrorMsg()).c_str());
```
Jamiras: not sure where my previous comment went, but ifsteam sets errno, so calling GetLastError won't return the expected message.

Jamiras: also RA_LOG expects MBCS strings. `NativeStr` should be replaced with `Narrow`

SyrianBallaS: Well, sometimes it returns "Cannot create a file that already exists", or "The file is in use by another process". So it seems legit to me.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: Sometimes it came out at nullptr when testing file i/o for local achievements.

src/RA_Core.cpp
SyrianBallaS: Well I can change it back it you want it that way. I was trying to limit needing Globals and use an API call instead.

src/RA_AchievementSet.cpp
Jamiras: That's leftover from somewhere else:
```
    RA_LOG("Non-Error %s\n", NativeStr(ra::GetLastErrorMsg()).c_str());

    if (std::ifstream ifile{ sFilename }; !ifile.is_open())
    {
        //	Cannot open file
        RA_LOG("Cannot open file %s\n", sFilename.c_str());
        RA_LOG("Error %s\n", NativeStr(ra::GetLastErrorMsg()).c_str());
        return FALSE;
    }
```
```
Non-Error Cannot create a file when that file already exists.
Cannot open file .\RACache\Data\1466-User.txt
Error Cannot create a file when that file already exists.
```

src/RA_Dlg_MemBookmark.cpp
Jamiras: buffer can't be nullptr as it's an array on the stack. if it contains nullptr, that means the sprintf_s failed. suggest checking the return value of sprintf_s vs running strlen.

src/RA_Core.cpp
Jamiras: FindWindow is an API call that has some amount of overhead for iterating through the windows on the system and doing strcmps on their titles. the previous implementation (albeit a global) was a fixed integer reference and required no overhead.

This code is triggered by selecting a menu option, so performance probably isn't key, but this change doesn't feel like an improvement and definitely isn't related to changing the RapidJSON version.

I'm fine with the pre-populating an achievement title, but you do _four_ ::FindWindow calls to get there.

src/RA_Dlg_MemBookmark.cpp
Jamiras: it may not even contain nullptr in a release build (uninitialized memory in buffer). definitely recommend checking the return value.

src/RA_Defs.h
```diff
@@ -87,26 +87,24 @@
...
-
-//	RA-Only
-#define RAPIDJSON_HAS_STDSTRING 1
-#pragma warning(push, 1)
-// This is not needed the most recent version
-#define _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS
 //	RA-Only
+#pragma warning(push)
+// Seems version 1.1.0 still uses deprecated stuff.
+#pragma warning(disable : 4996) // STL4015 deprecation: uses iterator as a base class instead of usings from iterator_traits.
```
Jamiras: the documented way to handle this is with `#define RAPIDJSON_NOMEMBERITERATORCLASS`

SyrianBallaS:
Triple checked that the default achievement title doesn't overwrite existing ones, and checked the error message under various circumstances. Applied the "limit globals" style that code block in InvokeDialog. For it to be completely "globaless" we need a stateless controller so we can use it locally to route creation and/or messages, but that's for another PR.

src/RA_Defs.h
```diff
@@ -87,26 +87,24 @@
...
+// NOMINMAX doesn't seem to have an effect on rapidjson
```
Jamiras: It does, but only for those files that include "RA_Defs.h" before any other windows header files. I had to add `NOMINMAX` to the following files to eliminate all of the warnings:
* RA_Dlg_Login.h
* RA_Dlg_RichPresence.h
* RA_Dlg_RomChecksum.h
* RA_ImageFactory.h
* RA_Errors.h

I just assumed you weren't seeing the warnings, because I also had to add "#include <memory>" to RA_AchievementOverlay.cpp, RA_AchievementSet.cpp and RA_GameData.h to even get this PR to compile. Whatever you're doing to define "std::unique_ptr" and "std::make_unique" isn't happening on my machine.

src/RA_Core.cpp
SyrianBallaS: alright, I'm gonna close this and branch out.

src/RA_Defs.h
SyrianBallaS: thanks.

SyrianBallaS:
Thanks a lot @Jamiras. You're feedback is very helpful. Gonna reopen this later.

SyrianBallaS:
Alright this should be ready, gonna leave some review comments.

execution_times.txt
```diff
@@ -0,0 +1,54 @@
+This file contains execution times of some functions to see what we should limit so the test code doesn't bloat
```
rzumer: We don't need to keep this file in the repository. Feel free to paste the contents in a PR comment, this way when referring back to your commits it will remain as a reference.

RA_Integration.sln
```diff
@@ -12,6 +12,11 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "RA_Integration.Tests", "tests\RA_Integration.Tests.vcxproj", "{876DEBDF-C7C8-4713-8CEF-EBDEE65306A7}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Notes", "Notes", "{630B7310-9EF8-4566-B9E8-06EEFEB2384F}"
```
rzumer: This is out of place in the code.

execution_times.txt
SyrianBallaS: Alright sure

RA_Integration.sln
SyrianBallaS: I'll just get rid of it

SyrianBallaS:
Got rid of the `execution_times.txt ` file and put it outside the repository. Basically it convinced me that `FindWindow `should be avoided.

ScottFromDerby:
There's a lot here that pushes the RA repo in a C++11/C++14 direction, which would be brilliant to be able to take advantage of the new features that are becoming available (auto, lambdas, delegates etc)

One of the focuses of RA is to ensure compatability. In the case of RetroArch, this means targeting C89 (!) in order to become compatible with as wide a variety of systems and integrations as possible.

This is something I want to aim towards, especially with respect to achievement parsing. This isn't necessarily something that needs to filter through to everything in the toolset (MFC itself didn't even exist in 1989, let alone most of Windows) but it would be lush to have RetroArch just pull in the integration as a submodule etc. and reference the necessary files. This is the ultimate goal IMO.

SyrianBallaS:
@ScottFromDerby 
I totally understand the focus of ensuring compatibility. Usually the main approach of using C++ functions in C is through a callback. Since you probably want it as a sub-module some of the code would have to be C89 compliant. I heard @leiradel is working on something like that. If there's something other than what he's doing you want written in C, I'm all for it.

Jamiras:
> it would be lush to have RetroArch just pull in the integration as a submodule etc. and reference the necessary files
We are doing exactly this with the rcheevos repository.

.gitignore
```diff
@@ -2,12 +2,17 @@
...
 src/clipboard.txt
 
 # Output files
 /obj
 /bin
 *.pdb
 *.dll
+
+# Misc. Folders
+Error lists/
```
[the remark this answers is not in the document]
Jamiras: If it's not an artifact of the build or the build tools, the repository shouldn't know about it. Consider putting your personal files in a directory outside the repository.

src/RA_Achievement.cpp
```diff
@@ -152,14 +153,17 @@ BOOL Achievement::Test()
 
 void Achievement::Clear()
 {
-    m_vConditions.Clear();
+    if(!std::empty(m_vConditions))
+        m_vConditions.Clear();
```
Jamiras: the check is redundant. Clear won't do anything if it's already empty.

Jamiras: similar comment applies to later lines.

src/RA_Achievement.cpp
```diff
@@ -170,9 +174,14 @@ void Achievement::Clear()
...
-    m_sProgressMax[0] = '\0';
-    m_sProgressFmt[0] = '\0';
+
+    if (!std::empty(m_sProgress))
+        m_sProgressMax.clear();
+    if(!std::empty(m_sProgressMax))
+        m_sProgressMax.clear();
+    if (!std::empty(m_sProgressFmt))
+        m_sProgressMax.clear();
+
```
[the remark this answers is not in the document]
Jamiras: Please focus this PR on just updating the rapidjson version. I provide you a couple lines of feedback for things to address and you respond with completely new changes i have to look at.

src/RA_Achievement.h
```diff
@@ -52,7 +50,7 @@ class Achievement
     inline BOOL GetPauseOnReset() const { return m_bPauseOnReset; }
     void SetPauseOnReset(BOOL bPause) { m_bPauseOnReset = bPause; }
 
-    BOOL IsCoreAchievement() const { return m_nSetType == Core; }
+    BOOL IsCoreAchievement() const { return m_nSetType == AchievementSetType::Core; }
```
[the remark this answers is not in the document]
Jamiras: Good change in the long run. Not expected in this PR. Removing the using for the rapidjson namespace is at least related.

src/RA_AchievementSet.cpp
```diff
@@ -431,38 +365,43 @@ BOOL AchievementSet::LoadFromFile(ra::GameID nGameID)
...
+
+    if (std::ifstream ifile{ sFilename }; !ifile.is_open())
+    {
+        //	Cannot open file
+        RA_LOG("Cannot open file %s\n", sFilename.c_str());
+        auto errBuf{ std::make_unique<char[]>(1024) };
+        strerror_s(errBuf.get(), 1024, errno);
+        RA_LOG("Error %s\n", errBuf.get());
+        ra::ShowError(errBuf.get());
+        return FALSE;
```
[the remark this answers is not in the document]
Jamiras: Curious why you chose to create a GetLastErrorMessage function, but left the errno conversion in local code.

src/RA_AchievementSet.cpp
Jamiras: This code will have to change to address #82

src/RA_Core.cpp
```diff
@@ -30,11 +30,14 @@
 #include "RA_Dlg_MemBookmark.h"
 
 #include <locale>
+#include <chrono> // for testing execution time
```
Jamiras: is it still needed? hopefully you didn't commit the execution time tests.

src/RA_Defs.h
```diff
@@ -53,6 +53,22 @@
...
+#define _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES_COUNT 1
+#endif // !_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES_COUNT
+
+#ifndef _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES
+#define _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES 1
+#endif // !_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES
+
+#ifndef _CRT_SECURE_CPP_OVERLOAD_SECURE_NAMES
+#define _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES 1
+#endif // !_CRT_SECURE_CPP_OVERLOAD_SECURE_NAMES
```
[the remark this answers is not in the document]
Jamiras: no one should be using `sprintf`. `snprintf` is a viable alternative.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -1,13 +1,16 @@
 #include "RA_Dlg_MemBookmark.h"
 
 #include <atlbase.h> // CComPtr
+#include <chrono> // execution time testing
```
Jamiras: please remove if no longer needed

src/RA_AchievementSet.cpp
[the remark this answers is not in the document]
SyrianBallaS: It helps me visualize some errors that are happening. If it wasn't already there it was removed. It seems like the message is only accurate from WinAPI functions. At first I thought it was required to have a static buffer for errorstr_s but then remembered in MSVC you need to put the buffer size if it's dynamic.

I have a few "experimental" branches with lots of stuff but my PRs only have a small portion of it to make it easier to manage.

src/ra_fwd.h
```diff
@@ -31,6 +28,9 @@ using AchievementID = std::size_t;
 using LeaderboardID = std::size_t;
 using GameID        = std::size_t;
 
+using HANDLE = void*;
+using LPVOID = void*;
```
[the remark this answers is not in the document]
rzumer: That sounds like it will obscure errors on unsupported platforms rather than help identify them.

src/RA_Achievement.cpp
```diff
@@ -159,7 +160,9 @@ void Achievement::Clear()
     m_sTitle.clear();
     m_sDescription.clear();
     m_sAuthor.clear();
-    m_sBadgeImageURI.clear();
+
+    if(!m_sBadgeImageURI.empty())
+        m_sBadgeImageURI.clear();
```
Jamiras: missed one.

src/RA_AchievementSet.cpp
```diff
@@ -431,38 +364,42 @@ BOOL AchievementSet::LoadFromFile(ra::GameID nGameID)
...
+        if (m_nSetType == AchievementSetType::Local)
         {
-            const char EndLine = '\n';
-            char buffer[4096];
-            unsigned long nCharsRead = 0;
+            auto buffer{ std::make_unique<char[]>(2048) };
 
             //	Get min ver:
-            _ReadTil(EndLine, buffer, 4096, &nCharsRead, pFile);
+            ifile.getline(buffer.get(), 2048);
```
Jamiras: OMG, why!?

I indicate this code will have to change for #82 in response to some comment you had about using make_unique to allocate a buffer and you interpret that as the buffer was too large? Did you even read #82? The problem is that the line was too big to fit in the buffer, so a smaller buffer is not the solution, nor should you be trying to fix it in this PR.

SyrianBallaS: I did read the issue and misunderstood. I will consult with you if I'm not sure what you mean.

src/ra_fwd.h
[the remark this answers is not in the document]
SyrianBallaS: I changed it somewhat to how it was before but only aliased the type if it didn't exist. These were originally typedefs in RA_httpthread.h

src/ra_fwd.h
```diff
@@ -5,25 +5,36 @@
...
 
+#ifndef _WINDEF_
+using DWORD  = unsigned long;
+using LPVOID = void*;
+#endif // !_WINDEF_
 
+#ifndef _WINNT_
+using HANDLE = void*;
+#endif // !_WINNT_
+
```
SyrianBallaS: Depending on where this file is included will determine if they need to be aliased or not to prevent a "redefinition".

src/ra_fwd.h
```diff
@@ -5,25 +5,37 @@
...
-
-namespace ra {
-
 #ifndef _TCHAR_DEFINED
 #if _MBCS
-using TCHAR = char;
+using TCHAR  = char;
+using PTCHAR = char*;
+using TBYTE  = unsigned char;
+using PTBYTE = unsigned char*;
```
Jamiras: BYTE is always unsigned char (8-bits). Why do we need a TBYTE?

src/RA_PopupWindows.h
```diff
@@ -2,6 +2,8 @@
 #define RA_POPUPWINDOWS_H
 #pragma once
 
+#define NOMINMAX
```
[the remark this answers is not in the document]
SyrianBallaS: It showed up while building. There was a warning saying "while compiling RA_PopupWindows.cpp". There was a warning "while compiling RA_ProgressPopup.cpp" as well". Thank you.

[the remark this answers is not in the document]
Jamiras: Maybe, but this isn't the correct solution. One of the other headers included by those CPP files is including a windows header file, and that's where you need to put the `NOMINMAX`.

src/ra_fwd.h
SyrianBallaS: A precaution. In WinNT.h.
<code>
#ifndef _TCHAR_DEFINED
typedef char TCHAR, *PTCHAR;
typedef unsigned char TBYTE , *PTBYTE ;
#define _TCHAR_DEFINED
#endif /* !_TCHAR_DEFINED */
<code>

<code>
#ifdef  UNICODE                     // r_winnt

#ifndef _TCHAR_DEFINED
typedef WCHAR TCHAR, *PTCHAR;
typedef WCHAR TBYTE , *PTBYTE ;
#define _TCHAR_DEFINED
#endif /* !_TCHAR_DEFINED */
...
<code>

src/ra_errors.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace ra {
+
+using tstring = std::basic_string<TCHAR>;
+
+_EXTERN_C
+
+_Success_(return != 0)
+int CALLBACK ShowError(_In_ const tstring& message, _In_opt_ HWND hwnd = nullptr) noexcept;
+_NODISCARD bool CALLBACK InvalidModule() noexcept;
```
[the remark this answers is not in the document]
SyrianBallaS: It's left over from another branch when testing ownership with HINSTANCE/HMODULE. Mainly in DllMain. I'll tell you what, If something like that get's left over I'll check if it's referenced, if not it'll get removed.

But really, there are a lot of functions (at least 50, most of them being getters and setters) in this project that are not referenced so I didn't think it was a big deal. If you want to find them, compile with /Wall. So why is it only mine that matter? You'd need to look for "unreferenced function has been removed".

src/RA_LeaderboardManager.h
```diff
@@ -4,13 +4,21 @@
...
+#endif // RA_EXPORTS
+
 #include <vector>
 
 class RA_LeaderboardManager
 {
 public:
+#if RA_EXPORTS
     static void OnSubmitEntry(const rapidjson::Document& doc);
+#endif // RA_EXPORTS
```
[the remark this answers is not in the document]
SyrianBallaS: Thanks I got you.

src/RA_GameData.h
```diff
@@ -2,7 +2,14 @@
 #define RA_GAMEDATA_H
 #pragma once
 
-#include "RA_Defs.h"
+#if RA_EXPORTS
+#define RAPIDJSON_NOMEMBERITERATORCLASS
+#define RAPIDJSON_HAS_STDSTRING 1
+#include <rapidjson/document.h>
+#endif // RA_EXPORTS
```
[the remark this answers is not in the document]
SyrianBallaS: OK that make makes sense.

src/RA_Core.cpp
```diff
@@ -1238,11 +1239,23 @@ API void CCONV _RA_InvokeDialog(LPARAM nID)
...
 
         case IDM_RA_FILES_ACHIEVEMENTEDITOR:
+        {
             if (g_AchievementEditorDialog.GetHWND() == nullptr)
-                g_AchievementEditorDialog.InstallHWND(CreateDialog(g_hThisDLLInst, MAKEINTRESOURCE(IDD_RA_ACHIEVEMENTEDITOR), g_RAMainWnd, g_AchievementEditorDialog.s_AchievementEditorProc));
+            {
+                g_AchievementEditorDialog.InstallHWND(CreateDialog(::GetModuleHandle(_T("RA_Integration.dll")),
+                                                      MAKEINTRESOURCE(IDD_RA_ACHIEVEMENTEDITOR),
+                                                      ::GetAncestor(::GetActiveWindow(), GA_ROOT),
+                                                      g_AchievementEditorDialog.s_AchievementEditorProc));
```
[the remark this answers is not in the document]
SyrianBallaS: Sure, I guess that's fine.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -650,19 +647,22 @@ void Dlg_MemBookmark::WriteFrozenValue(const MemBookmark & Bookmark)
...
-    sprintf_s(buffer, sizeof(buffer), "%0*x", width, Bookmark.Value());
-
-    for (unsigned int i = 0; i < strlen(buffer); i++)
+    
+    if (auto buffer{ std::make_unique<char[]>(32) }; sprintf_s(buffer.get(), 32, "%0*x", width, Bookmark.Value()) != -1)
     {
-        c = buffer[i];
-        n = (c >= 'a') ? (c - 'a' + 10) : (c - '0');
-        MemoryViewerControl::editData(addr, (i % 2 != 0), n);
+        std::string str{ buffer.release() };
```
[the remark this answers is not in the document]
SyrianBallaS: From what I've read, std;:string uses RAII and owns a pointer to string. I've also read std::string_view (immutable std::string) does not express ownership. I'll just change it to get and it should release it by itself.

SyrianBallaS:
Gonna close this, branch, adjust, merge when time allows.

src/RA_Dlg_MemBookmark.cpp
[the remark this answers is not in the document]
Jamiras: Internally, it may own a pointer to a string, but this is just calling the "std::string(const char*)" constructor. That copies from the pointer without taking ownership.

src/ra_errors.h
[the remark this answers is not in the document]
Jamiras: I know there's a lot of unused functions (and prototypes with bodies), but they're already in the code base. I'm just trying to not introduce more. I probably wouldn't have even noticed this one if you didn't attach a comment to it.

SyrianBallaS:
Heeded feedback. 
Found an opportunity to expand on ra_fwd.h in response to to the imagefactory header changed.
Have a few things in mind, but will focus on the enum to enum class conversion as it will greatly prevent bugs in some cases.

src/RA_httpthread.cpp
```diff
@@ -366,33 +367,34 @@ BOOL RAWeb::DoBlockingHttpGet(const std::string& sRequestedPage, std::string& Re
...
                     L"Content-Type: application/x-www-form-urlencoded",
-                    0,
+                    0_dw,
                     WINHTTP_NO_REQUEST_DATA, //WINHTTP_NO_REQUEST_DATA,
-                    0,
-                    0,
-                    0);
+                    0_dw,
+                    0_dw,
+                    0_dw);
```
Jamiras: I'm still not sold on this. No matter how many times I see "0_dw", it just looks ugly. I'm going to take a page from Google's C++ guidelines and ask that you remove these.

https://google.github.io/styleguide/cppguide.html#Operator_Overloading

> Overload operators judiciously. Do not create user-defined literals.

> User-defined literals allow the creation of new syntactic forms that are unfamiliar even to experienced C++ programmers.

SyrianBallaS: Sure, they're gone.

src/services/ImageRepository.cpp
```diff
@@ -353,7 +353,10 @@ HBITMAP ImageRepository::LoadLocalPNG(const std::string& sFilename, size_t nWidt
     // Scale the original IWICBitmapSource to the client rect size and convert the pixel format
     CComPtr<IWICBitmapSource> pToRenderBitmapSource;
     if (SUCCEEDED(hr))
-        hr = ConvertBitmapSource({ 0, 0, static_cast<LONG>(nWidth), static_cast<LONG>(nHeight) }, pOriginalBitmapSource, *&pToRenderBitmapSource);
+    {
+        RECT myRect{ 0, 0, to_signed(nWidth), to_signed(nHeight) };
+        hr = ConvertBitmapSource(myRect, pOriginalBitmapSource, *&pToRenderBitmapSource);
```
SyrianBallaS: I was looking over other files that may have been related to this before. I've seen a lot of usages as rvalues (temporaries) as lvalues, this was one such case. The other one's weren't too serious so I didn't change them. If you want to use as an rvalue, consider appending && to the typename.

SyrianBallaS:
branch has gotten pretty stale. gonna start over and leave the other stuff for later.

Issue #5, opened 2018-04-15
Support for AND clauses in leaderboard cancel conditions


meleu:
_From @Jamiras on March 21, 2018 14:55_

I was looking at the leaderboard code and noticed the following:
https://github.com/RetroAchievements/RASuite/blame/master/RA_Integration/RA_Leaderboard.cpp#L462
The cancel condition is passed TRUE when it's evaluated, which indicates `Test` should return true if any condition is true. This effectively makes the logic be `OR` for all conditions in the Cancel field (despite still using '_' as a separator). It looks like it's been that way since it was originally added, which is further supported by the fact that it's the only `timeTest` is passed TRUE.

There are currently 171 leaderboards across 22 games that have multiple conditions in their Cancel field, most notably Super Mario Bros and Super Mario Kart.

This behavior seems idiosyncratic (and I don't believe is properly implemented in RetroArch).

I don't know if there's any easy way to fix it. Ideally we'd want to make "_" behave like an and and introduce a new symbol for or (probably "S" like achievements).

But since it's dll behavior, we can't update the existing leaderboards until a new dll is published. Maybe have an interim release that defines both "\_" and "S" as `OR`, fix all the achievements, then redefine "\_" back to `AND` in a later release?

_Copied from original issue: RetroAchievements/RASuite#58_

meleu:
_From @kdecks on April 8, 2018 4:35_

I think this is a bug. I'll mark it as such.

Jamiras:
Affected leaderboards have been corrected, temporary code was removed before 0.73 was released. Closing.

PR #31, opened 2018-04-28
Standardize all condition processing

Jamiras:
This is preparatory work for supporting ANDs in leaderboard cancel conditions (#5).
* Renamed `ConditionSet` to `ConditionGroup` and added a new `ConditionSet` which is a collection of `ConditionGroup`s (foundation of an `Achievement`).
  * `ConditionSet` is a group of `ConditionGroup`s. `ConditionSet.Test` returns true if the first group evaluates true and at least one of the other groups evaluate true.
  * `ConditionGroup` is a collection of `Condition`s. `ConditionGroup.Test` returns true if all conditions in the group evaluate true.
  * `Condition` is a single condition. `Condition.Test` returns true if the condition evaluates true.
* Moved all parsing/serialization of Conditions into RA_Condition.cpp
  * Achievements, Leaderboards, and Conditional Rich Presence now fully support the entire Condition lexicon, although it seems impractical to use HitCounts/ResetIf/PauseIf in Leaderboards or Conditional Rich Presence.
* Added `ActiveBankRAMRead(addr, size)` to `MemManager` to eliminate duplicated code in Condition, Dlg_MemBookmark, Dlg_Memory, MemValue, and MemManager itself.

### DIFF
--- a/src/RA_Achievement.cpp
+++ b/src/RA_Achievement.cpp
@@ -1,30 +1,7 @@
 #include "RA_Achievement.h"
 
-#include <windows.h>
-#include <commctrl.h>
-#include <stdlib.h>
-#include <stdio.h>
-#include <memory.h>
-#include <assert.h>
-#include <io.h>
-#include <sstream>
-
-#include "RA_Dlg_Achievement.h"
-#include "RA_Dlg_GameTitle.h"
-
-#include "md5.h"
-#include "RA_md5factory.h"
-
-#include "RA_MemManager.h"
-#include "RA_User.h"
-#include "RA_PopupWindows.h"
-#include "RA_httpthread.h"
-#include "RA_Defs.h"
-#include "RA_ImageFactory.h"
 #include "RA_Core.h"
-#include "RA_Leaderboard.h"
-#include "RA_Condition.h"
-#include "RA_RichPresence.h"
+#include "RA_ImageFactory.h"
 
 //	No game-specific code here please!
 
@@ -34,7 +11,6 @@ Achievement::Achievement( AchievementSetType nType ) :
 	m_nSetType( nType ), m_bPauseOnTrigger( FALSE ), m_bPauseOnReset( FALSE )
 {
 	Clear();
-	m_vConditions.push_back( ConditionSet() );
 }
 
 void Achievement::Parse( const Value& element )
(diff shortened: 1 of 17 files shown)


Jamiras:
Once this code is released, we need to audit all existing leaderboard cancel conditions and change them from "A=1_B=2" to "1=1SA=1SB=2", then we can remove the temporary code that does the conversion internally. That will open up AND support for leaderboard cancel conditions to complete #5. On March 21, there were 171 that met this criteria.

RetroArch currently treats the cancel condition the same as the start and submit conditions, meaning that "A=1_B=2" means "A=1" AND "B=2", which is the desired result, but not currently correct. I do not believe any changes will be required in RetroArch.

https://github.com/libretro/RetroArch/blob/master/cheevos/cheevos.c#L815
https://github.com/libretro/RetroArch/blob/master/cheevos/cheevos.c#L1857

src/RA_Leaderboard.cpp
```diff
@@ -258,48 +235,30 @@ void RA_Leaderboard::ParseLBData( char* pChar )
...
+				for (size_t i = 0; i < m_cancelCond.GetGroup(0).Count(); ++i)
+				{
+					m_cancelCond.AddGroup();
+					m_cancelCond.GetGroup(i + 1).Add(m_cancelCond.GetGroup(0).GetAt(i));
+				}
+
+				m_cancelCond.GetGroup(0).Clear();
 			}
-			while( *pChar == '_' );
+			// end backwards compatibility conversion
```
Jamiras: This block (lines 245-256) should be removed once existing leaderboard cancel conditions are converted.

src/RA_Achievement.cpp
```diff
@@ -54,149 +30,112 @@ void Achievement::Parse( const Value& element )
...
 	{
-		std::string sMem = element["MemAddr"].GetString();
-		char buffer[8192];
-		sprintf_s( buffer, 8192, sMem.c_str() );
-		ParseMemString( buffer );
+		const char* sMem = element["MemAddr"].GetString();
+		if (!m_vConditions.ParseFromString(sMem) || *sMem != '\0') {
+			ASSERT(!"Invalid MemAddr");
+			m_vConditions.Clear();
+		}
```
Jamiras: Similar to RetroArch, this will disable an achievement if there's an issue parsing it. `ConditionSet.Test` will return false if there aren't any groups.

src/RA_Achievement.cpp
```diff
@@ -54,149 +30,112 @@ void Achievement::Parse( const Value& element )
...
+
 	if( bDirtyConditions )
 	{
 		SetDirtyFlag( Dirty_Conditions );
 	}
 
-	if( bResetConditions )
+	if( bResetConditions && bNotifyOnReset )
 	{
 		Reset();
```
Jamiras: This `Reset` is unnecessary, `Test` will have already done the Reset. I'll correct in a future commit.

The `bNotifyOnReset` boolean that was added only pops the dialog if Pause on Reset it checked and at least one condition with a non-zero HitCount target as reset.

src/RA_Achievement.cpp
```diff
@@ -210,16 +149,12 @@ BOOL Achievement::Test()
 		}
 	}
 
-	return bRetVal && bRetValSubCond;
+	return bRetVal;
```
Jamiras: subcondition evaluation was also done in `Test`.

src/RA_Achievement.h
```diff
@@ -122,7 +115,7 @@ class Achievement
 	/*const*/ AchievementSetType m_nSetType;
 
 	AchievementID m_nAchievementID;
-	std::vector<ConditionSet> m_vConditions;
+	ConditionSet m_vConditions;
```
Jamiras: `ConditionSet` was renamed `ConditionGroup` and a group of `ConditionGroup`s is now a `ConditionSet`. `ConditionSet` contains the logic for supporting OR'd groups.

src/RA_Condition.cpp
```diff
@@ -10,10 +10,7 @@ static_assert( SIZEOF_ARRAY( COMPARISONTYPE_STR ) == NumComparisonTypes, "Must m
 const char* CONDITIONTYPE_STR[] = { "", "Pause If", "Reset If", "Add Source", "Sub Source", "Add Hits"};
 static_assert( SIZEOF_ARRAY( CONDITIONTYPE_STR ) == Condition::NumConditionTypes, "Must match!" );
 
-int g_AddBuffer = 0;
-int g_AddHits = 0;
```
Jamiras: these are now passed from `ConditionGroup` to `Condition`, instead of being global.

src/RA_Condition.h
```diff
@@ -77,18 +75,12 @@ class CompVariable
 
 	void ResetDelta()
 	{
-		m_nPreviousVal = m_nVal;
+		m_nPreviousVal = 0;
```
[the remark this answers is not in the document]
Jamiras: Reset is supposed to set things to their initial state, and the initial state is 0.

`ResetDelta` is only called by `RA_Leaderboard::Reset`, which is only called by `RA_LeaderboardManager::Reset`, which is only called by `_RA_OnLoadState`. 

When loading a state, the delta value is unknown, but I guess assuming it was the current value is more likely than assuming it was 0. I'll change it back.

GameDragon2k:
> we need to audit all existing leaderboard cancel conditions and change them from "A=1_B=2" to "1=1SA=1SB=2"
Is the `1=1` for Core group? Is that required for the cancel condition to work properly?

Jamiras:
Yes. The previous "A=1_B=2" is "A=1" or "B=2". To represent that in a standard form, we need a dummy core group and two alt groups.

GameDragon2k:
But this is remedied here, yeah?
https://github.com/RetroAchievements/RAIntegration/blob/48b44243832033ee2f68765c2ab1fd71792206be/src/RA_Condition.cpp#L523-L529

So it could be short-handed to `SA=1SB=2`?

Jamiras:
Absolutely, but I don't think RetroArch supports that shorthand, so _existing_ items should use the 1=1 format.

PR #32, opened 2018-04-29
Precompiled header (Probably should be accepted last)

SyrianBallaS:
From my PRs, this should probably be accepted last because it moves the includes of almost every header file.

It moves all library files to "stdafx.h", but keeps RA_Defs.h includes under the RA_EXPORTS define since emulators won't be required to use it. Also added include guards everywhere. Cleaned up includes in source files. Another thing we could do is forward declare as much as possible but that'll be for another time.

This should probably be accepted last, putting up here so you guys can see it. Tried my absolute hardest not to do any whitespace changes, they just happen automatically from Visual Studio.

### DIFF
--- a/src/RA_Achievement.cpp
+++ b/src/RA_Achievement.cpp
@@ -1,29 +1,19 @@
 #include "RA_Achievement.h"
 
-#include <windows.h>
-#include <commctrl.h>
-#include <stdlib.h>
-#include <stdio.h>
-#include <memory.h>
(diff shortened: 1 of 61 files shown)


SyrianBallaS:
Ill revert and reopen this later, and commit one file at time to make sure whitespace doesn't happen. I think I pressed the wrong keyboard shortcut because I was using snippets for automation.

PR #34, opened 2018-05-03
making RALibretro cross-compilable

meleu:
When trying to cross-compile RALibretro on a Linux machine, the compiler complains about these header files on `src/RA_Interface.cpp`. It's just a matter of "case-sensitiveness".

Unfortunately I'm unable to compile RA_Integration right now (Visual Code doesn't run on my Windows virtual machine) and I am only 99% sure that the changes here are pretty acceptable and won't cause any problem.

Can some of you guys test it?

### DIFF
--- a/src/RA_Interface.cpp
+++ b/src/RA_Interface.cpp
@@ -59,8 +59,8 @@ void RA_GetEstimatedGameTitle( char* sNameOut )
 
 #ifndef RA_EXPORTS
 
-#include <WTypes.h>
-#include <WinHttp.h>
+#include <wtypes.h>
+#include <winhttp.h>
 #include <assert.h>
 #include <string>
 


Jamiras:
wtypes.h and winhttp.h are windows headers. i'm surprised there's any form of them on a Linux machine.

Even if you do manage to get RALibRetro to build on Linux, the toolkit DLL won't. It's heavily integrated with the Windows API.

meleu:
I'm talking about **cross** compilation (the art of creating an executable for a platform other than the one on which the compiler is running). I'm not using the default linux toolchain, but one for cross compiling for Windows.

After changing those strings to their lower case equivalents I was able to cross compile RALibretro on my Linux and perfectly run it on my Windows virtual machine. More details here: https://github.com/RetroAchievements/RALibretro/pull/23

My request is just to check if those header files in lowercase is OK to build on Visual Studio. I doubt it would causes problem, but it doesn't hurt to check it, right?

Unfortunately I'm unable to run Visual Studio on my virtual machine, that's why I'm asking for your help.

Thanks in advance.

Jamiras:
`winhttp.h` is lowercase, so that indicates that case sensitivity is not an issue on Windows. `WTypes.h` is cased correctly, but I have no issues changing it for your use case.

PR #36, opened 2018-05-04
CppCoreCheck Class Rules (Wait for existing PRs)

SyrianBallaS:
These basically just define and/or delete every default operation if one was defined. Default constructors where defaulted/deleted based on the context of how the classes were designed. It does not give defaults to every class.

Classes that had globals which flagged this had copying and moving deleted.

### DIFF
--- a/src/RA_AchievementOverlay.cpp
+++ b/src/RA_AchievementOverlay.cpp
@@ -98,21 +98,12 @@ void AchievementOverlay::SelectNextTopLevelPage( BOOL bPressedRight )
 	}
 }
 
-AchievementOverlay::AchievementOverlay()
-{
-	m_hOverlayBackground = nullptr;
-	m_nAchievementsSelectedItem = 0;
-	m_nFriendsSelectedItem = 0;
-	m_nMessagesSelectedItem = 0;
-	m_nNewsSelectedItem = 0;
-	m_nLeaderboardSelectedItem = 0;
-}
 
-AchievementOverlay::~AchievementOverlay()
+AchievementOverlay::~AchievementOverlay() noexcept
 {
 	if( m_hOverlayBackground != nullptr )
 	{
-		DeleteObject( m_hOverlayBackground );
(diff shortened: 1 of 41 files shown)


Jamiras:
Ok, I'm going to have to ask you to stop.

What is your goal with this project? So far, all you've done is consume large portions of my time. You haven't created a single achievement, so why do you care so much about changes to the toolkit?

You clearly aren't testing your changes: https://github.com/RetroAchievements/RAIntegration/pull/23#pullrequestreview-117482750
https://github.com/RetroAchievements/RAIntegration/pull/29/commits/4f093ac997fac972f54e3ce6b151736580f704b0

Sometimes you're not even building the code:
https://github.com/RetroAchievements/RAIntegration/pull/29#pullrequestreview-117449235

And you continue to make new PR's fully knowing that I plan to do a whitespace refactoring in the next couple days that's going to cause conflicts with everything. Not to mention the conflicts you're going to have with your own changes. As of this morning, you have 6 open PRs, with a combined total of nearly 12,000 lines changed!

When I make changes, I typically run with them for at least three or four days before I even create the PR. You're spitting out PRs faster than I can review them. And so far, none of them have added anything to the toolkit. I agree that some of the code needs to be cleaned up, but it needs to be done with care, not a fire hose. 

Additionally, you're introducing a lot of C++17 code. I've only recently started using C++11 with any sort of frequency, so much of this code seems foreign. I have a general feeling that the C++17 code is going to turn away developers that could otherwise be contributing.

At this point, I'm going to step back from your work and focus on mine. When I do start looking at your changes again, I will do so one PR at a time. I will not look at a second PR until the first is fully merged. I'm also going to spend some time contemplating the benefit of introductin C++17 features. The code functions as is, and we want it to remain accessible to as many potential future devs as possible.

SyrianBallaS:
Well I'll stop as you say. I could just skip straight to the ui part. I
care because of the crashing and hanging that keep happening.

PR #39, opened 2018-05-05
apply coding style, primarily tabs->spaces and spaces in/around parentheses

Jamiras:
recommend viewing with "Hide whitespace changes" checked:
image


GameDragon2k:
This looks good, really good job.

The only nitpick I have atm is no indentation on switch cases. It's not mentioned anywhere in the `CONTRIBUTING.md` file either. But again, this is personal preference.

Jamiras:
I tend to agree, even though the majority of the code was did not previously indent the switch cases. I've updated the code and the CONTRIBUTING.md file.

PR #41, opened 2018-05-07
support for ResetIf HitCount

Jamiras:


meleu:
@jamiras would it be possible to post here a compiled version so we can play with it?

Jamiras:
[RA_Integration.zip](https://github.com/RetroAchievements/RAIntegration/files/1983958/RA_Integration.zip)
Note that HitCount behavior hasn't changed, so your best experience will be using this with a Delta or the ChangesTo trick (ResetIf value = X (N), PauseIf delta = X)

PR #42, opened 2018-05-07
Update watched address when navigating filter results with keyboard

Jamiras:
When clicking on a filter result, the watched address updates, but then using up/down to select the next/previous item does not. This change supports both.


meleu:
You've read my mind! I was about to open an issue with this feature request!! :)

PR #45, opened 2018-05-13
add unit tests for RA_Condition.cpp and RA_Defs.cpp

Jamiras:
Uses VS unit testing framework, which is available without additional installation in VS2017 community. Not sure about previous version of VS.

Requires pulling in the CPP files to be tested as the DLL does not expose the functionality that is desired for testing.

To run: After building, open the Test Explorer window and Run All

image


GameDragon2k:
Everything seems to be working fine. Compiled with no issues and all unit tests succeeded.

PR #46, opened 2018-05-18
Support for HitCount on PauseIf

Jamiras:
* PauseIf(0) has no behavioral change. When not true, it will unpause itself.
* PauseIf(N) will only pause if HitCount is N. Since we don't want to automatically reset the HitCount when the condition is not true, PauseIf(N) never resets itself. It must be reset with a ResetIf. As a paused group is no longer processed, the ResetIf must be in a different group. 
    * This includes PauseIf(1). Semantically, this means "once the condition is true, disable this group until I tell you to re-enable everything", whereas PauseIf(0) means "only disable this group while the condition is true".

### DIFF
--- a/src/RA_Condition.cpp
+++ b/src/RA_Condition.cpp
@@ -358,44 +358,64 @@ bool Condition::Compare(unsigned int nAddBuffer)
 
 bool ConditionGroup::Test(bool& bDirtyConditions, bool& bResetAll)
 {
-    unsigned int nAddBuffer = 0;
-    unsigned int nAddHits = 0;
-    bool bConditionValid = false;
-    bool bSetValid = true; // important: empty group should evaluate true
-    unsigned int i = 0;
-
     const unsigned int nNumConditions = m_Conditions.size();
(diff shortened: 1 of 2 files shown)


GameDragon2k:
Resetting pause from another group is a workable and viable solution, but I feel it's a bit sloppy and cumbersome, no?

Forcing a reset in the same group seems more user friendly. Thoughts?

GameDragon2k:
Rather than an UnPauseIf flag, I would prefer a secondary ResetIf flag that has priority and is functional during a pause.

Jamiras:
Call it what you want, it's still a new flag and out of scope for this ticket. We can always add it later if there's demand for it. I'm not really sure there's demand for PauseIf HitCount. Everything I've seen proposed for PauseIf HitCount can just as easily be implemented with ResetIf HitCount.

GameDragon2k:
I don't know if I agree it's outside the scope of this, and but if you believe this is the route we should take, I won't argue. Was just giving my own opinion.

I'll merge this then.

PR #49, opened 2018-05-25
Demacrofied where possible

SyrianBallaS:
Demacrofied constants, functions, and function aliases.

Macros don't follow the same rules as variables so some other stuff had to be changed too. Macros that were left alone were the Resource IDs and Versioning info since the files that use them can't understand variables.


src/RA_Core.cpp
```diff
@@ -1182,6 +1190,9 @@ void RememberWindowSize(HWND hDlg, const char* sDlgKey)
     iter->second.nHeight = rc.bottom - rc.top;
 }
 
+_Use_decl_annotations_
+BOOL CALLBACK SetDirectory() { return SetCurrentDirectory(NativeStr(g_sHomeDir).c_str()); }
+
```
SyrianBallaS: Found this to be extremely repetitive so I made a function for it.

src/RA_Defs.h
```diff
@@ -52,45 +55,58 @@
...
+inline constexpr auto RA_OVERLAY_BG_FILENAME         = TEXT(".\\Overlay\\overlayBG.png");
+inline constexpr auto RA_NEWS_FILENAME               = TEXT(".\\RACache\\Data\\ra_news.txt");
+inline constexpr auto RA_TITLES_FILENAME             = TEXT(".\\RACache\\Data\\gametitles.txt");
+inline constexpr auto RA_LOG_FILENAME                = TEXT(".\\RACache\\Data\\RALog.txt");
 
-#define RA_HOST_URL						"retroachievements.org"
-#define RA_HOST_IMG_URL					"i.retroachievements.org"
+inline constexpr auto RA_HOST_URL                    = TEXT("retroachievements.org");
+inline constexpr auto RA_HOST_IMG_URL                = TEXT("i.retroachievements.org");
+
```
SyrianBallaS: First off these are aligned, before the change they didn't look aligned to begin with.

We could make them constant expressions but not like before. I put TEXT for good measure because the Windows API flips between multibyte and byte string.

src/RA_Defs.h
```diff
@@ -52,45 +55,58 @@
...
+
+template<typename Array, class = std::enable_if_t<std::is_array_v<Array>>>
+inline constexpr auto SIZEOF_ARRAY(const Array& ar) noexcept->decltype(sizeof(ar) / sizeof(ar[0]))
+{
+    return { sizeof(ar) / sizeof(ar[0]) };
+}
```
SyrianBallaS: Not sure if we even need this anymore after #40 but demacrofied anyway. It has to be passed by reference or it will decay into a pointer giving a result of 1 each time.

src/RA_Defs.h
```diff
@@ -275,15 +290,29 @@ extern std::string Narrow(const std::string& wstr);
...
+using NativeCharType   = NativeString::value_type;
+using NativeCString    = NativeString::const_pointer;
+#endif // UNICODE
+
+
+NativeString NativeStr(NonNativeCString str) noexcept;
+NativeString NativeStr(const NonNativeString& str) noexcept;
+NativeString NativeStr(NonNativeString&& str) noexcept;
+NativeString NativeStr(const NativeString& str) noexcept;
+NativeString NativeStr(NativeCString str) noexcept;
```
SyrianBallaS: Yes, all these overloads are needed since Widen/Narrow are overloaded.

src/RA_Dlg_Achievement.cpp
```diff
@@ -515,11 +515,13 @@ INT_PTR Dlg_Achievements::AchievementsProc(HWND hDlg, UINT nMsg, WPARAM wParam,
...
-                        TEXT("Refresh from Server"),
-                        MB_YESNO | MB_ICONWARNING) == IDYES)
+                    else if (auto msg = []() noexcept {
+                        tstring tstr{ TEXT("Are you sure that you want to download fresh achievements from ") };
+                        tstr.append(RA_HOST_URL);
+                        tstr.append(TEXT("?\n"));
+                        tstr.append(TEXT("This will overwrite any changes that you have made with fresh achievements from the server.\n"));
+                        return tstr;
+                    }; MessageBox(hDlg, msg().c_str(), TEXT("Refresh from Server"), MB_YESNO | MB_ICONWARNING) == IDYES)
                     {
```
SyrianBallaS: Some of you guys might not like this, but this was the most optimal way I could thing of.

src/RA_Dlg_Achievement.cpp
```diff
@@ -560,7 +562,7 @@ INT_PTR Dlg_Achievements::AchievementsProc(HWND hDlg, UINT nMsg, WPARAM wParam,
                     //	Add a new achievement with default params
                     Achievement& Cheevo = g_pActiveAchievements->AddAchievement();
                     Cheevo.SetAuthor(RAUsers::LocalUser().Username());
-                    Cheevo.SetBadgeImage("00000");
+                    Cheevo.SetBadgeImage(RA_UNKNOWN_BADGE_IMAGE_URI);
```
SyrianBallaS: We might as well use it or remove it.

src/RA_ImageFactory.cpp
```diff
@@ -14,9 +14,6 @@
 
 UserImageFactoryVars g_UserImageFactoryInst;
 
-#ifndef SAFE_RELEASE
-#define SAFE_RELEASE( p ) { if ( (p) != nullptr ) { (p)->Release(); (p) = nullptr; } }
-#endif
```
SyrianBallaS: Should've noticed it before but this isn't used at all anymore.

src/RA_Defs.h
```diff
@@ -246,19 +262,18 @@ const int SERVER_PING_DURATION = 2 * 60;
...
 
-#define RA_LOG RADebugLog
+
+inline constexpr auto& RA_LOG = RADebugLog;
 
 #ifdef _DEBUG
 #undef ASSERT
-#define ASSERT( x ) assert( x )
+inline auto ASSERT(bool x) noexcept->decltype(assert(x)) { assert(x); }
 #else
```
SyrianBallaS: It's labeled `noexcept `because assertions do not throw exceptions.

src/RA_AchievementSet.cpp
```diff
@@ -67,8 +67,7 @@ void AchievementSet::OnRequestUnlocks(const Document& doc)
 {
     if (!doc.HasMember("Success") || doc["Success"].GetBool() == false)
     {
-        ASSERT(!"Invalid unlocks packet!");
-        return;
+        throw std::invalid_argument{"Invalid unlocks packet!"};
```
Jamiras: I think the choice to use asserts was intentional. An uncaught throw will terminate the app. I think better error handling/reporting _is_ desirable, but these throws are not.

RA_Integration.sln
```diff
@@ -15,14 +15,19 @@ EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Win32 = Debug|Win32
+		Debug|x64 = Debug|x64
 		Release|Win32 = Release|Win32
+		Release|x64 = Release|x64
 		ReleaseWithDebug|Win32 = ReleaseWithDebug|Win32
+		ReleaseWithDebug|x64 = ReleaseWithDebug|x64
```
Jamiras: i don't think we're ready for x64 yet, please remove these.

src/RA_AchievementOverlay.cpp
```diff
@@ -1281,10 +1281,10 @@ void AchievementOverlay::Render(HDC hRealDC, RECT* rcDest) const
 
         //	Draw control text:
         sprintf_s(buffer, 1024, " ->:%s ", "Next");
-        TextOut(hDC, nRightPx - nControlsX1, nControlsY1, NativeStr(buffer).c_str(), strlen(buffer));
+        TextOut(hDC, nRightPx - nControlsX1, nControlsY1, NativeStr(std::string{ buffer }).c_str(), strlen(buffer));
 
         sprintf_s(buffer, 1024, " <-:%s ", "Prev");
-        TextOut(hDC, nRightPx - nControlsX1, nControlsY2, NativeStr(buffer).c_str(), strlen(buffer));
+        TextOut(hDC, nRightPx - nControlsX1, nControlsY2, NativeStr(std::string{ buffer }).c_str(), strlen(buffer));
```
Jamiras: Why wrap buffer in a `std::string`? `NativeStr` accepts a `NativeCString`, which if I understand it is just a `const char*` (i.e. buffer). Converting to a `std::string` seems like overhead that can be avoided.

src/RA_Core.cpp
```diff
@@ -401,8 +401,10 @@ API int CCONV _RA_OnLoadNewRom(const BYTE* pROM, unsigned int nROMSize)
         {
             //	Some other fatal error... panic?
             ASSERT(!"Unknown error from requestgameid.php");
-
-            MessageBox(g_RAMainWnd, NativeStr("Error from " RA_HOST_URL "!\n").c_str(), TEXT("Error returned!"), MB_OK);
+            tstring tstr{TEXT("Error from ")};
+            tstr.append(RA_HOST_URL);
+            tstr.append(TEXT("!\n"));
```
Jamiras: Previously, a single string was created at compile time. Unless the compiler optimizes this back into a single string, you're looking at possibly introducing one or more reallocs.

src/RA_Core.cpp
```diff
@@ -500,10 +502,11 @@ API BOOL CCONV _RA_OfferNewRAUpdate(const char* sNewVer)
 {
     char buffer[1024];
     sprintf_s(buffer, 1024, "Update available!\n"
-        "A new version of %s is available for download at " RA_HOST_URL ".\n\n"
+        "A new version of %s is available for download at %s.\n\n"
         "Would you like to update?\n\n"
         "Current version:%s\n"
         "New version:%s\n",
+        Narrow(RA_HOST_URL).data(),
```
Jamiras: Prefer `c_str()` for consistency.

src/RA_Core.cpp
```diff
@@ -852,20 +855,25 @@ API void CCONV _RA_CheckForUpdate()
         }
         else
         {
+            // basic_ostringstream would seem overkill here
+            tstring tstr{ TEXT("Error in download from ") };
+            tstr.append(RA_HOST_URL);
+            tstr.append(TEXT("...\n"));
+            tstr.append(TEXT("Please check your connection settings or RA forums!\n"));
```
Jamiras: Similar comment about string building. Will this get resolved at compile time?

src/RA_Core.cpp
Jamiras: but you never call the function. 

also, recommend a more descriptive name. maybe `ChangeToHomeDirectory()`?

src/RA_Core.cpp
Jamiras: I think I remember something about the compiler optimizing the `+ operator`, but I don't remember anything about it optimizing the `append` method.

src/RA_Defs.cpp
```diff
@@ -62,6 +62,52 @@ std::string Narrow(const std::string& str)
...
 
+
+NativeString NativeStr(NonNativeCString str) noexcept
+{
+#ifdef UNICODE
+	return Widen(str);
+#else
+	return Narrow(str);
+#endif // UNICODE
+}
```
Jamiras: Do you really need non-const versions of these functions?

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -21,7 +21,13 @@
 #include "RA_httpthread.h"
 #include "RA_md5factory.h"
 
-#define KEYDOWN(vkCode) ((GetAsyncKeyState(vkCode) & 0x8000) ? true : false)
+
+inline constexpr auto CALLBACK KEYDOWN(int vk)->decltype((GetAsyncKeyState(vk) & 0x8000) ? true : false)
+{
+    return { (GetAsyncKeyState(vk) & 0x8000) ? true
+                                             : false };
+}
```
Jamiras: I admit that I don't understand the `->decltype()` syntax, but it seems there should be a way to do this without duplicating the body of the function in the `decltype`. Can't you just use `decltype(bool)` or change the `auto` to an `bool`?

RA_Integration.sln
SyrianBallaS: Sure thing

src/RA_AchievementSet.cpp
SyrianBallaS: Alright, I'll put it back until we can figure out a way to handle it properly.

src/RA_Core.cpp
SyrianBallaS: std::string is not a literal type and cannot result in a constant expression. I can see what you mean but the speed is the same. A literal type either has all it's members as literal types or it's default constructor is constexpr.

src/RA_Core.cpp
SyrianBallaS: I'm sorry, when I saw the warning in VS it said it expected char*. As long as std::string isn't const data() returns char* and c_str() returns const char* (C++17) but it doesn't matter in this case so it can be const char*.

src/RA_Core.cpp
SyrianBallaS: std::string cannot be built in compile time, it's just not possible. const char*, const wchar_t can be but it has to be in the same expression (constexpr).

src/RA_AchievementOverlay.cpp
SyrianBallaS: OK I must have misunderstood the concept, for some reason I thought the argument inside NativeStr was supposed be std::string but then realized I made overloads for c strings.

src/RA_Core.cpp
SyrianBallaS: Yeah I think that's a better name, since it always sets it to the home directory.

src/RA_Defs.cpp
SyrianBallaS: `NonNativeCString `and `NativeCString `are always `const`, if you mean `const `in the beginning there's no point.

src/RA_Core.cpp
Jamiras: Admittedly, _this_ code probably isn't called frequently enough to worry about execution speed, but it's something I'm always thinking about.

In this case, `NativeStr` is just used to Narrow/Widen the string, it probably should just be a `TEXT`:
```
MessageBox(g_RAMainWnd, TEXT("Error from " RA_HOST_URL "!\n"), TEXT("Error returned!"), MB_OK);
```
The following code results in a single contiguous block of memory at compile tile:
```
const char* text = "Hello" ", " "world";
```
I'm probably remembering another language, since `operator +` isn't available for `const char*`, but I thought this did too:
```
std::string text = "Hello" + ", " + "world";
```

src/RA_Defs.cpp
Jamiras: Nevermind. One takes a `NonNativeString` and the other is a `NonNativeCString`. (missed the C).

src/RA_Defs.h
Jamiras: alignment issues are due to tabs v. spaces. I've noticed that the reformatter doesn't change them in the middle of lines. In this particular block of code, there's a tab immediately after `RA_GAME_HASH_FILENAME` and `RA_GAME_LIST_FILENAME` (try highlighting the single space after each).

src/RA_Dlg_GameLibrary.cpp
Jamiras: > Do you see a disambiguation is important?
Yes, but in this case, why use `auto` and `decltype` when you can just explicitly say `bool`? My concern here is the duplication of code. This snippet can never return anything other than a bool, so why have the compiler deduce that?

SyrianBallaS: Sure, it's not a big deal.

src/RA_Core.cpp
SyrianBallaS: That's a linear logarithmic operation. I would need to double check but I don't think variables can be used like that (macros can be though).

src/RA_Defs.h
SyrianBallaS: I think it would be easier for me to just convert tabs to spaces in Notepad++, it would show up as lots of lines. That might be better another PR, but for now I'll just look at what I'm changing.

src/RA_AchievementPopup.cpp
```diff
@@ -50,7 +54,7 @@ void AchievementPopup::Update(ControllerInput input, float fDelta, bool bFullScr
 {
     if (bPaused)
         fDelta = 0.0f;
-    fDelta = RAClamp<float>(fDelta, 0.0f, 0.3f);	//	Limit this!
+    fDelta = std::clamp(fDelta, 0.0F, 0.3F);	//	Limit this!
```
SyrianBallaS: Does the same exact thing but can be evaluated at compile time.

src/RA_AchievementPopup.cpp
```diff
@@ -3,6 +3,10 @@
 #include "RA_AchievementOverlay.h"
 #include "RA_ImageFactory.h"
 
+#ifdef WIN32_LEAN_AND_MEAN
+#include <MMSystem.h>
+#endif // WIN32_LEAN_AND_MEAN
```
SyrianBallaS: Decided to move this include and ShellAPI.h to where they are needed because only a few files did. MMSystem.h was only needed in this file (PlaySound).

src/RA_AchievementSet.cpp
```diff
@@ -381,7 +379,7 @@ BOOL AchievementSet::FetchFromWebBlocking(GameID nGameID)
         doc.HasMember("PatchData"))
     {
         const Value& PatchData = doc["PatchData"];
-        SetCurrentDirectory(NativeStr(g_sHomeDir).c_str());
+        ChangeToHomeDirectory();
```
SyrianBallaS: It's called everywhere, for now it's discardable but we'll put some checks in later.

src/RA_Core.cpp
```diff
@@ -33,48 +29,52 @@
...
-    bool bLoaded;
+    int nLeft{ 0 };
+    int nTop{ 0 };
+    int nWidth{ 0 };
+    int nHeight{ 0 };
+    bool bLoaded{ false };
 
-    static const int nUnset = -99999;
-} WindowPosition;
+    static const int nUnset{ -99999 };
```
SyrianBallaS: Tried making this constexpr but then the entire class would need to be changed.

src/RA_Core.cpp
SyrianBallaS: The way this is done now is guaranteed to built in compile-time and checks that RA_HOST_URL exists in that string_view (readonly string) at compile time (only string literals can be constant expressions, basic_string cannot).

src/RA_Core.cpp
```diff
@@ -852,28 +862,45 @@ API void CCONV _RA_CheckForUpdate()
...
+
+            // NB: It cannot be static or built with std::string because either a
+            //     data race (deadlock) or pointer dangling (could be leak,
+            //     access violoation, etc.) will occur
+            constexpr tstring_view view{
+                TEXT("Error in download from retroachievements.org...\nPlease check your connection"
+                " settings or RA forums!\n")
+            };
+            static_assert(view.find(RA_HOST_URL)); // just incase it changed
+
```
SyrianBallaS: The `constexpr ` and `static_assert` is all you need as proof that it's compile time, it's very small.

src/RA_Defs.cpp
```diff
@@ -10,68 +10,150 @@ GetParseErrorFunc GetJSONParseErrorStr = GetParseError_En;
 #endif
 
 static_assert(sizeof(BYTE*) == sizeof(char*), "dangerous cast ahead");
+
+_Use_decl_annotations_
```
SyrianBallaS: Don't know if you guys know what this is but it's for SAL (structured annotation language). It's if the declaration and definition are in separate locations.

src/RA_Defs.cpp
```diff
@@ -10,68 +10,150 @@ GetParseErrorFunc GetJSONParseErrorStr = GetParseError_En;
...
 {
     return reinterpret_cast<char*>(stream.data());
 }
 
+_Use_decl_annotations_
 std::string Narrow(const wchar_t* wstr)
 {
-    static std::wstring_convert< std::codecvt_utf8_utf16< wchar_t >, wchar_t > converter;
+    std::wstring_convert< std::codecvt_utf8_utf16< wchar_t >, wchar_t > converter;
     return converter.to_bytes(wstr);
```
SyrianBallaS: Having these static made no sense, RichPresense stuff still worked. The only thing I can think of that should be static is a mutex or semaphore, and that's about it or anything that deals with synchronization

src/RA_Defs.cpp
```diff
@@ -10,68 +10,150 @@ GetParseErrorFunc GetJSONParseErrorStr = GetParseError_En;
...
     //SetCurrentDirectory( g_sHomeDir.c_str() );//?
-    FILE* pf = nullptr;
-    if (fopen_s(&pf, RA_LOG_FILENAME, "a") == 0)
-    {
-        fwrite(data, sizeof(char), strlen(data), pf);
-        fclose(pf);
-    }
+    // SAL didn't like this
+    std::unique_ptr<FILE, decltype(&std::fclose)> fp{ std::fopen(RA_LOG_FILENAME, "a"), std::fclose };
+    fwrite(data, sizeof(char), strlen(data), fp.get());
```
SyrianBallaS: `FILE* ` is never allowed to be `nullptr` according to the specification in stdio.h. I have a class wrapper for this already but probably will hold off on it until you guys accept #57

src/RA_Defs.cpp
```diff
@@ -91,19 +173,77 @@ void RADebugLog(const char* format, ...)
...
+_Use_decl_annotations_
 BOOL DirectoryExists(const char* sPath)
 {
     DWORD dwAttrib = GetFileAttributes(NativeStr(sPath).c_str());
     return(dwAttrib != INVALID_FILE_ATTRIBUTES && (dwAttrib & FILE_ATTRIBUTE_DIRECTORY));
 }
+
+_Use_decl_annotations_
+ResizeContent::ResizeContent(HWND parentHwnd, HWND contentHwnd, AlignType newAlignType, bool isResize) noexcept
+{
```
SyrianBallaS: This constructor can't be constexpr and isn't from a class template so it shouldn't be in the header file.

src/RA_Defs.cpp
```diff
@@ -91,19 +173,77 @@ void RADebugLog(const char* format, ...)
...
+    hwnd.reset(contentHwnd);
+    nAlignType = newAlignType;
+    bResize = isResize;
+
+
+    auto lpRect{ std::make_unique<RECT>() };
+    GetWindowRect(hwnd.get(), lpRect.get());
```
SyrianBallaS: I was experimenting between `std::make_unique<RECT>()` and `std::make_unique<RARect>()`. As far as I can tell, they have the same effect. I made compile-time be removing inheritance and using a conversion operator.

src/RA_Defs.h
```diff
@@ -50,25 +50,19 @@
...
 #include <tchar.h>
 
-#ifdef WIN32_LEAN_AND_MEAN
-#include <MMSystem.h>
-#include <ShellAPI.h>
-#include <CommDlg.h>
-#endif // WIN32_LEAN_AND_MEAN
-
 #include <sstream>
+#include <string_view>
```
SyrianBallaS: compile time string building you can't modify it though. This one requires C++17. I noticed a lot of desire for compile-time stuff so C++17 should actually be a requirement because of its vast compile-time support via constant expressions.

src/RA_Defs.h
```diff
@@ -50,25 +50,19 @@
...
-#include <ShellAPI.h>
-#include <CommDlg.h>
-#endif // WIN32_LEAN_AND_MEAN
-
 #include <sstream>
+#include <string_view> 
 #include <queue>
 #include <map>
-#include <array>
+#include <ciso646>
```
SyrianBallaS: Reading &&, ||, etc. is sometimes difficult.

src/RA_Defs.h
```diff
@@ -89,60 +83,100 @@
...
 #pragma warning(pop)
 #endif	//RA_EXPORTS
 
-// Maybe an extra check just in-case
 
+// NB: Do NOT use _NORETURN on setters, varargs (C), C Asserts, or any void
+//     function that "returns", [[noreturn]] expects that you throw an exception
+//     on failure so if you put [[noreturn]] and "return" will throw a runtime
+//     error saying the function call was not saved properly.
 #define _NORETURN            [[noreturn]]
```
SyrianBallaS: I feel like it was important enough to leave a comment in-code.

src/RA_Defs.h
```diff
@@ -151,49 +185,124 @@ typedef DWORD			ARGB;
...
-        bottom = nY + nH;
     }
 
+    RARect(const RARect&) = delete;
+    RARect& operator=(const RARect&) = delete;
+    inline constexpr RARect(RARect&&) noexcept = default;
+    _NODISCARD inline constexpr RARect& operator=(RARect&&) noexcept = default;
+
+    inline constexpr RARect() noexcept = default; 
+
```
SyrianBallaS: Since it's default constructor is `constexpr`, this class is a literal type. Meaning it can be used in `if constexpr` or `static_assert`.

src/RA_Defs.h
```diff
@@ -151,49 +185,124 @@ typedef DWORD			ARGB;
...
+    _NODISCARD explicit constexpr operator LPRECT() noexcept { return &m_Rect; }
+
 public:
-    inline int Width() const { return(right - left); }
-    inline int Height() const { return(bottom - top); }
+    _NODISCARD inline constexpr LONG Width() const noexcept { return{ m_Rect.right - m_Rect.left }; }
+    _NODISCARD inline constexpr LONG Height() const noexcept { return{ m_Rect.bottom - m_Rect.top }; }
+
+    _NODISCARD inline constexpr LONG Area() const noexcept { return{ Width()*Height() }; }
+    _NODISCARD inline constexpr bool empty() const noexcept { return{ Area() == 0L }; }
```
SyrianBallaS: It's lower case to be compatible with `std::empty` (which is a constant expression, most container's empty method is not a constant expression). Also for a lot of other classes we can make them compatible with `ranged for` be supplying "`begin`" and "`end`" functions like `CodeNotes`, `ConditionSet` or `MemBookmarks`, or any Container based classed.

src/RA_Defs.h
```diff
@@ -151,49 +185,124 @@ typedef DWORD			ARGB;
...
+                nY + nH } // bottom
     {
-        left = nX;
-        right = nX + nW;
-        top = nY;
-        bottom = nY + nH;
     }
 
+    RARect(const RARect&) = delete;
+    RARect& operator=(const RARect&) = delete;
```
SyrianBallaS: Didn't see any need for expensive copies, cheap moves are better anyway.

src/RA_Defs.h
```diff
@@ -151,49 +185,124 @@ typedef DWORD			ARGB;
...
 };
 
-const RASize RA_BADGE_PX(64, 64);
-const RASize RA_USERPIC_PX(64, 64);
+inline constexpr RASize RA_BADGE_PX{ 64, 64 };
+inline constexpr RASize RA_USERPIC_PX{ 64, 64 };
+
+// Let's see it in action
+static_assert(RA_BADGE_PX == RA_USERPIC_PX);
+
```
SyrianBallaS: To show RASize can be used in constant expressions (compile-time ops).

src/RA_Defs.h
```diff
@@ -207,69 +316,28 @@ class ResizeContent
...
-
-        if (!bResize)
-            SetWindowPos(hwnd, nullptr, xPos, yPos, 0, 0, SWP_NOSIZE | SWP_NOZORDER);
-        else
-            SetWindowPos(hwnd, nullptr, 0, 0, xPos, yPos, SWP_NOMOVE | SWP_NOZORDER);
-    }
+private:
+    // So we don't have to define a destructor, plus if it doesn't use default_delete we have initialize it explicitly
+    std::unique_ptr<HWND__, decltype(&::DestroyWindow)> hwnd{ nullptr, &::DestroyWindow };
+    POINT pLT = POINT{};
```
SyrianBallaS: Jumped the gun a little, there's a simpler version in one my branches (not in a PR) but would rather have destruction be handled more efficiently automatically.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -873,10 +874,10 @@ void Dlg_MemBookmark::OnLoad_NewRom()
...
-    GetWindowRect(hDlg, &windowRect);
-    pDlgMemBookmarkMin.x = windowRect.Width();
-    pDlgMemBookmarkMin.y = windowRect.Height();
+    auto windowRect{ std::make_unique<RARect>() };
+    GetWindowRect(hDlg, LPRECT{ *windowRect });
+    pDlgMemBookmarkMin.x = windowRect->Width();
+    pDlgMemBookmarkMin.y = windowRect->Height();
 
     vDlgMemBookmarkResize.push_back(ResizeContent(hDlg,
         GetDlgItem(hDlg, IDC_RA_LBX_ADDRESSES), ResizeContent::ALIGN_BOTTOM_RIGHT, TRUE));
```
SyrianBallaS: I tested by using a managed pointer and it worked.
`RARect` and `RECT` were 16 bytes.
`RARect*` and `LPRECT` were 4 bytes.

src/RA_Integration.vcxproj
```diff
@@ -94,7 +94,7 @@
     <ProjectName>RA_Integration</ProjectName>
     <ProjectGuid>{825CB292-F069-4B27-82CF-3417746ACDF3}</ProjectGuid>
     <RootNamespace>RA_Integration</RootNamespace>
-    <WindowsTargetPlatformVersion>7.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
```
SyrianBallaS: I'm gonna need to check, but I'm pretty sure it says 7.0 in VS. Yes it does, source tree says there are no changes either.

src/RA_Dlg_Achievement.cpp
```diff
@@ -513,13 +516,11 @@ INT_PTR Dlg_Achievements::AchievementsProc(HWND hDlg, UINT nMsg, WPARAM wParam,
...
-                        tstr.append(RA_HOST_URL);
-                        tstr.append(TEXT("?\n"));
-                        tstr.append(TEXT("This will overwrite any changes that you have made with fresh achievements from the server.\n"));
-                        return tstr;
-                    }; MessageBox(hDlg, msg().c_str(), TEXT("Refresh from Server"), MB_YESNO | MB_ICONWARNING) == IDYES)
+                    else if (auto msg = []() constexpr noexcept {
+                        constexpr tstring_view view{ TEXT("Are you sure that you want to download fresh achievements from retroachievements.org?\nThis will overwrite any changes that you have made with fresh achievements from the server.\n") };
+                        static_assert(view.find(RA_HOST_URL));
+                        return view;
+                    }; MessageBox(hDlg, msg().data(), TEXT("Refresh from Server"), MB_YESNO | MB_ICONWARNING) == IDYES)
```
SyrianBallaS: string_view doesn't have `c_str()` member function

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -11,7 +11,7 @@
 #include "RA_md5factory.h"
 
 
-inline constexpr auto CALLBACK KEYDOWN(int vk)->decltype((GetAsyncKeyState(vk) & 0x8000) ? true : false)
+inline constexpr bool CALLBACK KEYDOWN(int vk) noexcept
 {
```
SyrianBallaS: It can't be auto deduced since it's not simple enough so we either need to put the type or a deduction. Didn't really matter to me so we put the type.

src/RA_Defs.h
```diff
@@ -89,60 +83,100 @@
...
+_CONSTANT_VAR RA_HOST_IMG_URL                = TEXT("i.retroachievements.org");
+
+// https://msdn.microsoft.com/en-us/library/windows/desktop/aa365530(v=vs.85).aspx
+#if _MBCS
+_CONSTANT_VAR RA_MAX_PATH = _MAX_PATH;
+#elif _UNICODE
+_CONSTANT_VAR RA_MAX_PATH = 32'767;
+#else
+#error Illegal character-set detected! Please specify a character-set!
+#endif // _MBCS
```
SyrianBallaS: Easy way to grab what the max path is supposed to be for buffered functions.

src/RA_Defs.h
```diff
@@ -207,69 +316,28 @@ class ResizeContent
...
-    }
+private:
+    // So we don't have to define a destructor, plus if it doesn't use default_delete we have initialize it explicitly
+    std::unique_ptr<HWND__, decltype(&::DestroyWindow)> hwnd{ nullptr, &::DestroyWindow };
+    POINT pLT = POINT{};
+    POINT pRB = POINT{};
+    AlignType nAlignType = AlignType{};
+    int nDistanceX{ 0 };
+    int nDistanceY{ 0 };
+    bool bResize{ false };
```
SyrianBallaS: these are the values that `inline constexpr ResizeContent() noexcept = default;` uses.

SyrianBallaS: We could've just use "<code>{}</code>" but wasn't sure that would be clear enough.

SyrianBallaS:
There are serious problems with Dlg_MemBookmark in release mode. This will be closed until it's finished in another branch.

PR #56, opened 2018-06-06
Owning Pointers non-custom

SyrianBallaS:
Basically this used RAII on pointers that did not need custom deleters. Everywhere I saw `new `and/or `delete ` (some functions didn't even have `delete `paired with `new`) it used `std::unique_ptr `because of it's small size (same size as the pointer, granted it uses default_delete) and speed (no reference counting).

The Memory Bookmark dialog could not be managed automatically and seemed like a mess (lots of undefined behavior). I was not able to observe side-effects because the pointers could not be read and always seemed corrupted. I basically changed it so we **_don't have to use pointers with it at all_** and maintain the same speed. Iterators are guaranteed to only be used if valid, if they are invalidated they won't be used. It's as simple as that.


src/RA_Dlg_MemBookmark.h
```diff
@@ -8,93 +8,169 @@
...
-    bool Decimal() const { return m_bDecimal; }
-
+    // if you're worried about speed just delete copying, the pointer is always
+    // corrupted, I don't know how you guys got it to work before. It still needs
+    // to be movable
+    MemBookmark() noexcept = default;
+    ~MemBookmark() noexcept = default;
+    MemBookmark(const MemBookmark&) = delete;
+    MemBookmark& operator=(const MemBookmark&) = delete;
+    MemBookmark(MemBookmark&&) noexcept = default;
```
SyrianBallaS: Copying is deleted so MemBookmark would be cheap to use without making it a pointer. For something to be a NullablePointer it has to be CopyContructible and CopyAssignable, this one is neither so that means it doesn't need to be a pointer.

src/RA_Dlg_MemBookmark.h
```diff
@@ -8,93 +8,169 @@
...
+
+    void UpdateBookmarks(_In_ bool bForceWrite);
+    auto AddBookmark(MemBookmark&& newBookmark) noexcept // ok it's void
+        ->decltype(std::declval<MyBookmarks&>().push_back(std::move(newBookmark)))
+    {
+        MemBookmark myClone;
+        newBookmark.Clone(myClone);
+        m_vBookmarks.push_back(std::move(newBookmark));
+        AddBookmarkMap(std::move(myClone));
+
```
SyrianBallaS: The clone absolutely refused to enter AddBookmarkMap after AddBookmark even though it was valid. It does make sense though. If "Bookmarks" are a part of "BookmarkMap" then the Bookmarks need to be added after it's updated.

src/RA_Dlg_MemBookmark.h
```diff
@@ -8,93 +8,169 @@
...
 
-    void AddBookmarkMap(MemBookmark* bookmark)
-    {
-        if (m_BookmarkMap.find(bookmark->Address()) == m_BookmarkMap.end())
-            m_BookmarkMap.insert(std::map<ByteAddress, std::vector<const MemBookmark*>>::value_type(bookmark->Address(), std::vector<const MemBookmark*>()));
-
-        std::vector<const MemBookmark*> *v = &m_BookmarkMap[bookmark->Address()];
-        v->push_back(bookmark);
-    }
+    void AddBookmarkMap(_In_ MemBookmark&& bookmark) noexcept;
```
SyrianBallaS: Put some annotations and ran SAL. All of these failed [[noreturn]], just using return; makes the app terminate. Might also be a bug because sometimes doesn't throw as well.

src/RA_Dlg_MemBookmark.h
```diff
@@ -8,93 +8,169 @@
...
 
 public:
-    const MemBookmark* FindBookmark(const ByteAddress& nAddr) const
+    _NODISCARD auto FindBookmark(_In_ const ByteAddress nAddr) const noexcept
+        ->decltype(std::declval<const MyBookmarkMap&>().find(nAddr))
     {
-        std::map<ByteAddress, std::vector<const MemBookmark*>>::const_iterator iter = m_BookmarkMap.find(nAddr);
-        return(iter != m_BookmarkMap.end() && iter->second.size() > 0) ? iter->second.back() : nullptr;
+        return m_BookmarkMap.find(nAddr);
```
SyrianBallaS: Nice and simple. The deduction guide is to eliminate ambiguity.

src/RA_Dlg_Memory.cpp
```diff
@@ -341,9 +341,9 @@ bool MemoryViewerControl::OnEditInput(UINT c)
 
         if (g_MemBookmarkDialog.GetHWND() != nullptr)
         {
-            const MemBookmark* Bookmark = g_MemBookmarkDialog.FindBookmark(nByteAddress);
-            if (Bookmark != nullptr)
-                g_MemBookmarkDialog.WriteFrozenValue(*Bookmark);
+            auto Bookmark = g_MemBookmarkDialog.FindBookmark(nByteAddress);
+            if (Bookmark != std::end(g_MemBookmarkDialog))
+                g_MemBookmarkDialog.WriteFrozenValue(ra::at(nByteAddress, Bookmark->second));
         }
```
SyrianBallaS: This was honestly all you had to do to use iterators and not get an exception thrown.

src/RA_Defs.h
```diff
@@ -286,4 +296,18 @@ typedef std::basic_string<TCHAR> tstring;
...
+// temp, something similar is in the GSL but need to fix up the Memory Bookmarker - Samer
+// Didn't include my type_traits here for fear people would be scared of lots of lines
+template<typename Container>
+_NODISCARD auto at(typename Container::size_type pos, Container& myContainer) noexcept
+->decltype(std::declval<Container&>().at(pos))
+{
+    return myContainer.at(pos);
+}
+
+} // namespace ra
```
SyrianBallaS: This was used for the MemBookmark dialog since there wasn't a getter for the bookmark map.

src/RA_Dlg_Memory.cpp
```diff
@@ -622,14 +622,21 @@ void MemoryViewerControl::RenderMemViewer(HWND hTarget)
                 for (int j = 0; j < 16; ++j)
                 {
                     notes |= (g_MemoryDialog.Notes().FindCodeNote(addr + j) != nullptr) ? (1 << j) : 0;
-                    const MemBookmark* bm = g_MemBookmarkDialog.FindBookmark(addr + j);
-                    bookmarks |= (bm != nullptr) ? (1 << j) : 0;
-                    freeze |= (bm != nullptr && bm->Frozen()) ? (1 << j) : 0;
-
-                    if (bm != nullptr && bm->Frozen())
+                    if (auto myIter = g_MemBookmarkDialog.FindBookmark(addr + j)
+                        ; myIter == std::end(g_MemBookmarkDialog))
```
SyrianBallaS: this will make sure the iterator is not invalidated early.

src/RA_Dlg_Memory.cpp
```diff
@@ -622,14 +622,21 @@ void MemoryViewerControl::RenderMemViewer(HWND hTarget)
...
                     {
-                        if (g_MemBookmarkDialog.GetHWND() != nullptr)
-                            g_MemBookmarkDialog.WriteFrozenValue(*bm);
+                        break;
+                    }
+                    else {
+                        auto& bm = ra::at(addr + j, myIter->second);
+                        bookmarks |= (1 << j);
```
SyrianBallaS: You can't copy just move or reference (alias).

src/RA_Dlg_Memory.cpp
```diff
@@ -1470,8 +1477,12 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
         }
 
         case WM_CLOSE:
-            EndDialog(hDlg, 0);
-            return TRUE;
+        {
+            // hDlg was invalid sometimes
+            if (hDlg)
+                EndDialog(hDlg, IDCLOSE);
```
SyrianBallaS: The Memory Inspector was open when RA thought I had no connection to the internet. After I closed it said "An invalid handle was specified", so I'm guessing is was invalidated early. nullptr doesn't just mean 0, It's any value not valid for pointer type.

src/RA_Interface.cpp
```diff
@@ -366,15 +367,13 @@ void WriteBufferToFile(const char* sFile, const char* sBuffer, int nBytes)
 void FetchIntegrationFromWeb()
 {
     const int MAX_SIZE = 2 * 1024 * 1024;
-    char* buffer = new char[MAX_SIZE];
-    if (buffer != nullptr)
+
+    if (auto buffer{ std::make_unique<char[]>(MAX_SIZE) }; buffer != nullptr)
     {
```
SyrianBallaS: Save an extra step by destroying it earlier.
It really should be intuitive.
`if precondition; is condition satisfied?` then `postcondition` is automatically handled (destroying the owned pointer in the if..else if..else whatever it may be chain).

src/RA_MemManager.h
```diff
@@ -2,22 +2,12 @@
...
-
-public:
-    unsigned int m_nAddr;
-    unsigned int m_nLastKnownValue;		//	A Candidate MAY be a 32-bit candidate!
-    bool m_bUpperNibble;				//	Used only for 4-bit comparisons
-    bool m_bHasChanged;
+    unsigned int m_nAddr{ 0U };
+    unsigned int m_nLastKnownValue{ 0U };		//	A Candidate MAY be a 32-bit candidate!
+    bool m_bUpperNibble{ false };				//	Used only for 4-bit comparisons
+    bool m_bHasChanged{ false };
```
SyrianBallaS: These had narrowing problems, `BOOL `is a `typedef `of `int `(4 bytes on windows) `bool `is 1 byte everywhere.

PR #59, opened 2018-06-11
show code note tooltips when hovering over memory addresses in the achievement editor

Jamiras:
Works on Mem or Delta fields:
image
If no notes are available, tooltip indicates such:
image


meleu:
:heart_eyes:

PR #61, opened 2018-06-19
Precompiled header (w/ some notes for after this)

SyrianBallaS:
The majority of the lines added are notes in the PCH. They are just reminders on what to do when we'll start resolving the higher level warnings. They will be removed once the warning(s) worked on are resolved.

OK, the purpose of the PR is two-fold. 

1. Make the dll build in less than 5 seconds.
2. Identify problems in the code by compiling with all warnings enabled (these have nothing to do with code analysis)

Total warning count came to about 3500.

So after the PCH was done I disabled all the warnings being generated from library files.
After that I tried fixing up all the warnings that came up for us. Somewhere along the way I realized it would be too big for one PR so I discarded all those changes and disabled all the warnings that came for us as well in the PCH. All warnings have been documented with a strategy to resolve them after this PR is accepted (mainly for myself).

There are some serious problems in our code, the most dire would have to be Spectre vulnerabilities. I already know how to patch them out by using OOP instead of DOP (data-oriented) as was done before. (everything is treated as an object instead of data). Basically `accessing by object` instead of `by index`. Thankfully, spectre was only triggered 8 times in our code and once in ATL that don't use (COM is dangerous but it's already used in our code, COM/RPC is synonymous with ActiveX).

It's not a big deal that I reverted all the fixes because I already know how to resolve them now.


src/RA_Dlg_MemBookmark.cpp
```diff
@@ -651,17 +648,21 @@ void Dlg_MemBookmark::WriteFrozenValue(const MemBookmark & Bookmark)
             break;
     }
 
-    char buffer[32];
-    sprintf_s(buffer, sizeof(buffer), "%0*x", width, Bookmark.Value());
+    // spectre stuff, needs to be a string
+    std::string buffer;
+    buffer.reserve(32);
+    sprintf_s(buffer.data(), 32, "%0*x", width, Bookmark.Value());
+    buffer = buffer.data();
```
SyrianBallaS: This is required before using the ranged for. The string is in memory but hasn't been saved to the object yet.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -118,8 +115,7 @@ void Dlg_AchievementEditor::SetupColumns(HWND hList)
     {
         col.mask = LVCF_TEXT | LVCF_WIDTH | LVCF_SUBITEM | LVCF_FMT;
         col.cx = COLUMN_WIDTH[i];
-        tstring colTitle = COLUMN_TITLE[i];	//	Take non-const copy
-        col.pszText = const_cast<LPTSTR>(colTitle.c_str());
+        col.pszText = NativeStr(COLUMN_TITLE[i]).data();
```
SyrianBallaS: In C++17 as long as the string is not a reference to const std::string::data gives you a non-const c-string.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -553,7 +549,7 @@ BOOL CreateIPE(int nItem, int nSubItem)
             {
                 ComboBox_AddString(g_hIPEEdit, NativeStr(CONDITIONTYPE_STR[i]).c_str());
 
-                if (strcmp(g_AchievementEditorDialog.LbxDataAt(nItem, nSubItem), CONDITIONTYPE_STR[i]) == 0)
+                if (g_AchievementEditorDialog.LbxDataAt(nItem, nSubItem) == std::string{ CONDITIONTYPE_STR[i] })
```
SyrianBallaS: I'm thinking of making a namespaced equality operator for "const CharT*" later.

src/RA_Dlg_Achievement.cpp
```diff
@@ -227,22 +223,19 @@ BOOL LocalValidateAchievementsBeforeCommit(int nLbxItems[1])
...
-                sprintf_s(buffer, 2048, "Achievement title contains an illegal character: '%c'\nPlease remove and try again", cNextChar);
+                sprintf_s(buffer, 2048, "Achievement title contains an illegal character: '%c'\nPlease remove and try again", c);
                 MessageBox(nullptr, NativeStr(buffer).c_str(), TEXT("Error!"), MB_OK);
                 return FALSE;
             }
-            if (strchr(Ach.Description().c_str(), cNextChar) != nullptr)
+            if (Ach.Description().find_first_of(c) != std::string::npos)
             {
-                sprintf_s(buffer, 2048, "Achievement description contains an illegal character: '%c'\nPlease remove and try again", cNextChar);
+                sprintf_s(buffer, 2048, "Achievement description contains an illegal character: '%c'\nPlease remove and try again", c);
```
SyrianBallaS: Forgot to leave a comment in code here but this was another case of spectre vulnerability. There's still 8 in the code but thought I'd just fix them later.

src/RA_Defs.h
```diff
@@ -34,7 +36,6 @@
 #define NOSYSCOMMANDS
 #define OEMRESOURCE
 #define NOATOM
-#define NOCLIPBOARD
```
SyrianBallaS: When this was put in the pch it gave an error. We probably won't need it later when using HeapAlloc instead of GlobalAlloc.

SyrianBallaS:
Opening this back up later.

PR #65, opened 2018-06-25
UI Base Interfaces and Login Dialog (POC)

SyrianBallaS:
Hi all,

This isn't really a POC because I've already done something similar before. It kind of is because I only changed one dialog to see what you guys think about about it before PR'ing more of them (one at a time). Most of the code is added code and documentation comments, very little of existing code was actually changed. I'm using XML comments because it's easier to auto-generate API docs with them. Some message handlers have default functionality while most of them don't and will need to implemented. Messages like WM_COPY/WM_CUT/WM_UNDO and stuff like that don't have defaults because they have variants for ComboBox, ListBox, ListView, and Edit Controls.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 *.user
 /RetroAchievementsKey
 test_app/
+packages/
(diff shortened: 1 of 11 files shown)


SyrianBallaS:
Opening this back up later.

PR #67, opened 2018-07-02
Memory Search filtering improvements

Jamiras:
Extracts the search filtering logic out of the memory viewer dialog and MemManager classes and reduces the memory footprint of stored results by about 90%. 

_NOTE: I did not use any profiling tools to get these numbers, they're estimated from the data structures_

_Before_: `std::vector<MemCandidate>`: 10 bytes per matching item. 
- Initial footprint on SNES (128K) would take 1.3MB of memory. 
- Filtering out half the results would take an additional 650K 
- total = **2MB**.

_After_: `SearchResults`: 
* raw memory captured (bytes only take 8-bits)
* previous value pulled from previous results (no additional storage required)
* size information moved to container (not duplicated on every item)
* modified items kept in vector, instead of having property on every item
- Initial footprint on SNES (128K) is about 130K. 
- Filtering out half the results takes an additional 70K or so 
- total = **200KB**.

The "history" feature keeps track of the last 50 filter results, so if you're not shrinking your results very quickly, the used memory grows quickly. This is particularly noticeable on the N64, which has 4MB of mapped memory, so the initial unfiltered data carves out a 40MB block of memory. After six or seven filterings, you might be using upwards of 200MB. That value is expected to be about 20MB after these changes - still less than the starting footprint with the previous implementation.

[RA_Integration-memsearch.zip](https://github.com/RetroAchievements/RAIntegration/files/2156959/RA_Integration-memsearch.zip)


src/RA_Dlg_Memory.h
```diff
@@ -50,18 +51,19 @@ class MemoryViewerControl
...
 
-class SearchResult
+struct SearchResult
 {
-public:
-    SearchResult() {}
-public:
-    std::vector<MemCandidate> m_ResultCandidate;
-    unsigned int m_nCount = 0;
+    ra::services::SearchResults m_results;
```
Jamiras: `m_results` contains the raw memory and the matching addresses.

src/RA_Dlg_Memory.h
```diff
@@ -50,18 +51,19 @@ class MemoryViewerControl
...
+
+    std::vector<unsigned int> m_modifiedAddresses;
```
Jamiras: `m_modiifiedAddresses` is the list of addresses that should be highlighted.

src/RA_Dlg_Memory.cpp
```diff
@@ -890,23 +881,39 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
...
                         {
                             SetTextColor(pDIS->hDC, RGB(0, 100, 150));
-                            _stprintf_s(buffer, sizeof(buffer), _T("%s"), m_SearchResults[m_nPage].m_sSecondLine.c_str());
+                            unsigned int nMatches = m_SearchResults[m_nPage].m_results.MatchingAddressCount();
+                            if (nMatches > MIN_RESULTS_TO_DUMP)
+                                _stprintf_s(buffer, sizeof(buffer), _T("Found %u matches! (Displaying first %u results)"), nMatches, MIN_RESULTS_TO_DUMP);
+                            else if (nMatches == 0)
+                                _stprintf_s(buffer, sizeof(buffer), _T("Found *ZERO* matches!"));
+                            else
+                                _stprintf_s(buffer, sizeof(buffer), _T("Found %u matches!"), nMatches);
```
Jamiras: rebuild these strings instead of storing them.

src/RA_Dlg_Memory.cpp
```diff
@@ -1040,9 +1048,19 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
...
                     EnableWindow(GetDlgItem(hDlg, IDC_RA_RESULTS_FORWARD), FALSE);
 
-                    unsigned int nValueQuery = 0;
+                    SearchResult& srPrevious = *(m_SearchResults.end() - 2);
+                    SearchResult& sr = m_SearchResults.back();
+                    sr.m_nCompareType = nCmpType;
 
+                    if (IsDlgButtonChecked(hDlg, IDC_RA_CBO_GIVENVAL) == BST_UNCHECKED)
                     {
+                        sr.m_results.Initialize(srPrevious.m_results, nCmpType);
```
Jamiras: pass in the previous result set for a list of addresses to check, and their previous values

src/RA_Dlg_Memory.cpp
```diff
@@ -1121,49 +1129,33 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
...
-                    }
-                    else
-                    {
-                        ClearLogOutput();
-                        m_nPage = 0;
-                        m_SearchResults.clear();
-                        SearchResult sr;
-                        m_SearchResults.push_back(sr);
-                        m_SearchResults[m_nPage].m_sFirstLine = "Invalid Range";
+                        sr.m_results.Initialize(start, end - start + 1, nCompSize);
```
Jamiras: without a previous result set, a range has to be specified.

Issue #21, opened 2018-04-17
Ability to develop achievements offline


Jamiras:
As a developer, I'd like to be able to create achievements without an active internet connection. Currently, to access any of the tools, I have to log in first.

* Minimally, allow updating and testing of Local achievements (no internet required)
* Ideally, allow adding/updating code notes (these would have to be queued and submitted once reconnected).
* Modification of non-Local achievements would not be expected.
* Earning of achievements while offline should be a separate issue if desired, and has its own set of concerns around possible cheating.

Suggest adding a new "Work Offline" menu option to enable the behavior. The user could just click past the "error connecting to retroachievements.org" dialog, or it could be modified to quickly enable the feature. In either case, once enabled, a dialog should be shown letting the user know they're in offline mode and that achievements cannot be earned until they reconnect.

kdecks:
From #112 by @magusLOGS:

A Offline Mode would be nice for those of us who have a unstable internet connection or just want to work/play offline and then sync their progress manually to the server, similar to PSN/XBL.

Developing Offline
Making the entire developing progress offline compatible and allow to save a set/cheevos in one package-file, which can easily be transfered from one device to another device.

Earning Offline
A switch in the Emulators (like Hardcore) which turns on the 'Offline Mode'. Offline Mode would function similar to Softcore but it will require to Sync the earnings to users profile, all progress will be saved locally as well. Also being able to load the one package-file which contains every data for developing and earning cheevos incl. cheevo badges anytime a user want, with saving progress of which cheevo has been earned.

This way users can share their custom sets in form of a single file with other users without uploading their content to the Site.

kdecks:
from @Jamiras, response:

Developing offline is already covered by #21, but it specifically states that earning achievements offline should be a separate issue, so please update this to reflect that.

The package-file concept is a new idea. The plan was just to leverage the existing files in RACache. Will have to think about that more.

PR #69, opened 2018-07-08
allow specifying alternate server via text file

Jamiras:
Looks for a `host.txt` file in the emulator directory and uses whatever site is provided in that file as the host for ALL API calls (except for image requests, which are still hard-coded to go through `i.retroachievements.org`).

If the file is not present, the default value `retroachievements.org` is used.

I tried pointing at `stage.retroachievements.org`, but the API there doesn't seem to be hooked up. The login call returned an HTML page, and the other calls all returned "success false, unknown request".

This should allow for testing changes (i.e. leaderboard AND/OR, commit logs) without affecting the main site.

I've also associated it with #21 - if `OFFLINE` is put in the `host.txt` file, the emulators will call `RA_InitOffline` instead of `RA_Init`. Right now, there's no difference between them, but it's a first step towards enabling separate behavior. Introducing this logic now allows for the emulators to be rebuilt in anticipation of future support without requiring yet another rebuild of the emulators when the future support arrives.


meleu:
Really useful for tests. I'll try to implement something similar on RetroArch at some point.

meleu:
@Jamiras could you post here a binary with the changes of this PR? I'm specially interested in the `OFFLINE` feature.

Jamiras:
This PR specifically adds the ability to point at a different host. (i.e. `stage.retroachievements.org`). _It does not have full `OFFLINE` functionality_. That will be handled by #21. However, as changes were required in `RA_Interface.cpp`, a new build of the emulators is required to support this feature (it tries to validate the toolkit version from within the emulator, and we want to get that data from the alternate host as well). Since I was changing the emulator interface code, I thought I'd stub in the `OFFLINE` support so it can be implemented later without requiring another new build of the emulators.

With these changes, setting the host to `OFFLINE` will cause the toolkit to try to communicate with a server at `http://OFFLINE`. The remaining changes will be implemented in #21.

Do you still want a binary?

meleu:
Oh, if I will still be unable to use the Memory Inspector while offline then no. No need to post a binary.

Thanks for clarifying.

SyrianBallaS:
This seems extremely useful. Initial testing on a dev/staging server should be preferable to avoid accidents on the production server until the changes are ready.

src/RA_Core.cpp
```diff
@@ -400,7 +428,9 @@ API int CCONV _RA_OnLoadNewRom(const BYTE* pROM, unsigned int nROMSize)
             //	Some other fatal error... panic?
             ASSERT(!"Unknown error from requestgameid.php");
 
-            MessageBox(g_RAMainWnd, NativeStr("Error from " RA_HOST_URL "!\n").c_str(), TEXT("Error returned!"), MB_OK);
+            std::ostringstream oss;
+            oss << "Game not loaded.\nError from " << _RA_HostName() << "!";
+            MessageBox(g_RAMainWnd, oss.str().c_str(), TEXT("Error returned!"), MB_OK | MB_ICONERROR);
```
SyrianBallaS: Be careful with the `oss.str().c_str()` part. Sometimes it dangles (appends garbage). I'd suggest using a "`tostringstream`" since `MessageBox `could be `ASCII `or `Unicode`.
like: `using tostringstream = std::basic_string<TCHAR>;`

src/RA_Core.cpp
```diff
@@ -503,22 +533,15 @@ API void CCONV _RA_ClearMemoryBanks()
...
-    //strcpy_s( g_sKnownRAVersion, 50, sNewVer );
-
-    if (MessageBox(g_RAMainWnd, NativeStr(buffer).c_str(), TEXT("Update available!"), MB_YESNO) == IDYES)
+    std::ostringstream oss;
+    oss << "Would you like to update?\n\n"
+        << "A new version of " << g_sClientName << " is available for download at " << _RA_HostName() << ".\n\n"
+        << "Current version: " << g_sClientVersion << "\n"
+        << "New version: " << sNewVer;
+
+    if (MessageBox(g_RAMainWnd, NativeStr(oss.str()).c_str(), TEXT("Update available!"), MB_YESNO | MB_ICONINFORMATION) == IDYES)
```
SyrianBallaS: There was a different comment regarding using `oss.str().c_str()`. I think the way you handled it here is good.

src/RA_Core.cpp
Jamiras: Good catch. I've wrapped it in a `NativeStr` call for Unicode support.

I'm not sure what you mean by "dangles (appends garbage)". `str()` returns a `std::string` object, which should live long enough for the `MessageBox` call. If not, using `NativeStr` isn't going to help because it does the same thing.

PR #74, opened 2018-07-18
Window Message Handler POC and Level 4 Warning Fixes

SyrianBallaS:
Basically what this does is make a stateless controller to work as a helper for processing Windows Messages. I've only applied this concept to the Login and ROM Checksum Dialog. There will be additional review comments from myself. The main reason I made it stateless so it would be easier to cast to void* if we wanted to make a C API out of it.


src/RA_Dlg_Login.cpp
```diff
@@ -5,118 +5,150 @@
...
+        }
     }
+};
 
+INT_PTR Dlg_Login::DoModal()
+{
+    return DialogBoxParam(::GetModuleHandle(_T("RA_Integration.dll")), MAKEINTRESOURCE(IDD_RA_LOGIN),
+                          ::GetAncestor(::GetActiveWindow(), GA_ROOT), Dlg_LoginProc,
+                          reinterpret_cast<LPARAM>(this));
+}
```
SyrianBallaS: Figured this was one way to limit globals, so far what I've noticed is the only difference between these "DoModal" is the resource ID, and maybe the the parent window (Memory Bookmarker).

`::GetAncestor(::GetActiveWindow(), GA_ROOT)` is `g_RAMainWnd`.
It was called this way because ::GetActiveWindow() by itself might grab the wrong handle.

`reinterpret_cast<LPARAM>(this)` gives a default WindowClass, might not seem too useful here but when these dialogs get abstracted it will be required because the `Message Queue` will need to differentiate between the Dialogs or it will call the wrong procedure.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -503,26 +503,26 @@ void Dlg_MemBookmark::PopulateList()
...
-        ASSERT(item.iItem == m_nNumOccupiedRows);
-
+        // kind of like zero memory
+        auto lpItem{ std::make_unique<LV_ITEM>() }; 
+        lpItem->mask       = LVIF_TEXT;
+        lpItem->iItem      = m_nNumOccupiedRows;
+        lpItem->cchTextMax = 256;        
+
+        lpItem->iItem = ListView_InsertItem(hList, lpItem.get());
+        ASSERT(lpItem->iItem == m_nNumOccupiedRows);
```
SyrianBallaS: There was a warning about the enumerable not being used. So we did the same thing just with iterators. (you need a `begin `and `end `to use ranged for).

A ranged for works exactly like `foreach `in C#, there is a VC++ "`for each`" that works and looks exactly like it but it's not portable.

src/RA_Dlg_RomChecksum.cpp
```diff
@@ -2,53 +2,76 @@
...
     {
-        case WM_INITDIALOG:
-            SetDlgItemText(hDlg, IDC_RA_ROMCHECKSUMTEXT, NativeStr(std::string("ROM Checksum: ") + g_sCurrentROMMD5).c_str());
-            return FALSE;
+        // It seems hwndFocus is the thing with the default focus, (usually OK button)
+        using WindowH = std::unique_ptr<HWND__, decltype(&::DestroyWindow)>;
+        WindowH hRomChecksum{ ::GetDlgItem(hwnd, IDC_RA_ROMCHECKSUMTEXT), ::DestroyWindow };
+
+        // just pray it doesn't leak, we can't destroy it or we won't see anything
+        OnSetText(hRomChecksum.release(), NativeStr("ROM Checksum: " + g_sCurrentROMMD5).c_str());
```
SyrianBallaS: We'd need to destroy during `WM_NCDESTROY `but it wasn't used before so I didn't but it in. This is just a POC.

src/RA_Dlg_Login.cpp
```diff
@@ -5,118 +5,150 @@
...
+
+INT_PTR CALLBACK Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
+{
+    switch (auto helper = std::make_unique<LoginHelper>(); uMsg)
+    {
+        HANDLE_MSG(hDlg, WM_INITDIALOG, helper->OnInitDialog);
+        HANDLE_MSG(hDlg, WM_COMMAND, helper->OnCommand);
+        HANDLE_MSG(hDlg, WM_CLOSE, helper->OnClose);
```
SyrianBallaS: It seems we can use `DestroyWindow `on modals since it will always get destroyed and recreated. I made a variant `OnCloseModal `that used `EndDialog `instead. It might also be because we destroyed all of this dialog's controls before this message was reached. In the checksum dialog I couldn't delete them ahead of time so I used the modal version.

src/ra_wm_handler.h
```diff
@@ -0,0 +1,132 @@
...
+// We can make this an abstract interface later
+/// <summary>This class serves as a base Windows Message handler.</summary>
+/// <remarks>
+///   Only Messages that are guaranteed to be processed or have the same behavior
+///   will be here. Any additional handlers should derive this class
+/// </remarks>
+struct WM_Handler
+{
+#pragma region Constructors
+    WM_Handler() noexcept = default;
```
SyrianBallaS: Default constructor is not `constexpr `so it can be inherited

src/ra_wm_handler.h
```diff
@@ -0,0 +1,132 @@
...
+    inline constexpr WM_Handler(WM_Handler&&) noexcept = default;
+    inline constexpr WM_Handler& operator=(WM_Handler&&) noexcept = default;
+#pragma endregion
+
+    // Basic rule: Messages with pointers (the first hwnd is not part of it) are
+    //             sent, otherwise they are posted. If you don't want it to processed immediate post it the message queue, but make sure it has no pointers.
+#pragma region Event Handlers
+    // This base method should only be used in the dialogs message queue and not it's procedure 
+    _NODISCARD virtual BOOL OnInitDialog(_In_ HWND hwnd, _In_ HWND hwndFocus, _In_ LPARAM lParam) noexcept;
+    _NORETURN virtual void OnCommand(_In_ HWND hwnd, _In_ int id, _In_ HWND hwndCtl, _In_ UINT codeNotify) noexcept;
```
SyrianBallaS: All of these could be const qualified but didn't want to force it, just in case someone wants to make a derived class stateful.

src/ra_wm_handler.cpp
```diff
@@ -0,0 +1,227 @@
...
+{
+    ::SetFocus(hwndFocus);
+    // If it returns nonzero, the focus will go back to hwnd
+    return ::SetWindowLongPtr(hwnd, DWLP_USER, lParam);
+}
+
+_Use_decl_annotations_
+void WM_Handler::OnCommand(HWND hwnd, int id, HWND hwndCtl, UINT codeNotify) noexcept
+{
+    return FORWARD_WM_COMMAND(hwnd, id, hwndCtl, codeNotify, SendMessage);
```
SyrianBallaS: Be careful with these "`Send/PostMessage`" defaults, they shouldn't be called directly in the Proc. However, it should be fine to use them while processing messages for a different window. Should these be protected?

src/ra_wm_handler.cpp
```diff
@@ -0,0 +1,227 @@
...
+{
+    FORWARD_WM_ENTERIDLE(hwnd, source, hwndSource, ::DefWindowProc);
+}
+
+_Use_decl_annotations_
+void WM_Handler::OnNCPaint(HWND hwnd, HRGN hrgn) noexcept
+{
+    using WindowDC = std::unique_ptr<HDC__ , decltype(&::DeleteDC)>;
+    WindowDC hdc{ GetDCEx(hwnd, hrgn, DCX_WINDOW | DCX_INTERSECTRGN), ::DeleteDC };
+    ReleaseDC(hwnd, hdc.release());
```
SyrianBallaS: We did our part in not making it leak.
If Windows leaks it so be it. We can't use `ReleaseDC `as a deleter because `HDC `is a handle and doesn't have constructors.

src/ra_wm_handler.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+_Use_decl_annotations_
+BOOL WM_Handler::OnEraseBkgnd(HWND hwnd, [[maybe_unused]] HDC hdc) noexcept
+{
+    // if it returns 0, the background will be erased if it's GCLP_HBRBACKGROUND it will not be
+    return ::GetClassLongPtr(hwnd, GCLP_HBRBACKGROUND);
+}
```
SyrianBallaS: I did it this way to prevent accidents, I know for sure what will happen with this result but not sure how to do manually.

src/ra_wm_handler.cpp
```diff
@@ -0,0 +1,227 @@
...
+void WM_Handler::OnRetry(HWND hDlg) noexcept { ::EndDialog(hDlg, IDRETRY); }
+
+_Use_decl_annotations_
+void WM_Handler::ButtonCheck(HWND hwnd) noexcept { Button_SetCheck(hwnd, BST_CHECKED); }
+
+_Use_decl_annotations_
+void WM_Handler::ButtonUncheck(HWND hwnd) noexcept { Button_SetCheck(hwnd, BST_UNCHECKED); }
+
+_Use_decl_annotations_
+void WM_Handler::DisableButtonCheck(HWND hwnd) noexcept { Button_SetCheck(hwnd, BST_INDETERMINATE); }
```
SyrianBallaS: This might be a better approach instead using `WM_CANCELMODE`.

src/ra_wm_handler.h
```diff
@@ -0,0 +1,132 @@
+#ifndef RA_WM_HANDLER_H
+#define RA_WM_HANDLER_H
+#pragma once
+
+#include <WTypes.h>
+#include <yvals.h>
```
SyrianBallaS: for `_NODISCARD`

src/ra_wm_handler.h
```diff
@@ -0,0 +1,132 @@
...
+//    _NORETURN virtual void OnDestroyClipboard(_In_ HWND hwnd) noexcept;
+//    _NORETURN virtual void OnDrawClipboard(_In_ HWND hwnd) noexcept;
+//#pragma endregion
+//
+//};  
+#pragma endregion
+
+// Mostly for identifying what the window/control is for debugging but could be used normally
+// These can't be used in the WindowProc user callbacks though
+_Success_(return != nullptr)
```
SyrianBallaS: I should have added it can be used with `FindWindow `to avoid globals.
Was thinking of making one to get the `Window Class` name in the same fashion but most of the dialogs don't have one

src/RA_AchievementOverlay.cpp
```diff
@@ -114,7 +114,7 @@ AchievementOverlay::~AchievementOverlay()
     }
 }
 
-void AchievementOverlay::Initialize(HINSTANCE hInst)
+void AchievementOverlay::Initialize(_UNUSED HINSTANCE hInst)
```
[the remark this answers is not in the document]
rzumer: I'm not familiar with this function, but my first thought is that this is hiding the warning rather than solving the issue it is pointing at.

If there is a legitimate reason to conceal the warning, then please document it here.

Is there no native warning suppression system?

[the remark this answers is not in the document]
rzumer: LGTM then.

.gitignore
```diff
@@ -5,3 +5,4 @@
 *.user
 /RetroAchievementsKey
 test_app/
+packages/
```
rzumer: This needs to be reverted, ignore your packages locally if you are testing things.

SyrianBallaS:
Mostly handled by other PRs in different form, closing this.

PR #76, opened 2018-07-24
Move runtime logic to external library to be shared with RetroArch

Jamiras:
* incorporates Lua for build, but it's not yet used in the runtime


rzumer:
If it's not too difficult, I would favor including https://github.com/lua/lua as a submodule and including it in the build.

Jamiras:
I've rebuilt this branch with lua as a submodule. I used `filter-branch` and a force push to remove the library references. If you had this branch checked out, you will have to delete it and check it out again to get the new history.

rzumer:
Thanks!

Issue #12, opened 2018-04-15
Ability to specify string for "unknown" value in RichPresence Lookup


meleu:
_From @Jamiras on March 21, 2018 15:24_

As a developer, I'd like to have a catch-all value in my lookup to minimize repetition:
```
Lookup:Level
1=Level 1
2=Level 2
3=Level 3
4=Titles
5=Titles
6=Titles
7=Titles
```
could be simplified to:
```
Lookup:Level
1=Level 1
2=Level 2
3=Level 3
*=Titles
```

_Copied from original issue: RetroAchievements/RASuite#65_

meleu:
_From @Jamiras on March 21, 2018 15:25_

Note that this is less of an issue now with conditional rich presence:
```
?byte(0x1234)>3?Player is in the titles
Player is in @Level(byte(0x1234))
```

PR #80, opened 2018-08-04
Support for Rich Presence lookup default and comments, add unit tests

Jamiras:
Moved reading/writing the `XXX-Rich.txt` file outside of the interpreter.

Added support for:
* default item in lookup map (#12)
* official comment format `//`
```
// This is an example Rich Presence script for Some Game written by User
Lookup:Level
1=Level 1-1
2=Level 1-2
*=Titles           // Anything that's not explicitly defined will display "Titles"

Display:
@Level(0x1234)
```

### DIFF
--- a/src/RA_Core.cpp
+++ b/src/RA_Core.cpp
@@ -318,6 +318,7 @@ API int CCONV _RA_Shutdown()
 
     if (g_RichPresenceDialog.GetHWND() != nullptr)
     {
+        g_RichPresenceDialog.ClearMessage();
         DestroyWindow(g_RichPresenceDialog.GetHWND());
         g_RichPresenceDialog.InstallHWND(nullptr);
     }
@@ -484,6 +485,7 @@ API int CCONV _RA_OnLoadNewRom(const BYTE* pROM, unsigned int nROMSize)
     g_MemoryDialog.OnLoad_NewRom();
     g_AchievementOverlay.OnLoad_NewRom();
     g_MemBookmarkDialog.OnLoad_NewRom();
+    g_RichPresenceDialog.OnLoad_NewRom();
 
     g_nProcessTimer = 0;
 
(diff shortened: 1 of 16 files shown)


src/RA_Core.cpp
```diff
@@ -1379,7 +1379,18 @@ API void CCONV _RA_InvokeDialog(LPARAM nID)
                 sprintf_s(sRichPresenceFile, 1024, "%s%u-Rich.txt", RA_DIR_DATA, g_pCurrentGameData->GetGameID());
 
                 //	Then install it
-                g_RichPresenceInterpreter.ParseRichPresenceFile(sRichPresenceFile);
+                std::string sRichPresence;
+                std::ifstream file(sRichPresenceFile);
```
rzumer: How about a `_ReadBufferFromFile()` function to mirror the use of `_WriteBufferToFile`? That will also make factoring out file operations and other supporting functions out of this large file easier in the future.

src/RA_Dlg_RichPresence.cpp
```diff
@@ -20,7 +21,12 @@ INT_PTR CALLBACK Dlg_RichPresence::RichPresenceDialogProc(HWND hDlg, UINT nMsg,
 
         case WM_TIMER:
         {
-            std::wstring sRP = ra::Widen(g_RichPresenceInterpreter.GetRichPresenceString());
+            std::wstring sRP;
+            if (g_pCurrentGameData->GetGameID() != 0)
```
rzumer: What's the rationale behind extracting logic into the dialog window? That tends to be considered bad practice.

src/RA_RichPresence.cpp
```diff
@@ -1,266 +1,260 @@
...
-    return sUnknown;
+    return m_sDefault;
 }
 
-RA_ConditionalDisplayString::RA_ConditionalDisplayString(const char* pBuffer)
+RA_RichPresenceInterpreter::DisplayString::DisplayString()
 {
-    ++pBuffer; // skip first question mark
+    // add an empty core group to ensure Test() always returns true
+    m_conditions.AddGroup();
```
rzumer: Won't this cause problems during general use? Setting an error state seems safer (and possibly cheaper) than adding or removing groups.

src/RA_RichPresence.cpp
```diff
@@ -1,266 +1,260 @@
...
-    {
-        // valid condition
-        *pBuffer++;
-        m_sDisplayString = pBuffer;
-    }
-    else
+    if (*pBuffer != '\0')
     {
         // not a valid condition, ensure Test() never returns true
         m_conditions.Clear();
```
rzumer: Idem.

src/RA_RichPresence.cpp
```diff
@@ -1,266 +1,260 @@
...
-
-    bool bParsingLookupName = false;
-    bool bParsingLookupContent = false;
-
-    std::string sLookupName;
-    std::string sLookupValue;
+        size_t index = sLine.find("//");
+        if (index != std::string::npos)
+        {
+            // if a comment marker was found, remove it and any trailing whitespace
```
rzumer: If comments can be inserted in the middle of a line, an escape character is required in order to be able to display the marker.

src/RA_RichPresence.cpp
rzumer: Also, can you clarify the goal of this new function? Is it to display an empty message?

src/RA_Core.cpp
Jamiras: will do

src/RA_Dlg_RichPresence.cpp
Jamiras: It's not really a property of the interpreter. That should return whatever the currently loaded script generates. If no script is loaded, empty string is returned. 

Until we move to an MVVM or MVC architecture, I don't see any better place to put this. I will move it up one more step so its set once without using the timer.

src/RA_RichPresence.cpp
Jamiras: This isn't actually an error state. This constructor is for the default display string (no condition). The other constructor is for a conditional display string. After either constructor is called, `InitializeParts` is called to process the display string itself.
```
        for (std::map<std::string, std::string>::const_iterator iter = mDisplayStrings.begin(); iter != mDisplayStrings.end(); ++iter)
        {
            auto& displayString = m_vDisplayStrings.emplace_back(iter->first);
            displayString.InitializeParts(iter->second, mFormats, m_vLookups);
        }

        auto& displayString = m_vDisplayStrings.emplace_back();
        displayString.InitializeParts(sDisplayString, mFormats, m_vLookups);
```
`InitializeParts` could be folded into the constructor.

Jamiras: Regarding the "error state": there's actually three states that have to be supported:
1) no condition (always true)
2) invalid condition (always false)
3) valid condition (must be evaluated)

An empty group always evaluates true, a condition with no groups always evaluates false. By doing it this way, the Test logic can just evaluate the condition field without having to check a separate state variable.

src/RA_RichPresence.cpp
Jamiras: `//` style comments can't appear in the middle of a line. It is very unlikely that a `//` literal would be desired.

src/RA_RichPresence.cpp
rzumer: My mistake, I didn't realize it was a constructor.

src/RA_RichPresence.cpp
rzumer: From what I can tell, any comment will be stripped from the line before it is interpreted, and comment markers are searched using string::find(), so comments can be inserted in the middle of a line. In fact, your example in this PR's description includes one. Is this incorrect?

Regardless of how likely one is to want to display `//` as part of a script, I don't think it's fair to remove that possibility. For this version I think it would be best to only strip comments found at the beginning of a line, and restore support for comments that begin in the middle of a line along with some escape sequence.

rzumer: From what I can tell, any comment will be stripped from the line before it is interpreted, and comment markers are searched using string::find(), so comments can be inserted in the middle of a line. In fact, your example in this PR's description includes one. Is this incorrect?

Regardless of how likely one is to want to display `//` as part of a script, I don't think it's fair to remove that possibility. For this version I think it would be best to only strip comments found at the beginning of a line, and restore support for comments that begin in the middle of a line along with some escape sequence.

Jamiras: You're right, `//` in the middle of a line acts as a line terminator.

The "`@`" marker for a lookup/value in a display string is just as destructive. If the interpreter sees that it looks for an open and close parenthesis to identify the parameter to the lookup:
```
The hero is @ here.
```
Will always generate "The hero is " because the lookup definition is considered incomplete.
```
The hero is @ @Location(0x1234).
```
Will also generate "The hero is ." because it sees " @Location" as the name of the lookup, not "Location". And an undefined lookup generates an empty string.

`//` can be worked around by adding an invalid lookup: `/@()/`. To work around the `@` sign, you have to define a lookup that always evaluates to `@`:
```
Lookup:At
*=@

Display:
The hero is @At(0x0000) here.
```
This was even harder before this PR as you'd have to find a memory value that never changed so the lookup would map to the `@` entry.

Display strings are really the only time escaping would be necessary. I'll modify `GetLine` to ignore a `//` immediately following a `\` and modify `InitializeParts` to treat any character following a `\` literally.

rzumer: Sounds good, thanks.

PR #81, opened 2018-08-05
don't hold images in memory when they're not needed

Jamiras:
* only pre-fetches badge images for achievements the user hasn't earned
* on-demand fetches badge images (and their companion `_lock` images) as needed for overlay and achievement editor
* doesn't hold `overlayBG.png` in memory when overlay is not visible (saves 4MB of RAM)
* removes HBITMAP references from Achievement and User classes
* fixes issue where user image was not being displayed on welcome message
* fixes friends overlay always being empty


src/RA_AchievementSet.cpp
```diff
@@ -85,6 +85,14 @@ void AchievementSet::OnRequestUnlocks(const Document& doc)
         else if (g_pUnofficialAchievements->Find(nNextAchID) != nullptr)
             g_pUnofficialAchievements->Unlock(nNextAchID);
     }
+
+    // pre-fetch images for any achievements the player hasn't earned
```
rzumer: Have you tested this? I know that in some situations it is possible to "re-earn" an achievement (the server will return an error if I remember correctly), what will happen with the image in that case?

Jamiras: All images can be fetched on demand. The thought process here is that achievements that haven't been earned will need the image when they are earned, so get them ahead of time so they're ready. Previously ALL images were pre-fetched, including the `_lock` variations.

I don't believe it's actually possible to recreate the scenario you've described. In order to activate an achievement after this logic has ran, you have to select it in the achievements list, which fetches the image for display in the achievement editor. In this situation, the image is ready for when the achievement does trigger and does display properly.

The on-demand loading can be seen in the overlay. Placeholder icons will appear and slowly be filled in as the actual images are fetched. If there were a way to activate the achievement without selecting it in the achievements list, I would expect the same behavior in the popup.

Jamiras: Replicated the described behavior. After starting a game, deleted all pre-downloaded images. When achievement triggered, placeholder was displayed while image was downloaded. Placeholder was replaced as soon as download finished. Unfortunately, I could not capture screenshots of both states.

image

Jamiras: The average file size for a badge is 5KB. With 40 badges in a game, that's 200KB for the normal, or 400KB if also fetching the greyed out ones. A future PR could even just do the greyscale adjustment on the client.

In addition to potentially saving user disk space, this will save bandwidth. In particular, not fetching the greyed out images saves half the download and the only place they'll ever be used is in the overlay, which I suspect most players never open - I usually only do as a side-effect of hitting ESC to pause the game.

Assuming users never clear their cache, they'll download all of the badges the first time they play. Only downloading unearned badges only really applies if the user deletes their cache, reinstalls, or switches to another machine. In order to create this behavior, I had to explicitly delete the cached files while the game was running.

Jamiras:
> Saving a few kilobytes during the initial transfer, which will likely be fetched later on, doesn't strike me as an improvement

They won't "likely be fetched", which is why they're not being fetched. The typical player only ever sees the images for achievements they earn, which are exactly what are being pre-fetched. The additional images are only used in the overlay, at which point the application can decide whether it needs to fetch the colored icon for achievements the player has already earned or the disabled icon for achievements they haven't earned.

> Was this behavior ever requested?

Not specifically, but I hear a lot of general complaints about how much memory and disk the application consumes. This commit frees 4MB of memory that was always being consumed previously, and for the average player, uses half as much disk space for badges. Something else I've been considering is flushing old cache data - anything older than say three months, but that would be in a separate PR.

> There has been talk of custom locked icons if I recall correctly.

I don't remember this conversation, nor am I proposing eliminating the `_lock` files. I merely suggested that a client-side filter could be applied.

src/RA_AchievementSet.cpp
rzumer: Do we have any data on how many people use the overlay? I don't either, but that doesn't mean it's typical.

>I hear a lot of general complaints about how much memory and disk the application consumes
Do you have any examples of that by any chance?

Jamiras: Given how broken the overlay is (everything but the main achievements page has issues of one kind or another), I'd wager the number is pretty small, but no, I don't have any actual numbers. Nor do I know of any way to get those numbers as the data is fetched from the server at login, not when the overlay is displayed (more opportunities for improvement).

I also don't have any specific examples of complaints about memory usage, as discord's search isn't being helpful with multi-term queries and searching on "memory" alone has too many unrelated hits.

Regarding your request to "just download it all". You win. Deleting old cache data has a much better return on savings in the long run. Once a user has finished a game, they aren't likely to return to it, so the 400K of images can be cleaned out later, rather than trying to smartly download some portion of the 400K up front. But that'll be something for a future PR.

I'll change the logic to download the colored icon for all achievements and the greyed icon only for unachieved achievements. The only reason the greyed out icon would be needed after a user earned an achievement is if they reset the earned state within the emulator as described above, at which point the greyed out image would need to be fetched if the user looked at the reset achievement in the overlay.

Unless a user has earned an achievement on another machine or previous installation, this will behave exactly the same as it does now. A user starting a game with no earned achievements will download both the regular and greyed out icon for every badge in the game.

The other changes in this PR still address the runtime memory consumption. Images will not be loaded from disk unless they're needed. Previously, the overlay (4MB) and every badge (both normal and grey - 32KB for each achievement) would be loaded in memory all of the time.

rzumer: That makes sense to me. Great work on the memory savings.

PR #83, opened 2018-08-09
correctly display UTF-8 characters in popup messages

Jamiras:
Before:
image

After:
image

### DIFF
--- a/src/RA_AchievementPopup.cpp
+++ b/src/RA_AchievementPopup.cpp
@@ -143,8 +143,8 @@ void AchievementPopup::Render(HDC hDC, RECT& rcDest)
         //	meh
     }
 
-    const std::wstring sTitle = L" " + ra::Widen(ActiveMessage().Title()) + L" ";
-    const std::wstring sSubTitle = L" " + ra::Widen(ActiveMessage().Subtitle()) + L" ";
+    const std::wstring sTitle = L" " + ActiveMessage().Title() + L" ";
+    const std::wstring sSubTitle = L" " + ActiveMessage().Subtitle() + L" ";
 
     SelectObject(hDC, hFontTitle);
     TextOutW(hDC, nTitleX, nTitleY, sTitle.c_str(), sTitle.length());
--- a/src/RA_AchievementPopup.cpp
+++ b/src/RA_AchievementPopup.cpp
@@ -143,20 +143,20 @@ void AchievementPopup::Render(HDC hDC, RECT& rcDest)
         //	meh
     }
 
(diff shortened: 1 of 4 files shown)


Jamiras:
We're still a long ways from full unicode support, but it's a step.

rzumer:
I noticed recently that non-ASCII characters in achievement titles (and likely other fields) are not recovered correctly from my local cache files. Do you know what introduced this issue? It was not there six months ago. It seems related to this issue.

As you can see, despite the corrupted text in the achievement editor here, the overlay displays the correct text string:
image

On the other hand, with your change, the overlay will crash the application upon trying to process achievements that contain non-ASCII characters (for example, changing the achievement title in the editor to the correct accented letters, then unlocking it, will do that).

As it turns out, the overlay is able to display non-ASCII characters just fine on the master branch. I think that the scrambled text is a result of a wrong conversion elsewhere, possibly in multiple places. But I don't think that this is an appropriate solution.

Jamiras:
Data received from the server is assumed to be in UTF-8. The local file is stored as ANSI. This causes problems when using the editor to create UTF-8 data, but data (such as the title) can be entered safely through the website.

I added an achievement with the unicode string from your example and saved a local file, resulting in this:
```
0.030
AstÃ©rix
0:0xH0000=0:Déjà vu:Déjà vu: : : :Jamiras:0:0:0:0:0:00000
```
Note the utf-8 sequence in the title (from the server) but the ANSI characters in the achievement. If I try to display this with my changes, it causes errors (will fix). Reloading the game (and the file) shows the correct ANSI characters in the achievement list. 

If I manually change the encoding of the local file to UTF-8 and reload the game, I see the sequences in the editor and the popup displays correctly with my changes:

image

With a UTF-8 local file on master, I see the utf-8 sequences in both places:

image

How did you get the UTF-8 sequences in your editor such that they show up as a UTF-8 sequence in the editor, but correctly in the popup?

Jamiras:
With the fix and a ANSI local file:

image

rzumer:
That's better, but having to choose between broken characters in the overlay or in the rest of the application (achievement editor at least) is a regression, so I would rather pull a change that solves the issue in both.

Jamiras:
I'm going to close this based on @luchaos's comment in Discord:
> the astérix games and other titles are not all correctly stored in db, nor converted properly in the api
there‘s a lot of sanitization work to be done

I'll migrate https://github.com/RetroAchievements/RAIntegration/pull/83/commits/6ee55e7a9f7affcc36466d9fbbfc5f79f5d06789 into one of the other PR's. 

I'm still curious how you managed to get it to display differently in the popup vs. the window.

Issue #79, opened 2018-08-02
Linking or caching issues in certain conditions


rzumer:
Launching `RANes` from a search (through the Windows 10 Start Menu for example) causes the following issues to occur:
* Credentials cannot be retrieved and sessions cannot be restored
* Achievement lists cannot be retrieved (silent failure)
* Most items in the menu indicate that no ROM is loaded, despite an ID being shown in the Achievement Sets window
* An ID of 0 is sent to the server, and an empty entry is shown in the recently played games list on the logged in user's profile

Logging on and linking a hash to a database entry work correctly.

When launched from Windows Explorer by navigating to the directory of the executable, none of these issues are present.

Jamiras:
My Windows 10 Start Menu was unable to find RANes. Apparently its set up to only index my %%user%% directory, so I copied the RANes folder into there and tried again.

Windows 10 Search found the copy in its new location and started the app. I was prompted to login, which I did, then the app closed. I never got to any of the behaviors you described.

Replacing the binary with a custom build, it reported that the program was running in the "C:\WINDOWS\system32" directory. Not running as an administrator, the program doesn't have write permissions there. When I did run as administrator, it told me it was still running in "C:\WINDOWS\system32", but all the supplimental files were written to "C:\WINDOWS\SysWOW64".

Once I had the supplemental files in place, I tried running without administrator again. Since my credentials were saved, it didn't immediately crash. It did crash when I tried to open the same game that I was able to successfully open when running as administrator.

I was unable to recreate any of the issues you've described, but did see a number of issues related to not having write permission to the directory the application thought it was running in. 

RANes and RASnes9x root all of their paths with an internal variable derived from the GetModuleFileName WinApi result. I didn't check the other emulators to see if they're doing something similar, but I expect they probably are. 

RAIntegration uses GetCurrentDirectory to root its paths. It seems like we could mostly fix this by changing the GetCurrentDirectory call to GetModuleFileName. However, that only solves the issue once RAIntegration is loaded. The code that loads RAIntegration.dll expects to find the existing DLL in the current directory and writes the downloaded DLL to the current directory. This would also have to be changed to address the problem, and would require distributing new versions of each emulator to leverage those changes.

PR #85, opened 2018-08-13
Use executable path as home directory

Jamiras:
* Changes the home directory from the "startup directory" to the startup executable's directory.
* Eliminates all calls to SetCurrentDirectory and replaces relative paths with absolute paths.

Additionally adds similar logic to the LoadLibrary/download functionality for the emulators. Emulators will have to be rebuilt to pick up that functionality. 

Have tested using the "Windows Search" functionality described in #79. Requires updated emulator and DLL, but these changes seem to make the application functional whereas I couldn't load a game with the prior code. Hopefully these changes address the other issues original reported as I was not able to recreate them.

### DIFF
--- a/src/RA_AchievementOverlay.cpp
+++ b/src/RA_AchievementOverlay.cpp
@@ -1506,8 +1506,9 @@ void AchievementOverlay::InstallNewsArticlesFromFile()
 {
     m_LatestNews.clear();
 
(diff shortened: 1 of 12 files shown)


rzumer:
It seems to work correctly with these changes, judging from a cursory overview of the features that were not working in the original issue report.

PR #89, opened 2018-08-18
support non-ANSI characters in code notes (fixes #88)

### LINKED ISSUE #88: Non-ASCII characters are not sent correctly in code notes
This is a regression, as this was working as of January. Characters outside of standard extended ASCII, such as `ō`, are either replaced by standard ASCII "equivalents" (in this case by `o`), or by interrogation marks (at least for Chinese characters, e.g. `好`).

The database was updated recently to support UTF-8 characters in code notes. They are currently not displayed correctly on the website, but this will be fixed in a future update (likely 2.0.0). In the meantime, support should be restored on the side of RAIntegration as well.

A secondary issue is that characters outside of standard ASCII, i.e. extended characters such as `ö`, now return an error despite being sent and processed correctly by the server. This can be moved to a separate issue once confirmed if the solution is different from that of the main issue.

Jamiras:
Note that this specifically addresses code notes. Game titles are known to be broken (see #83). Achievement titles/descriptions/leaderboards/etc have not been tested and most likely do not work correctly.

### DIFF
--- a/src/RA_Dlg_Memory.cpp
+++ b/src/RA_Dlg_Memory.cpp
@@ -1185,8 +1185,8 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
                     HWND hMemWatch = GetDlgItem(hDlg, IDC_RA_WATCHING);
                     HWND hMemDescription = GetDlgItem(hDlg, IDC_RA_MEMSAVENOTE);
 
-                    TCHAR sNewNoteWide[512];
-                    GetDlgItemText(hDlg, IDC_RA_MEMSAVENOTE, sNewNoteWide, 512);
+                    WCHAR sNewNoteWide[512];
+                    GetDlgItemTextW(hDlg, IDC_RA_MEMSAVENOTE, sNewNoteWide, 512);
                     const std::string sNewNote = ra::Narrow(sNewNoteWide);
 
                     const ra::ByteAddress nAddr = MemoryViewerControl::getWatchedAddress();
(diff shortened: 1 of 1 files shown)


rzumer:
Two issues observed so far: firstly, the edit dialog prompt, and the memory inspector, display scrambled characters instead of the correct ones. However, the memory inspector decodes correctly when the memory address is entered or edited manually. It is only when selecting the address from the dropdown that it displays the wrong result. It seems to me that there is some inconsistent duplicated logic.

体力

Besides this display issue, the characters appear to be sent correctly (at least they are displayed correctly on the front-end), so thank you for this fix.

For extended ASCII characters, such as basic accented characters including `ö`, a new issue seems to have come up. I can confirm that it is present in the current release version as well. On the Web front-end, they are not decoded correctly, and on the RAIntegration end, editing a note with such a character results in the message "Could not save note! Please check you are online and retry." regardless of success. Finally, RAIntegration silently fails to restore code notes from the server when any one of them contains one such character.

Jamiras:
Your "scrambled" characters are just the ANSI representation of the UTF-8 multi-byte sequence. As you suggested, there were simply more places in the code referencing the notes without properly converting them.

I'm not sure what you're describing in your second issue. I was able to create a note with ö, save it without error, and view it in the web page.

rzumer:
> Your "scrambled" characters are just the ANSI representation of the UTF-8 multi-byte sequence.

Thanks, I know that. There are many ANSI character sets, and I did not bother checking which one this corresponds to, but it looks like Windows-1252.

> I was able to create a note with ö, save it without error, and view it in the web page.

I tried it again and it behaved as expected, so I must have mixed up the current version with your changes.

PR #94, opened 2018-08-25
Rapidjson update

SyrianBallaS:
This _should_ be ready. Tried ignoring non-rapidjson related stuff, if it had nothing to do with it was left alone for another PR since they will need to be changed eventually.

**Breakdown**:
- Used tag 1.1.0 for rapidjson submodule.
- Adjusted existing functions I made in `ra_utility`
- Got rid of the user-defined literals
- Updated .gitignore for common build artifacts and other common binaries to prevent the need to ever change it again.
- Got rid of `using namespace rapidjson;`
- Replaced `FileStream `with `I/OStreamWrapper`
- Scoped some ops for speed (sometimes when `unique_ptr `or `ostringstream `was used).

There will be a few review comments from myself as well.

PR #96, opened 2018-08-26
Enum to enum class and misc. (size/speed/type/conformance)

SyrianBallaS:
OK, combed over this like 5 times so it should be fine. Review this at your leisure. I don't really expect this be accepted until #94 is but thought people should at least see if it's acceptable. For the sake of not changing too much at once I went back and manually put stuff back. I do expect some merge conflicts (no conflicts currently) by the time this is looked at, I will fix them when the time comes.

This mainly has to do with `enum `to `enum class` conversion. Almost everything here is either directly or indirectly related to the conversion except for the changes in `RA_md5factory.cpp`, `ra_fwd.h`, and `ra_utility.h` (some was related) which I'll discuss in review comments since GitHub won't let me do them until this is actually a PR. The change in `RA_md5factory.cpp` was an attempt to understand dynamic memory ownership semantics, there will be a review comment in there explaining that.

**Breakdown**:
- All unit tests pass.
- Removed controversial stuff discussed before (mainly operator"").
   - An attempt to avoid future conflicts.
- Made all `enums `into `enum class`; a.k.a scoped `enum`.
- It's possible to compile with minimum standard conformance on, it is currently off because it bugs out intellisense. There will be review comments on this to show why so other people will know in the future why something there was incorrect.
- Overloaded every type of` bitwise-operator` for `enums `only to prevent bugs.
- Put more forward declares in `ra_fwd.h` to reduce the size of the binary some what.
- Made an "`Integral to Enum`" function, which is less strict than the `Enum To Integral `function such that the Integral type has to be implicity convertible to the `underlying type` of the `enum `instead of an exact match.
- Shortened the names of some `enums `since scope resolution is required now.
- Put some (left a few alone) `enums `outside of the the class where they were declared mainly because the scope resolution would be too verbose.
- Found some functions (at a glance) that could easily be made into a constant expression (guaranteed `O(1)` running time). Some of these were static, to avoid having to change too much at this time they were changed to `inline static constexpr `instead of just `inline constexpr`. I did ignore a lot of them if they didn't have to do with the enums.

Tried to ignore a lot of stuff to save them for later. There will be some review comments by myself but will not appear immediately.


src/RA_md5factory.cpp
```diff
@@ -1,66 +1,69 @@
...
+              "%02x%02x%02x%02x%02x%02x%02x%02x"
+              "%02x%02x%02x%02x%02x%02x%02x%02x",
+              digest[0], digest[1], digest[2], digest[3], digest[4], digest[5], digest[6], digest[7],
+              digest[8], digest[9], digest[10], digest[11], digest[12], digest[13], digest[14], digest[15]);
+
+    return buffer; // Implicit promotion to std::string
 }
 
-std::string RAGenerateMD5(const std::vector<BYTE> DataIn)
+std::string RAGenerateMD5(const std::vector<BYTE>& DataIn) noexcept
```
SyrianBallaS: this was being passed by value, very slow for structures with copy semantics. Wasn't sure if was desired to just move it into the function so it's still const.

src/RA_md5factory.h
```diff
@@ -2,10 +2,11 @@
...
 
-#include "RA_Defs.h"
+#include "ra_fwd.h"
+#include <vector>
 
-extern std::string RAGenerateMD5(const std::string& sStringToMD5);
-extern std::string RAGenerateMD5(const BYTE* pIn, size_t nLen);
-extern std::string RAGenerateMD5(const std::vector<BYTE> DataIn);
+_NODISCARD std::string RAGenerateMD5(_In_ const std::string& sStringToMD5) noexcept;
+_NODISCARD std::string RAGenerateMD5(_In_ const BYTE* const pIn, _In_ size_t nLen) noexcept;
```
SyrianBallaS: Not really sure in C but the placement of const in C++ does have a difference. This makes `pIn `a "const pointer to const", meaning that the pointer and the object are immutable.

src/ra_utility.h
```diff
@@ -180,85 +120,175 @@ tstrtoul(_In_z_ LPCTSTR _String,
...
+    using underlying_type = std::underlying_type_t<Enum>;
+    if constexpr (std::is_signed_v<underlying_type>)
+        return (itoe<Enum>(to_unsigned(etoi(a)) >> to_unsigned(b)));
+    else
+        return (itoe<Enum>((etoi(a)) >> b));
 }
 
-} // namespace rel_ops
+template<typename Enum, class = std::enable_if_t<std::is_enum_v<Enum>>> _CONSTANT_FN&
+operator>>=(_Inout_ Enum& a, _In_ const std::underlying_type_t<Enum> b) noexcept { return (a = a >> b); }
```
SyrianBallaS: These bitwise operators will work for any enum class, constraints were put to filter out invalid input.

src/services/ImageRepository.cpp
```diff
@@ -10,8 +10,8 @@
 namespace ra {
 namespace services {
 
-constexpr char* DefaultBadge = "00000";
-constexpr char* DefaultUserPic = "_User";
+inline constexpr const char* DefaultBadge = "00000";
+inline constexpr const char* DefaultUserPic = "_User";
```
SyrianBallaS: With conformance on, what was here was wrong.
Straight from Visual Studio: `info: Conversion from string literal loses const qualifier.`
If `constexpr `existed in C it would be correct, but there's a difference between `char*` (pointer to an array of characters) and` const char* ` (pointer to a null-terminated string) in C++. if you aren't sure what the value_type is just put auto.

tests/RA_CompVariable_Tests.cpp
```diff
@@ -58,97 +58,97 @@ TEST_CLASS(RA_CompVariable_Tests)
...
-        AssertParseCompVariable("0xU1234", Address, Nibble_Upper, 0x1234U);
-        AssertParseCompVariable("0xM1234", Address, Bit_0, 0x1234U);
-        AssertParseCompVariable("0xN1234", Address, Bit_1, 0x1234U);
-        AssertParseCompVariable("0xO1234", Address, Bit_2, 0x1234U);
-        AssertParseCompVariable("0xP1234", Address, Bit_3, 0x1234U);
-        AssertParseCompVariable("0xQ1234", Address, Bit_4, 0x1234U);
-        AssertParseCompVariable("0xR1234", Address, Bit_5, 0x1234U);
-        AssertParseCompVariable("0xS1234", Address, Bit_6, 0x1234U);
-        AssertParseCompVariable("0xT1234", Address, Bit_7, 0x1234U);
+        AssertParseCompVariable("0xH1234", ra::ComparisonVariableType::Address, ComparisonVariableSize::EightBit, 0x1234U);
```
SyrianBallaS: A big majority of the lines changed were just the enum scope resolution.

src/RA_httpthread.cpp
```diff
@@ -314,22 +309,20 @@ BOOL RAWeb::DoBlockingRequest(RequestType nType, const PostArgs& PostData, Docum
...
-        case RequestLogin:
-            return DoBlockingHttpPost("login_app.php", PostArgsToString(args), Response);
-        default:
-            return DoBlockingHttpPost("dorequest.php", PostArgsToString(args), Response);
-    }
+    PostArgs args = PostData; //  Take a copy
+    args['r'] = ra::RequestTypeToPost.at(ra::etoi(nType)); //  Embed request type
+    
+    if( nType == ra::RequestType::Login) 
+        return DoBlockingHttpPost("login_app.php", PostArgsToString(args), Response);
```
SyrianBallaS: Falling through caused problems, so it's an if else instead. there's a few other switches that were changed as well.

src/ra_fwd.h
```diff
@@ -2,34 +2,124 @@
...
+namespace ra {
+
+/* usings/typedefs matter in C++ */
+#if _MBCS
+using tstring = std::basic_string<char, std::char_traits<char>, std::allocator<char>>;
+#elif  _UNICODE
+using tstring = std::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t>>;
+#else
+#error Unknown Character set
+#endif /* _MBCS */
```
SyrianBallaS: Was attempting to make this file "include-less" but it was too much work and would a lot more lines added than already present so I just wanted to save it for later.

Jamiras:
I highly recommend figuring out some way to chunk this into smaller PRs - maybe only enum from one file at a time.

SyrianBallaS:
That's what I'm gonna do, chunk them.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -1523,6 +1552,7 @@ INT_PTR Dlg_AchievementEditor::AchievementEditorProc(HWND hDlg, UINT uMsg, WPARA
                 case CBN_SELENDCANCEL:
                 case CBN_SELCHANGE:
                 case CBN_CLOSEUP:
+                    _FALLTHROUGH;
```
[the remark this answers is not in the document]
SyrianBallaS: fallthrough is used before the after the last unhanded case label.

src/RA_Condition.h
```diff
@@ -15,54 +18,56 @@ enum ComparisonVariableSize
...
-enum ComparisonType
+enum class ComparisonType { Equals, LessThan, LessThanOrEqual, GreaterThan, GreaterThanOrEqual, NotEqualTo };
+enum class ConditionType { Standard, PauseIf, ResetIf, AddSource, SubSource, AddHits };
+
+namespace enum_sizes {
+
+_CONSTANT_VAR NUM_COMPARISON_VARIABLE_SIZETYPES{ 13U };
+_CONSTANT_VAR NUM_COMPARISON_VARIABLE_TYPES{ 4U };
+_CONSTANT_VAR NUM_COMPARISON_TYPES{ 6U };
+_CONSTANT_VAR NUM_CONDITION_TYPES{ 6U };
```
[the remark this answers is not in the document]
SyrianBallaS: While chunking these will be gone.

src/RA_Condition.h
```diff
@@ -15,54 +18,56 @@ enum ComparisonVariableSize
...
+    Address,         // compare to the value of a live address in RAM
+    ValueComparison, // a number. assume 32 bit 
+    DeltaMem,        // the value last known at this address.
+    DynamicVariable  // a custom user-set variable
 };
-extern const char* COMPARISONVARIABLETYPE_STR[];
 
-enum ComparisonType
+enum class ComparisonType { Equals, LessThan, LessThanOrEqual, GreaterThan, GreaterThanOrEqual, NotEqualTo };
+enum class ConditionType { Standard, PauseIf, ResetIf, AddSource, SubSource, AddHits };
```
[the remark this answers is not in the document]
SyrianBallaS: sure

src/RA_AchievementPopup.cpp
```diff
@@ -139,43 +146,52 @@ void AchievementPopup::Render(HDC hDC, RECT& rcDest)
...
-    SelectObject(hDC, hFontTitle);
-    TextOut(hDC, nTitleX, nTitleY, NativeStr(sTitle).c_str(), sTitle.length());
-    SIZE szTitle = { 0, 0 };
-    GetTextExtentPoint32(hDC, NativeStr(sTitle).c_str(), sTitle.length(), &szTitle);
+    auto lpszTitle{ std::make_unique<SIZE>() };
+    GetTextExtentPoint32(hDC, NativeStr(sTitle).c_str(), ra::strlen_as_int(sTitle), lpszTitle.get());
 
-    SIZE szAchievement = { 0, 0 };
-    if (ActiveMessage().Subtitle().length() > 0)
+    auto lpszAchievement{ std::make_unique<SIZE>() };
```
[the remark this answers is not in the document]
SyrianBallaS: Alright, with my findings it seems objects that need long lifetimes need to be allocated on the heap (like globals) and objects with very short lifetimes should be allocated on the stack. I reaffirmed that if something needs a large amount of memory allocated, it should be on the heap.

According the Microsoft (since this is for Windows currently) stack size is 1 KB in kernel and 16 KB in user mode. What should we define as "too large".

PR #97, opened 2018-08-29
Enum to enum class (BookmarkSubItems)

SyrianBallaS:
_*I couldn't reopen the other PR because I had to force push_

For the sake of this PR not going to waste I backed this up by branching out and reset this branch to when I did BookmarkSubItems. Subsequent conversions will be in different branches however but there's a high chance of merge conflicts so I'm just leaving this one and saving the rest locally until #94 (most "accurate") is merged. 

Breakdown:
- Required changes from previous feedback were applied, there might be some conflicts (not currently) and will be resolved when the time comes.
- `BookmarkSubItems `is a member `enum `type of `Dlg_MemBookmark`.
   - Renamed to `SubItems`
   - Constants in the **enum** were renamed without "CSI_" and changed to Pascal Case
- A conversion function from integral to enum type
   - The default is the `underlying_type `of the `enum `but will work if the integral is implicitly convertible to the `underlying_type`.
- Some C arrays (the namespaced ones) were converted to `std::array` to address a memory exploit (Spectre)
   - There's still a few but will get to them when the time comes.
   - Not all CPUs are spectre protected, such as mine (AMD)


SyrianBallaS:
I'll reopen this once #94 is merged. There will be conflicts.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -539,31 +536,38 @@ void Dlg_MemBookmark::SetupColumns(HWND hList)
 
     //	Remove all data.
     ListView_DeleteAllItems(hList);
+    {
+        auto lplvColumn{ std::make_unique<LV_COLUMN>() };
```
Jamiras: Why move this to the heap?

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -705,10 +709,10 @@ void Dlg_MemBookmark::ExportJSON()
 
     CComPtr<IFileSaveDialog> pDlg;
 
-    HRESULT hr;
+    auto hr{ S_OK };
```
Jamiras: Why change this? It was very clearly an HRESULT before, now anyone reading the code has to rely on an external tool (i.e. Visual Studio) to identify the type. Additionally, it's immediately assigned a value on the next line, so why waste time setting it here?

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -539,31 +536,38 @@ void Dlg_MemBookmark::SetupColumns(HWND hList)
...
-    bSuccess = ListView_EnableGroupView(hList, FALSE);
-
+    auto bSuccess = ListView_SetExtendedListViewStyle(hList, LVS_EX_FULLROWSELECT | LVS_EX_GRIDLINES | LVS_EX_DOUBLEBUFFER);
+#if _WIN32
+    bSuccess = ra::to_unsigned(ListView_EnableGroupView(hList, FALSE));
+#elif _WIN64
+    bSuccess = static_cast<DWORD>(ra::to_unsigned(ListView_EnableGroupView(hList, FALSE)));
+#else
+#error Unknown platform, Windows is currently the only supported platform!
+#endif // _WIN32
```
Jamiras: `bSuccess` isn't even used. Why go through all this trouble for it?

src/RA_Dlg_MemBookmark.h
```diff
@@ -42,8 +42,16 @@ class MemBookmark
...
-    //Dlg_MemBookmark();
-    //~Dlg_MemBookmark();
+    enum class SubItems
+    {
+        Desc,
+        Address,
+        Value,
+        Previous,
+        Changes
+    };
```
Jamiras: this enum isn't exposed by the class. recommend keeping it private.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: It used ZeroMemory before so I thought it needed an extendend lifetime. From now on everything will be stack based based. Using constructors instead of ZeroMemory.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: It has to be initilized something since it's a primitive.
Does HRESULT hr{}; work for you?

src/RA_Dlg_MemBookmark.h
SyrianBallaS: sure

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: I like being thorough, it seems you'd prefer to just check the return value instead. That'll be less trouble.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -536,38 +536,33 @@ void Dlg_MemBookmark::SetupColumns(HWND hList)
...
-            ListView_InsertColumn(hList, idx, lplvColumn.get());
-            idx++;
-        }
+            col.mask       = LVCF_TEXT | LVCF_WIDTH | LVCF_SUBITEM | LVCF_FMT,
+            col.fmt        = LVCFMT_CENTER | LVCFMT_FIXED_WIDTH,
+            col.cx         = ra::COLUMN_WIDTH.at(idx),
+            col.pszText    = tszText.data(),
+            col.cchTextMax = 255,
+            col.iSubItem   = idx
+        };
```
SyrianBallaS: Made it stack-based instead, the rest of the params aren't pointers so they get initilized to 0 automatically. Hopefully this looks clear.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -536,38 +536,33 @@ void Dlg_MemBookmark::SetupColumns(HWND hList)
...
-    auto bSuccess = ListView_SetExtendedListViewStyle(hList, LVS_EX_FULLROWSELECT | LVS_EX_GRIDLINES | LVS_EX_DOUBLEBUFFER);
-#if _WIN32
-    bSuccess = ra::to_unsigned(ListView_EnableGroupView(hList, FALSE));
-#elif _WIN64
-    bSuccess = static_cast<DWORD>(ra::to_unsigned(ListView_EnableGroupView(hList, FALSE)));
-#else
-#error Unknown platform, Windows is currently the only supported platform!
-#endif // _WIN32
+#if _WIN32_WINNT >= _WIN32_WINNT_LONGHORN
+    if (ListView_SetExtendedListViewStyle(hList, LVS_EX_FULLROWSELECT | LVS_EX_GRIDLINES | LVS_EX_DOUBLEBUFFER) != 0UL)
```
SyrianBallaS: This macro is Vista+ only when I went to go look it up. It didn't seem right to get rid of it. The other macro; `ListView_EnableGroupView`, was useless since group view is disabled by default (return of 0). Enabling it broke adding bookmarks, so I just got rid of it since this basically gets the same result.

src/RA_Dlg_MemBookmark.cpp
Jamiras: i was just commenting that the return value wasn't being used, so you didn't need seven lines of code to make sure it was set correctly.

src/RA_Dlg_MemBookmark.cpp
Jamiras: Yes, but it's given an actual value on the next line. There's no need to explicitly initialize it here.
```
HRESULT hr;
if (SUCCEEDED(hr = doSomething()))
{
```
I hate to suggest putting hr in the if statement, but you wouldn't initialize it there.
```
if (HRESULT hr; SUCCEEDED(hr = doSomething()))
{
```

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -833,17 +832,17 @@ std::wstring Dlg_MemBookmark::ImportDialog()
 
     CComPtr<IFileOpenDialog> pDlg;
 
-    HRESULT hr;
+    auto hr{ S_OK };
```
Jamiras: another abstracted HRESULT

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -540,30 +537,32 @@ void Dlg_MemBookmark::SetupColumns(HWND hList)
...
 
-        ListView_InsertColumn(hList, i, (LPARAM)&col);
+        ListView_InsertColumn(hList, idx, &col);
+        idx++;
     }
-
     m_nNumOccupiedRows = 0;
 
-    BOOL bSuccess = ListView_SetExtendedListViewStyle(hList, LVS_EX_FULLROWSELECT | LVS_EX_GRIDLINES | LVS_EX_DOUBLEBUFFER);
-    bSuccess = ListView_EnableGroupView(hList, FALSE);
```
Jamiras: is EnableGroupView not required?

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: Most of it will be gone later but I'll unabstract it for clarity when I can. I understand not everyone will know what s_ok is.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: I didn't see any different effect. It was false and it returned 0 meaning it was already disabled. Enabling it by putting true instead, I couldn't add/remove bookmarks.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: I got you. It makes sense unlike the switch statements.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: I got you. It did seem like a waste.

PR #103, opened 2018-09-06
Use Windows XP Compatible CommDlg APIs

SyrianBallaS:
So there's been recent discussions about stuff being broken for XP. I noticed some of the CommDlg (Common Dialog Library) Interfaces used were for Windows Vista and later. Not 100% sure if this is absolutely the best way, but I'm certain this works and is better than what was here before regarding requirements.

What this does is use CommDlg APIs that can be used with Windows XP (some of them even Windows 2000). The COM APIs not changed by this PR are already supported by XP.

Breakdown:
- `IFileOpenDialog/IFileSaveDialog `w/ `OPENFILENAME`
   - With `GetOpenFileName/GetSaveFileName`
- When choosing the ROM Directory in the game scanner it uses `OLE `w/ `BROWSEINFO`.
   - I chose to use OLE to prevent a deadlock (docs recommended it because it initializes OLE and COM)
   - Uses a callback to get it to work faster.
- Replaced the `_COMDLG_FILTERSPEC `(needs at least Vista) with a straight file filter
   - Files can be loaded and saved in `.json` format as well.

Other Stuff:
- `RA_Leaderboard `failed a runtime check during the unit tests in debug mode. I noticed it didn't have a `virtual destructor` despite having `virtual methods`. Making the destructor virtual got rid of the runtime error.
   - You either need a public virtual or protected non-virtual destructor if you want virtual methods but it seems unnecessary if it isn't going to be derived.
- Delibrately didn't change some stuff that will be by merging #94 .

Problems Noticed:
- The parser seems to broken when scanning game library files (those giant json files RACache/Data) in `RA_Dlg_GameLibrary `(Should be resolved after merging #94)
   - It either fails a debug assert or throws an exception (release mode)


src/RA_Dlg_MemBookmark.cpp
```diff
@@ -821,51 +810,57 @@ void Dlg_MemBookmark::ImportFromFile(std::string sFilename)
...
+    ofn.lpstrTitle   = _T("Import Bookmark File...");
+    ofn.nMaxFile     = BUF_SIZE - 1UL;
+    ofn.Flags        = OFN_ENABLESIZING | OFN_EXPLORER | OFN_PATHMUSTEXIST | OFN_FILEMUSTEXIST;
+    ofn.lpstrDefExt  = _T("txt");
+
+    ra::tstring sDefaultFilename;
+    sDefaultFilename.reserve(BUF_SIZE);
+    sDefaultFilename.append(ra::to_tstring(g_pCurrentGameData->GetGameID()));
+    sDefaultFilename.append(_T("-Bookmarks.txt"));
+    sDefaultFilename = sDefaultFilename.data();
```
SyrianBallaS: Building this string "normally" kept on causing heap corruption, so it's a bit non-standard but standard c++ complaint nonetheless.

.gitignore
```diff
@@ -3,7 +3,7 @@
 *.user
 /RetroAchievementsKey
 test_app/
-src/*_REMOTE_*.cpp
+packages/
```
SyrianBallaS: I don't know why this is shown as a change I didn't even touch this file. It might be from my master. I expect the whole thing to get (hopefully) finalized later.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -695,82 +700,66 @@ void Dlg_MemBookmark::ExportJSON()
...
+                   _T("Error!"), MB_OK);
         return;
     }
 
-    std::string defaultDir = g_sHomeDir + RA_DIR_BOOKMARKS;
-
-    CComPtr<IFileSaveDialog> pDlg;
-
-    HRESULT hr;
-    if (SUCCEEDED(hr = CoCreateInstance(CLSID_FileSaveDialog, nullptr, CLSCTX_ALL, IID_IFileSaveDialog, reinterpret_cast<void**>(&pDlg))))
```
Jamiras: Similar comment here. I suspect this wasn't working on XP before: https://github.com/RetroAchievements/RAIntegration/blob/610021460e6a1d4aa42f5892cd31240f7cde5e6d/src/RA_Dlg_MemBookmark.cpp#L710

src/RA_Defs.h
```diff
@@ -313,7 +286,13 @@ _NODISCARD std::string ByteAddressToString(_In_ ra::ByteAddress nAddr);
 
 } // namespace ra
 
-
+#if _MBCS
+_CONSTANT_VAR RA_MAX_PATH{ _MAX_PATH }; // multibyte max path
+#elif _UNICODE
+_CONSTANT_VAR RA_MAX_PATH{ 32'767 }; // Unicode max path
```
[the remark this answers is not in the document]
SyrianBallaS: It's a digit separator introduced in C++14.

src/RA_Defs.h
```diff
@@ -87,38 +75,23 @@
...
-//	RA-Only
-
-
 //	RA-Only
 #define RAPIDJSON_HAS_STDSTRING 1
-#pragma warning(push, 1)
-// This is not needed the most recent version
-#define _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS
-//	RA-Only
+#define RAPIDJSON_NOMEMBERITERATORCLASS 1
```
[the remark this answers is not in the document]
SyrianBallaS: This whole file will be put back to how it was except for the max path part.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: I think there was some buffer overrun, I'll put it back to ostringstream. The if block below should prevent a buffer overrun.

SyrianBallaS: I'm sorry dude it's still causing heap corruption when I want to use ostringstream, it's probably better the way I had it before.

SyrianBallaS:
Pretty much changed stuff to match feedback except for the `ostringstream `part. If someone can figure out a way to use it without causing heap corruption, be my guest.

src/RA_Defs.h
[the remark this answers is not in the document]
Jamiras: please remove it. its unnecessary, and not widely known.

src/RA_Dlg_MemBookmark.cpp
Jamiras: Please provide the `ostringstream` version you're having issues with. I can guess, but it would be best to know exactly.

Also, what is the purpose of line 851? `sDefaultFilename` should already be itself.

src/RA_Defs.h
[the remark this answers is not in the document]
SyrianBallaS: Sure.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: Alright, For line 851 that's required or the length/size will be 0.
The only thing I can think of is the extra capacity reserved by the string, I'm gonna see if adding 16 to the buffer size does anything.

SyrianBallaS: Figured it out, had to make ofn.lpstrFile from OPENFILENAME saved as a static c buffer. Now there's no heap corruption. It works with ostringstream now. There's extra capacity allocated just incase.

SyrianBallaS:
Need to install mfc on my laptop, reopening this later.

SyrianBallaS:
Should be safe to reopen, if there's anymore merge conflicts it won't be a big deal to fix them this time.

PR #106, opened 2018-09-15
Strict Conformance Mode

SyrianBallaS:
As a heads up for those reading this on Discord, there are only 16 additions and 4 deletions. I wanted to see if I could something as a "baby-step" instead of including these changes in a different PR.

**Summary**

- Basically, this makes our code compliant with ISO C++ Standard at the bare minimum (/permissive-). 
   - The reason for this is that some code was incorrect and might lead to undefined behavior in release mode, such as exceptions getting thrown randomly as well as an attempt to keep it bug-free in a non-OS specific context.
   - Usually access violations or invalidated references/handles
 - Further compliance will need to adhere to CppCoreGuidelines. 
    - We can use the built-in CppCoreCheck code analysis feature of Visual Studio
    - Not all of them will be run because they need the Guideline Support Library which is currently “3rd Party” (Microsoft GitHub Repo)
      - It’s used to help to satisfy the CppCoreGuidelines specification
   - They’ll just be run one at a time to prevent too many changes at once
- Only lines needed to be changed are changed, RapidJSON related violations were ignored (8 instances). Therefore, the /permissive- switch had to be turned back off or the code would not compile.

There will be some review comments why stuff was changed.


src/RA_Defs.h
```diff
@@ -53,7 +53,7 @@
 #define NODEFERWINDOWPOS
 #define NOMCX  
 
-
+struct IUnknown;
```
SyrianBallaS: The error appeared directly in `ObjBase.h` (a Windows header file).
`Error C2760 syntax error: unexpected token 'identifier', expected 'type specifier'`

Research online suggested to forward declare it before including a Windows Header in a  translation unit. It made sense since `IUnknown `was used before it was defined.

Files affected were `RA_Defs.h, RA_Dlg_Login.h, RA_Dlg_RomChecksum.h, and RA_Dlg_RichPresence.h`

src/RA_Dlg_Achievement.h
```diff
@@ -56,7 +56,7 @@ class Dlg_Achievements
     void RemoveAchievement(HWND hList, int nIter);
     size_t AddAchievement(HWND hList, const Achievement& Ach);
 
-    INT_PTR Dlg_Achievements::CommitAchievements(HWND hDlg);
+    INT_PTR CommitAchievements(HWND hDlg);
```
SyrianBallaS: `Error C4596 'CommitAchievements': illegal qualified name in member declaration`

It might have worked before because the compiler adjusted for it with an extention but putting the scoped name inside the same scope is wrong. It could have been mistaken as `Dlg_Achievements::Dlg_Achievements::CommitAchievements`.

src/services/ImageRepository.cpp
```diff
@@ -10,8 +10,8 @@
 namespace ra {
 namespace services {
 
-constexpr char* DefaultBadge = "00000";
-constexpr char* DefaultUserPic = "_User";
+inline constexpr const char* DefaultBadge{ "00000" };
```
SyrianBallaS: `Error C2440 'initializing': cannot convert from 'const char [6]' to 'char* const '`


I could have just used the macro I put a while back; `_CONSTANT_VAR ` ...
- Expands to` inline constexpr auto`, where `auto `is `const char* const`), 
- The `const `at the end usually doesn't show up in intellisense

..but felt it would be better do it this way to show people how to do it explicitly.
The inline is a performance boost (still correct) but the `const `part is what was needed because the left side was` char*` and the right side was `char* const` (const pointer)

src/RA_Dlg_Memory.cpp
```diff
@@ -838,7 +838,8 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
             {
                 Col.iOrder = i;
                 Col.iSubItem = i;
-                Col.pszText = _T("Search Result");
+                ra::tstring str{ _T("Search Result") };
```
SyrianBallaS:
```
Error C2440 '=': cannot convert from 'const char [14]' to 'LPSTR'			
         E0513 a value of type "const char *" cannot be assigned to an entity of type "LPSTR"
```
Basically the `rvalues `type does not match the `lvalues `type, this makes sure they do so the compiler knows we intend to drop the `const `qualifier.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -628,8 +628,8 @@ void Dlg_MemBookmark::WriteFrozenValue(const MemBookmark & Bookmark)
     if (!Bookmark.Frozen())
         return;
 
-    unsigned int addr;
-    unsigned int width;
+    unsigned int addr{};
```
SyrianBallaS: There were some warnings about potential uninitialized variables, so these make sure these are initialized to the default value.
value_types and pointers usually aren't automatically default constructible.

SyrianBallaS:
Somethings wrong, need to close this real quick.

src/RA_Dlg_Achievement.h
Jamiras: Most likely the result of copying the prototype from the cpp file and not stripping the class.

src/RA_Integration.vcxproj
```diff
@@ -235,6 +235,7 @@
       <DebugInformationFormat>EditAndContinue</DebugInformationFormat>
       <ControlFlowGuard>false</ControlFlowGuard>
       <ShowIncludes>false</ShowIncludes>
+      <ConformanceMode>true</ConformanceMode>
```
Jamiras: > Therefore,  the /permissive- switch had to be turned back off or the code would not compile.

This seems to contradict the comment in the description. I see `/permissive-` in the property pages, and it seems to compile.

SyrianBallaS: I should have edited the comment. After I merged the updated master I turned it back on, it was in the commit message but should've been changed in the comment as well.

PR #113, opened 2018-09-20
Enum to enum class (AchievementSetType)

SyrianBallaS:
This time I chose an "easy" one. At first I thought it would be trivial but there were a few minor things I had to get to work. Mostly making `AchievementSetType` a member type (more like a member type alias) and using the scoped names. 

There will be review comments below.
My commit history is starting to get too large in the PRs, I think I have to squash some commits in my master branch because i didn't merge master with this repo before. I'll do that before the next PR.


src/RA_Achievement.h
```diff
@@ -4,7 +4,18 @@
...
 
+enum class AchievementSetType
+{
+    Core,
+    Unofficial,
+    Local
+};
+
+} /* namespace detail */
+} /* namespace ra */
```
SyrianBallaS: This couldn't be a member type of AchievementSet because it made a circular dependency resulting in `AchievementSet  `and `Achievement `being incomplete types. I just aliased it instead. They are type aliases so we really shouldn't get any type clashes. Most of the time we used the AchievementSet::Type one like discussed before.

It's in a detail namespace to prevent it from being used directly as the member type was supposed be the one used.

src/RA_AchievementSet.cpp
```diff
@@ -17,36 +14,50 @@
...
-        case Local:
-            return g_sHomeDir + RA_DIR_DATA + std::to_wstring(nGameID) + L"-User.txt";
-        default:
-            return L"";
+        case Type::Core:
+            return str += L".txt";
+        case Type::Unofficial:
+            return str += L".txt";	// Same as Core
+        case Type::Local:
+            return str += L"-User.txt";
```
SyrianBallaS: Speed things up a bit.

src/RA_AchievementSet.h
```diff
@@ -4,6 +4,10 @@
 
 #include "RA_Achievement.h" // RA_Condition.h (RA_Defs.h)
 
+#if !defined(RAPIDJSON_DOCUMENT_H_) && defined(RA_UTEST)
+#define RAPIDJSON_NOMEMBERITERATORCLASS 1
+#include <rapidjson\include\rapidjson\document.h>
+#endif // !!defined(RAPIDJSON_DOCUMENT_H_) && defined(RA_UTEST)
```
SyrianBallaS: Unit test wouldn't compile, probably because of the `RA_EXPORTS `macro.

src/RA_AchievementSet.h
```diff
@@ -12,7 +16,9 @@
 class AchievementSet
 {
 public:
-    AchievementSet(AchievementSetType nType) :
+    using Type = ra::detail::AchievementSetType;
```
SyrianBallaS: This'll let us use `AchievementSet::Type`, defining it here caused problems.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -550,7 +550,7 @@ void Dlg_MemBookmark::SetupColumns(HWND hList)
             col.iSubItem   = idx
         };
 
-        if (idx == (ra::COLUMN_TITLE.size() - 1))
+        if (idx == (ra::to_signed(ra::COLUMN_TITLE.size()) - 1))
```
SyrianBallaS: Came up as warning. Level 4 warnings weren't enabled when I originally did this.

tests/RA_Achievement_Tests.cpp
```diff
@@ -1,6 +1,6 @@
 #include "CppUnitTest.h"
 
-#include "RA_Achievement.h"
+#include "RA_AchievementSet.h"
```
SyrianBallaS: It's a cpp file anyway. To keep it consisted with typing out `AchievementSet::Type`.

src/RA_Achievement.h
Jamiras: I believe you can eliminate the circular dependency by removing the `Achievement::IsCoreAchievement` method and removing the `SetActive(IsCoreAchievement())` call from `Achievement::Parse` and just call `SetActive(TRUE)` after calling `Parse` in `AchievementSet::LoadFromFile` here:

```
                    if (nFlags == 3 && m_nSetType == Core)
                    {
                        Achievement& newAch = AddAchievement();
                        newAch.Parse(AchievementsData[i]);
                        newAch.SetActive(TRUE); //Activate core by default
                    }
```

I'd rather clean up the problem than work around it.

src/RA_AchievementSet.cpp
Jamiras: might as well include this in the `oss` too:
```
    std::wstring str;
    {
        std::wostringstream oss;
        oss << g_sHomeDir << RA_DIR_DATA << nGameID;
        if (m_nSetType == Type::Local)
            oss << L"-User";
        oss << L".txt";
        str = oss.str();
    }
```

src/RA_AchievementSet.h
Jamiras: Preferable alternative is to exclude `OnRequestUnlocks` from the unit test compile:
```
#ifndef RA_UTEST
    static void OnRequestUnlocks(const Document& doc);
#endif
```
A similar change was made for `Achievement::Parse` in RA_Achievement.h

src/RA_Dlg_MemBookmark.cpp
Jamiras: I saw this too. My fix was to make `idx` unsigned: https://github.com/RetroAchievements/RAIntegration/pull/118/files#diff-ec82be412bc671652491c73da885f532R539

tests/RA_Achievement_Tests.cpp
Jamiras: This shouldn't be necessary if you eliminate `IsCoreAchievement`, which might also solve the issue that was requiring the `Document` support for unit tests.

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: I made it signed because of lv_columns  members being signed.

src/RA_Achievement.h
SyrianBallaS: So should we just call it "Achievement::Type" and/or alias it in achievementset?

src/RA_AchievementSet.cpp
SyrianBallaS: Haven't thought of that, thanks.

src/RA_AchievementSet.h
SyrianBallaS: Seems like a good approach to me.

tests/RA_Achievement_Tests.cpp
SyrianBallaS: Alright I'll try it out.

src/RA_Achievement.h
```diff
@@ -24,10 +22,33 @@ enum Achievement_DirtyFlags
...
+protected:
+#ifdef RA_UTEST
+    friend class ra::data::tests::RA_Achievement_Tests;
+#endif
+    enum class Type
+    {
+        Core,
+        Unofficial,
+        Local
+    };
```
SyrianBallaS: Had to make it a type member of achievement to satisfy other requirements. Made `protected `just in case we want to subclass it later and made only the classes that need it friends. All other classes use the `AchievementSet::Type` alias.

src/RA_Achievement.h
```diff
@@ -52,7 +73,8 @@ class Achievement
     inline BOOL GetPauseOnReset() const { return m_bPauseOnReset; }
     void SetPauseOnReset(BOOL bPause) { m_bPauseOnReset = bPause; }
 
-    BOOL IsCoreAchievement() const { return m_nSetType == Core; }
+    _Success_(return)
+        _NODISCARD _CONSTANT_FN IsCoreAchievement() const noexcept { return (m_nSetType == Type::Core); }
```
SyrianBallaS: Figured out a way to not need to get ride of this.

src/RA_AchievementSet.cpp
```diff
@@ -17,36 +14,44 @@
...
-        default:
-            return L"";
+        std::wostringstream oss;
+        oss << g_sHomeDir << RA_DIR_DATA << nGameID;
+        if (m_nSetType == Type::Local)
+            oss << L"-User";
+        oss << L".txt";
+        str = oss.str();
     }
+    return str;
```
SyrianBallaS: Gives it a stronger guarantee that an exception will not be thrown, got rid of the switch since it's already handled in the scope.

The `str = oss.str()` is a move assign and they aren't allowed to throw.

tests/RA_Achievement_Tests.cpp
```diff
@@ -14,15 +14,15 @@ TEST_CLASS(RA_Achievement_Tests)
 public:
     TEST_METHOD(TestStateEmpty)
     {
-        Achievement ach(AchievementSetType::Local);
+        Achievement ach{ Achievement::Type::Local };
```
SyrianBallaS: Since including `RA_AchievementSet.h `was considered unnecessary, I used the Achievement version instead. Had to make this `test class `a `friend `however since it needed access.

src/RA_Achievement.h
Jamiras: I'd like you to try to get rid of it. It's only used for activating the achievement in `Parse`, which overloads the functionality of `Parse` - `Parse` should _just_ parse the string. Activiation should happen outside of `Parse`, where we already know if it's a Core achievement or not.

SyrianBallaS: Accidently hit resolved here on my phone instead of reply. But sure it should be easy enough.

src/RA_Achievement.cpp
```diff
@@ -20,7 +20,8 @@ void _ReadStringTil(std::string& value, char nChar, const char*& pSource)
 
 //////////////////////////////////////////////////////////////////////////
 
-Achievement::Achievement(AchievementSetType nType) :
+_Use_decl_annotations_
+Achievement::Achievement(Type nType) noexcept :
```
Jamiras: With the removal of `IsCoreAchievement`, `m_nSetType` should no longer be used, eliminating the need to pass `nType` in the constructor, which should allow you to put the enum directly in `AchievementSet` and clean up some of the stuff you've been trying to work around.

PR #116, opened 2018-09-23
URL escape password before sending login request

ubergeek77:
Fixes #114

PR #117, opened 2018-09-23
URL-escape password before sending login request

### LINKED ISSUE #114: Passwords are not URL-escaped
I wasn't able to log in to RA through any of the emulator apps, so I changed my password to a simple one. I was able to log in with this simple password. After changing my password back to a more complex one, I wasn't able to log in again. This made me think it was an issue with not escaping special characters on the login request. I skimmed the source code, and saw that passwords are sent as-is without any checks to see if the characters are valid URL characters. Sure enough, after manually escaping my password and pasting the escaped version into RAIntegration's password field, I was able to log in.

For clarification, please consider the following example:

Let's say this string is my password (it's not my real password!): s3?AJhs:u+S=D<%-]@{O

If I use this string as-is in RAIntegration, my login will fail, even though I would be able to log in to RetroAchievements.org with this very password.

However, if I enter an _**escaped**_ version into RAIntegration, my login will succeed without issue. So my above password would become:

s3%3FAJhs%3Au%2BS%3DD%3C%25-%5D%40%7BO

So instead of sending this request:

`https://retroachievements.org/login_app.php?u=ubergeek77&p=s3?AJhs:u+S=D<%-]@{O`

it sends this request:

`https://retroachievements.org/login_app.php?u=ubergeek77&p=s3%3FAJhs%3Au%2BS%3DD%3C%25-%5D%40%7BO`

and the server can properly read the password and send the appropriate response.

RAIntegration should really be doing this manually, as sending _**any**_ URL request before properly sanitizing it is bad practice.

ubergeek77:
Fixes #114 

Basically just iterates through the password string and replaces it with a percent-encoded character if a non-valid character is found. Previously, if a password had any special characters outside of the range of valid URL characters, the login would just fail.

### DIFF
--- a/src/RA_Dlg_Login.cpp
+++ b/src/RA_Dlg_Login.cpp
@@ -5,6 +5,11 @@
 #include "RA_httpthread.h"
 #include "RA_PopupWindows.h"
 
+#include <cctype>
+#include <iomanip>
+#include <sstream>
+#include <string>
+
 //static 
 BOOL RA_Dlg_Login::DoModalLogin()
 {
@@ -46,9 +51,32 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
                         return TRUE;
                     }
 
+
+                    //URL-escape password so it gets sent to the server properly
+                    //https://stackoverflow.com/questions/154536/encode-decode-urls-in-c
+                    std::ostringstream escaped;
+                    escaped.fill('0');
+                    escaped << std::hex;
+
+                    const std::string &value = ra::Narrow(sPassEntry);
+                    for (std::string::const_iterator i = value.begin(), n = value.end(); i != n; ++i) {
+                        std::string::value_type c = (*i);
+
+                        // Keep alphanumeric and other accepted characters intact
+                        if (isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~') {
+                            escaped << c;
+                            continue;
+                        }
+
+                        // Any other characters are percent-encoded
+                        escaped << std::uppercase;
+                        escaped << '%' << std::setw(2) << int((unsigned char)c);
+                        escaped << std::nouppercase;
+                    }
+
                     PostArgs args;
                     args['u'] = ra::Narrow(sUserEntry);
-                    args['p'] = ra::Narrow(sPassEntry);		//	Plaintext password(!)
+                    args['p'] = escaped.str(); //Plaintext password(!)
 
                     rapidjson::Document doc;
                     if (RAWeb::DoBlockingRequest(RequestLogin, args, doc))


src/RA_Dlg_Login.cpp
```diff
@@ -46,9 +51,32 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
...
                     }
 
+
+                    //URL-escape password so it gets sent to the server properly
+                    //https://stackoverflow.com/questions/154536/encode-decode-urls-in-c
+                    std::ostringstream escaped;
+                    escaped.fill('0');
+                    escaped << std::hex;
+
+                    const std::string &value = ra::Narrow(sPassEntry);
```
SyrianBallaS: If you did this because you were afraid `sPassEntry `would be emptied don't be, it'll return a new string. I suggest getting rid `& `for speed `(O(1))`. Shallow copies are usually slow. Plus it gets hard to tell at a glance whether a reference has been invalidated with a static code analysis tool.

src/RA_Dlg_Login.cpp
```diff
@@ -46,9 +51,32 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
...
+
+                        // Keep alphanumeric and other accepted characters intact
+                        if (isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~') {
+                            escaped << c;
+                            continue;
+                        }
+
+                        // Any other characters are percent-encoded
+                        escaped << std::uppercase;
+                        escaped << '%' << std::setw(2) << int((unsigned char)c);
```
SyrianBallaS: I don't think boxing a type is required for `ostream `as it's insertion operator (`operator<<`) has overloads for all primitive types.

src/RA_Dlg_Login.cpp
Jamiras: `ra::Narrow` returns a `std::string` object created inside the method. Depending on how the compiler optimizes that, the variable may go out of scope before you're done with it. Recommend removing the reference: 
```
const std::string sValue = ra::Narrow(sPassEntry);
```
Also note the Hungarian prefix. Please follow the coding styles outlined in the [CONTRIBUTING.md](https://github.com/RetroAchievements/RAIntegration/blob/master/CONTRIBUTING.md) file

src/RA_Dlg_Login.cpp
```diff
@@ -46,9 +51,32 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
...
 
+
+                    //URL-escape password so it gets sent to the server properly
+                    //https://stackoverflow.com/questions/154536/encode-decode-urls-in-c
+                    std::ostringstream escaped;
+                    escaped.fill('0');
+                    escaped << std::hex;
+
+                    const std::string &value = ra::Narrow(sPassEntry);
+                    for (std::string::const_iterator i = value.begin(), n = value.end(); i != n; ++i) {
```
Jamiras: coding style requests braces on separate line.

ubergeek77:
I can attempt to make the necessary changes to RA_Dlg_Login.cpp to address the minor issues in this pull, but after my next attempt, and any requirements beyond that (specifically, handling it inside RAWeb::PostArgsToString as @Jamiras has noted), I don't see the value of me, personally, spending any more time with this. I could just keep submitting changes to this pull, and have the contributors here spend time reviewing it and finding things they don't like, but that time would be better spent by a contributor to implement a solution they are satisfied with. I am relatively unskilled, so me dragging this pull on and on would just waste my time, and your much more valuable time. Pushing back the timeframe of a successfully committed fix for this issue is the complete opposite of what my goal in opening this pull was.

Based on the relative simplicity on transforming a string, and the verbiage used by the two of you that demonstrates you are both highly knowledgeable, I am led to believe this would be a very quick fix for you to implement (at the risk of underestimating your backlog and busy schedules; I do not mean to offend or belittle here).

I won't personally close this pull, but I did what I wanted to do, which was bringing awareness to a problem that has been apparently completely ignored for [at least 3 years](https://github.com/RetroAchievements/RAIntegration/commit/eb0002f6f8ccd62cce937b28a3d436966b9fa9d3#diff-e403b511ee4aa10aab8c2b2650abb1fdR102), but quite possibly for [over 4 years, when the project saw its first commit](https://github.com/RetroAchievements/RAIntegration/commit/1990f74e557a9e1047e77f9767920ab582584f9d#diff-bd1f0353fd6394f164c5503d2ee34717R50). 

[There was a comment made that it had only been "three days" since I filed the issue](https://github.com/RetroAchievements/RAIntegration/pull/116#issuecomment-423817427), suggesting I was being too jumpy, but this has been an issue for quite a lot longer than 3 days, and has apparently not been acknowledged as a bug in all that time.

I am glad to see such strict guidelines nowadays, as it makes me look forward to the future of the project. Best of luck.

meleu:
@ubergeek77 hey dude, although the RetroAchievments project exist since 2013, in the early days the emulators, the integration and the site (database stuff, PHP, HTML, CSS, JavaScript, etc.) were developed by a single person: @ScottFromDerby. It's crazy, isn't it? All that myriad of things being done by a single person... As you can imagine the evolution occurred slowly. Also, after sometime I believe Scott became busy with RL stuff and the whole project went on a hiatus.

Only in the late 2017 things started to become more actively developed and with other skilled contributors joining the dev team.

The two guys who reviewed your code are the most active in this repository, but both are here working/improving "someone else's code".

IMO you didn't waste anybody's time. You are helping the current active developers to be aware of an issue/bug.

It would be cool if you could try to solve it in [PostArgsToString](https://github.com/RetroAchievements/RAIntegration/blob/542b5f8389ab49e988929b6a94c257e26d45a765/src/RA_httpthread.cpp#L973) (following the our coding style), as suggested by @Jamiras. But if you don't want, then we have to wait until someone does.

There's no need to rush in solving this in the next few days, because "Even if we did merge this today, it would probably by 3-4 months before it's actually published to the site" (as mentioned by @Jamiras ).

I hope you don't feel discouraged and keep your interest in being a contributor/entusiast of RetroAchievements.

Jamiras:
Closing as resolved by this change: https://github.com/RetroAchievements/RAIntegration/blob/master/src/RA_httpthread.cpp#L549

PR #109, opened 2018-09-17
Demacrofied where appropriate

SyrianBallaS:
I was looking back at earlier PRs and felt like the demacrofication felt proper now. The other PR was off-track and felt making a new one would be more efficient.

The other one had `715 additions and 357 deletions`. 
This one has `429 additions and 282 deletions`, which most are just prefixing `ra` everywhere (RA_AchievementOverlay.cpp is one of them).

Some other stuff changed as well, stuff around it that should were unchanged.

A test to see if a pointer satisfied the `NullablePointer` named requirement made for `ra::SafeDelete`; it was in the form of a `type_trait `(data contracts may happen in C++20, but it is a good learning experience).

There will be review comments explaining the above more specifically and other things that were too important to ignore.

PR #118, opened 2018-09-25
add interface/service for reading/writing text files

Jamiras:
* added abstract `TextReader`/`TextWriter` classes. Provided implementations for reading/writing files (runtime) and reading/writing strings (unit tests).
* moved string functions (`Widen`/`Narrow`/`NativeStr`) from `RA_Defs` to new `RA_StringUtils` file
* added virtual destructors to interface classes

### DIFF
--- a/src/RA_Defs.h
+++ b/src/RA_Defs.h
@@ -232,33 +232,10 @@ const int SERVER_PING_DURATION = 2 * 60;
 #define UNUSED( x ) ( x );
 #endif
 
+#include "RA_StringUtils.h"
+
 namespace ra {
-
-_NODISCARD std::string Narrow(_In_ const std::wstring& wstr);
-_NODISCARD std::string Narrow(_Inout_ std::wstring&& wstr) noexcept;
-_NODISCARD std::string Narrow(_In_z_ const wchar_t* wstr);
-_NODISCARD std::wstring Widen(_In_ const std::string& str);
-_NODISCARD std::wstring Widen(_Inout_ std::string&& str) noexcept;
-_NODISCARD std::wstring Widen(_In_z_ const char* str);
-
-//	No-ops to help convert:
-_NODISCARD std::wstring Widen(_In_z_ const wchar_t* wstr);
-_NODISCARD std::wstring Widen(_In_ const std::wstring& wstr);
-_NODISCARD std::string Narrow(_In_z_ const char* str);
-_NODISCARD std::string Narrow(_In_ const std::string& wstr);
 _NODISCARD std::string ByteAddressToString(_In_ ra::ByteAddress nAddr);
-
-} // namespace ra
-
-
-
-
-#ifdef UNICODE
-#define NativeStr(x) ra::Widen(x)
-#define NativeStrType std::wstring
-#else
-#define NativeStr(x) ra::Narrow(x)
-#define NativeStrType std::string
-#endif
+}
 
 #endif // !RA_DEFS_H
--- a/src/RA_Json.h
+++ b/src/RA_Json.h
@@ -17,6 +17,6 @@
 void _WriteBufferToFile(const std::wstring& sFileName, const rapidjson::Document& doc);
 
 bool LoadDocument(_Out_ rapidjson::Document& doc, ra::services::TextReader& reader);
-bool SaveDocument(_In_ rapidjson::Document& doc, ra::services::TextWriter& writer);
+bool SaveDocument(_In_ const rapidjson::Document& doc, ra::services::TextWriter& writer);
 
 #endif // !RA_JSON_H
--- a/src/RA_Json.h
+++ b/src/RA_Json.h
@@ -2,6 +2,9 @@
 #define RA_JSON_H
 #pragma once
 
+#include "services\TextReader.hh"
+#include "services\TextWriter.hh"
+
 #define RAPIDJSON_HAS_STDSTRING 1
 #define RAPIDJSON_NOMEMBERITERATORCLASS 1
 
@@ -11,6 +14,9 @@
 #include <rapidjson\ostreamwrapper.h>
 #include <rapidjson\error\en.h>
 
-extern void _WriteBufferToFile(const std::wstring& sFileName, const rapidjson::Document& doc);
+void _WriteBufferToFile(const std::wstring& sFileName, const rapidjson::Document& doc);
+
+bool LoadDocument(_Out_ rapidjson::Document& doc, ra::services::TextReader& reader);
+bool SaveDocument(_In_ rapidjson::Document& doc, ra::services::TextWriter& writer);
 
 #endif // !RA_JSON_H
--- a/src/ra_utility.h
+++ b/src/ra_utility.h
@@ -103,8 +103,8 @@ filesize(std::basic_string<CharT>&& filename) noexcept
     return file.tellg();
 } // end function filesize
 
-typedef const TCHAR *LPCTSTR;
-typedef TCHAR *LPTSTR;
+using LPCTSTR = const TCHAR*;
+using LPTSTR = TCHAR*;
 
 // More functions to be Unicode compatible w/o sacrificing MBCS
 _EXTERN_C
--- a/src/ra_utility.h
+++ b/src/ra_utility.h
@@ -92,6 +92,9 @@ filesize(std::basic_string<CharT>&& filename) noexcept
     return file.tellg();
 } // end function filesize
 
+typedef const TCHAR *LPCTSTR;
+typedef TCHAR *LPTSTR;
+
 // More functions to be Unicode compatible w/o sacrificing MBCS
 _EXTERN_C
 /* A wrapper for converting a string to an unsigned long depending on the character set specified */
--- a/src/services/IConfiguration.hh
+++ b/src/services/IConfiguration.hh
@@ -23,7 +23,7 @@ enum class Feature
 class IConfiguration
 {
 public:
-    virtual ~IConfiguration() {}
+    virtual ~IConfiguration() noexcept = default;
 
     virtual const std::string& GetUsername() const = 0;
     virtual void SetUsername(const std::string& sValue) = 0;
--- a/src/services/IConfiguration.hh
+++ b/src/services/IConfiguration.hh
@@ -20,8 +20,11 @@ enum class Feature
     PreferDecimal,
 };
 
-class IConfiguration {
+class IConfiguration
+{
 public:
+    virtual ~IConfiguration() {}
+
     virtual const std::string& GetUsername() const = 0;
     virtual void SetUsername(const std::string& sValue) = 0;
     virtual const std::string& GetApiToken() const = 0;
--- a/src/services/ILeaderboardManager.hh
+++ b/src/services/ILeaderboardManager.hh
@@ -10,7 +10,7 @@ namespace services {
 class ILeaderboardManager
 {
 public:
-    virtual ~ILeaderboardManager() {}
+    virtual ~ILeaderboardManager() noexcept = default;
 
     virtual void Test() = 0;
     virtual void Reset() = 0;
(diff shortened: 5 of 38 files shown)


src/RA_Json.h
```diff
@@ -11,6 +14,9 @@
 #include <rapidjson\ostreamwrapper.h>
 #include <rapidjson\error\en.h>
 
-extern void _WriteBufferToFile(const std::wstring& sFileName, const rapidjson::Document& doc);
+void _WriteBufferToFile(const std::wstring& sFileName, const rapidjson::Document& doc);
+
+bool LoadDocument(_Out_ rapidjson::Document& doc, ra::services::TextReader& reader);
+bool SaveDocument(_In_ rapidjson::Document& doc, ra::services::TextWriter& writer);
```
SyrianBallaS: If doc is an "in" parameter it should be marked with `const `since it's a reference.

src/services/IConfiguration.hh
```diff
@@ -20,8 +20,11 @@ enum class Feature
     PreferDecimal,
 };
 
-class IConfiguration {
+class IConfiguration
+{
 public:
+    virtual ~IConfiguration() {}
```
SyrianBallaS: this should be marked with `noexcept `and a `default`. Like this...

`virtual ~IConfiguration() noexcept = default;`

There's other stuff you should do too but don't worry about it, I was thinking of putting in a PR with proper/explicit constructors later.

src/ra_utility.h
```diff
@@ -103,6 +103,9 @@ filesize(std::basic_string<CharT>&& filename) noexcept
     return file.tellg();
 } // end function filesize
 
+typedef const TCHAR *LPCTSTR;
+typedef TCHAR *LPTSTR;
+
```
SyrianBallaS: Not a big deal but these should be `usings `just in case the compiler complains that the type signature doesn't match.

src/services/Initialization.cpp
```diff
@@ -3,13 +3,17 @@
...
+#include "services\impl\WindowsFileSystem.hh"
 
 namespace ra {
 namespace services {
 
-void Initialization::RegisterServices(const std::wstring& sHomeDir, const std::string& sClientName)
+void Initialization::RegisterServices(const std::string& sClientName)
 {
-    std::wstring sFilename = sHomeDir + L"RAPrefs_" + ra::Widen(sClientName) + L".cfg";
+    auto* pFileSystem = new ra::services::impl::WindowsFileSystem();
```
SyrianBallaS: You don't have to put `* `for pointers, but if that's what you prefer go ahead. Anyway, are these pointers being managed somewhere?

src/services/TextWriter.hh
```diff
@@ -0,0 +1,31 @@
...
+class TextWriter
+{
+public:
+    virtual ~TextWriter() {}
+
+    /// <summary>
+    /// Writes text to the output.
+    /// </summary>
+    /// <param name="sText">The string to write.</param>
+    virtual void Write(_In_ std::string& sText) = 0;
```
SyrianBallaS: In parameters that are references need to be `const `unless you mean `_Inout_ `(param is in a valid state before and after the function but can change).

src/services/impl/FileTextReader.hh
```diff
@@ -0,0 +1,52 @@
...
+#include <fstream>
+
+namespace ra {
+namespace services {
+namespace impl {
+
+class FileTextReader : public ra::services::TextReader
+{
+public:
+    FileTextReader(const std::wstring& sFilename)
```
SyrianBallaS: Default constructors need to marked with `noexcept`. Default constructors with parameters need to be marked with explicit to prevent `implicit `conversion.

src/services/impl/FileTextReader.hh
```diff
@@ -0,0 +1,52 @@
...
+        return true;
+    }
+
+    bool GetLine(_Out_ std::wstring& sLine) override
+    {
+        std::string sNarrowLine;
+        if (!std::getline(m_iStream, sNarrowLine))
+            return false;
+
+        sLine = ra::Widen(ra::TrimLineEnding(sNarrowLine));
```
SyrianBallaS: Not a big deal but you don't have to use the fully qualified name inside the same namespace but if that's what you prefer that's OK.

src/services/impl/FileTextReader.hh
```diff
@@ -0,0 +1,52 @@
...
+        return true;
+    }
+
+    std::ifstream& GetFStream() { return m_iStream; }
```
SyrianBallaS: Is this meant to be modified? If not, mark it `const `and `noexcept`.

src/services/impl/WindowsFileSystem.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return static_cast<bool>(CreateDirectoryW(sDirectory.c_str(), nullptr));
+}
+
+std::unique_ptr<TextReader> WindowsFileSystem::OpenTextFile(const std::wstring& sPath) const
+{
+    auto pReader = std::make_unique<FileTextReader>(sPath);
+    if (!pReader->GetFStream().is_open())
+        return std::unique_ptr<TextReader>();
+
+    return std::unique_ptr<TextReader>(pReader.release());
```
SyrianBallaS: I'm not really sure if this OK or not. I'd recommend using `std::swap` with a new `std::unique_ptr` or moving `pReader `into the new one. Something like this...
```
auto pRet = std::move(pReader);
return pRet;
```
If you're sure it doesn't leak don't worry about it.

src/services/impl/WindowsFileSystem.hh
```diff
@@ -0,0 +1,31 @@
...
+#include "services\IFileSystem.hh"
+
+namespace ra {
+namespace services {
+namespace impl {
+
+class WindowsFileSystem : public IFileSystem
+{
+public:
+    WindowsFileSystem();
```
SyrianBallaS: Default constructors have to be `noexcept`. Sure most of the ones in the code base aren't, but that'll change soon.

tests/mocks/MockFileSystem.hh
```diff
@@ -0,0 +1,84 @@
...
+        if (pIter == m_mFileContents.end())
+            return std::unique_ptr<TextReader>();
+
+        auto pReader = std::make_unique<ra::services::impl::StringTextReader>(pIter->second);
+        return std::unique_ptr<TextReader>(pReader.release());
+    }
+
+    std::unique_ptr<TextWriter> CreateTextFile(const std::wstring& sPath) const override
+    {
+        std::string& sContents = m_mFileContents[sPath];
```
SyrianBallaS: Assignment with array indexing is a possible security issue. It may cause a CPU to be vulnerable to Spectre. It won't show up as a warning unless /Wall is specified but it's  too much of a hassle to turn it on without a precompiled header to disable all the warnings for library files.

src/RA_Defs.h
```diff
@@ -232,33 +232,10 @@ const int SERVER_PING_DURATION = 2 * 60;
...
-
-
-
-#ifdef UNICODE
-#define NativeStr(x) ra::Widen(x)
-#define NativeStrType std::wstring
-#else
-#define NativeStr(x) ra::Narrow(x)
-#define NativeStrType std::string
-#endif
```
SyrianBallaS: This PR is probably more beneficial than #109. I'll just merge in the new master once this PR is merged while putting the demacrofied `ra::NativeStr `function templates where they should be.

src/RA_Json.h
Jamiras: will do

src/ra_utility.h
Jamiras: these were copied line-for-line from the windows header they were defined in so I could include this file without having to include the windows header. will try with the using.

src/services/IConfiguration.hh
Jamiras: will do. this is mostly out of habit - you'll probably have to remind me again

src/services/Initialization.cpp
Jamiras: `ServiceLocator::Service` has a `unique_ptr` for managing the instance.

Jamiras: the construct is typically written as:
```
ServiceLocator::Provide<Interface>(new Implementation);
```
the allocation is more obvious when on a separate line.

src/services/TextWriter.hh
Jamiras: i agree. missed it after copy/pasting TextReader.

src/services/impl/FileTextReader.hh
Jamiras: ok

src/services/impl/FileTextReader.hh
Jamiras: you cannot read from a const stream. the position gets updated.

in this particular case, it's exposed for consumption by the rapidjson helper function `LoadDocument`. it's not actually part of the interface.

src/services/impl/WindowsFileSystem.cpp
Jamiras: this was required to change the pointer type. the first pointer is the subclass, and a pointer to the base class is returned. `release` detaches the pointer, so ownership can be reassigned to the other `unique_ptr`.

src/services/impl/WindowsFileSystem.hh
Jamiras: will do

tests/mocks/MockFileSystem.hh
Jamiras: using the indexer allows for inserting or replacing without having to call find first. what's your proposed alternative?

tests/mocks/MockFileSystem.hh
```diff
@@ -0,0 +1,84 @@
...
+#include <map>
+
+namespace ra {
+namespace services {
+namespace mocks {
+
+class MockFileSystem : public IFileSystem
+{
+public:
+    MockFileSystem() : m_Override(this)
```
SyrianBallaS: You should mark this noexcept

src/services/impl/StringTextWriter.hh
```diff
@@ -0,0 +1,47 @@
...
+#include "RA_StringUtils.h"
+
+namespace ra {
+namespace services {
+namespace impl {
+
+class StringTextWriter : public ra::services::TextWriter
+{
+public:
+    explicit StringTextWriter() noexcept
```
SyrianBallaS: You should probably get rid of the  explicit here. It's only useful with default constructors with one or more parameters. Since most people here like assignment instead of initializers itll be too verbose as it is.

src/services/ILeaderboardManager.hh
```diff
@@ -10,6 +10,8 @@ namespace services {
 class ILeaderboardManager
 {
 public:
+    virtual ~ILeaderboardManager() {}
+
```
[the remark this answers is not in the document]
Jamiras: fixed

src/services/impl/StringTextWriter.hh
Jamiras: fixed

tests/mocks/MockFileSystem.hh
Jamiras: done

tests/mocks/MockFileSystem.hh
Jamiras: `at` doesn't work. from the documentation:

>  If no such element exists, an exception of type std::out_of_range is thrown.

the indexer will create the element if it doesn't exist.

SyrianBallaS: Alright dont worry about it. I'll do more research about it and address it later.

SyrianBallaS: If you dont care about order use std::unordered_map. Its find and at methods are constant on average but linear in the worst case. Std::map versions are logrithimic.

Jamiras: thanks for the reminder. in this particular case, it's a mock for unit tests, so the collection shouldn't contain more than a handful of items. I don't expect a noticeable difference using either.

PR #120, opened 2018-09-30
MessageBoxViewModel (beginnings of ViewModel framework)

Jamiras:
* Provide system-independent ViewModel for showing MessageBoxes
* Illustrate Win32-specific implementation
* Allows for mocking MessageBoxes in unit tests, though not implemented/illustrated in this PR
* Provide helper functions for common dialogs (message, warning, error)
* Convert messageboxes in RA_Core to use ViewModel - many more still exist
* Support for TaskDialog (if available), fallback to multiline message (WinXP) - hidden in implementation
image
image


src/RA_Core.cpp
```diff
@@ -905,6 +916,7 @@ static void RA_CheckForUpdate()
     std::string Response;
     if (RAWeb::DoBlockingRequest(RequestLatestClientPage, args, Response))
     {
+        // TODO: this doesn't work - response is JSON
```
Jamiras: this code just isn't working right now. "Check for Updates" menu option. it's expecting a single string instead of a JSON package

src/services/ServiceLocator.hh
```diff
@@ -130,6 +134,11 @@ private:
 
             RA_LOG("ERROR: %s\n", sMessage.c_str());
 
+#ifndef NDEBUG
+            // expanded definition of assert macro so we can use the constructed error message
+            _wassert(ra::Widen(sMessage).c_str(), _CRT_WIDE(__FILE__), (unsigned)(__LINE__));
+#endif
```
Jamiras: pops the error message up as an assert before throwing it when in a debug build

src/ui/ModelProperty.cpp
```diff
@@ -0,0 +1,41 @@
+#include "ModelProperty.hh"
```
Jamiras: ModelProperties are largely based on WPF DependencyProperties. A property has a unique identifier and a default value. A map in the ViewModel hashes the unique identifier to the current value of the property. If the value is not in the hash, the default value is returned.

The primary reason for using hashes of values is for binding the ViewModel to the UI for automatic synchronizing (not needed for MessageBoxes), but this PR sets things up for that. I plan to do the Rich Presence monitor next to illustrate simple binding before moving onto the Login dialog to illustrate control binding.

src/ui/ModelProperty.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <algorithm>
+
+namespace ra {
+namespace ui {
+
+int ModelPropertyBase::s_nNextKey = 1;
+std::vector<ModelPropertyBase*> ModelPropertyBase::s_vProperties;
```
Jamiras: Since the hash only contains the key, this provides a way to get from the key back to the property. Property instances are static singletons, this only holds a reference.

src/ui/ViewModelBase.cpp
```diff
@@ -0,0 +1,142 @@
...
+    else
+    {
+        if (iter != m_mIntValues.end())
+            return;
+
+        m_mIntValues[pProperty.GetKey()] = static_cast<int>(bValue);
+    }
+
+#ifdef _DEBUG
+    m_mDebugValues[pProperty.GetPropertyName()] = bValue ? L"true" : L"false";
```
Jamiras: The raw hash maps the unique identifier to the value. The debug hash maps the property name to a string representation of the value for viewing in the debugger - it should never be read in the code.

src/ui/ViewModelBase.cpp
```diff
@@ -0,0 +1,142 @@
...
+#endif
+
+    if (!m_vNotifyTargets.empty())
+    {
+        BoolModelProperty::ChangeArgs args{ pProperty, !bValue, bValue };
+
+        // create a copy of the list of pointers in case it's modified by one of the callbacks
+        NotifyTargetSet vNotifyTargets(m_vNotifyTargets);
+        for (NotifyTarget* target : vNotifyTargets)
+            target->OnViewModelBoolValueChanged(args);
```
Jamiras: This is how bindings will be notified when a property changes (and _only_ if it changes). Only used by unit tests in this PR.

src/ui/ViewModelBase.hh
```diff
@@ -0,0 +1,101 @@
...
+#pragma once
+
+#include "ModelProperty.hh"
+
+#include <string>
+
+namespace ra {
+namespace ui {
+
+class ViewModelBase
```
Jamiras: `ViewModelBase` serves as a `ModelProperty` bucket. It has no properties of its own.

src/ui/WindowViewModelBase.hh
```diff
@@ -17,7 +19,65 @@ struct Size
...
+
+    /// <summary>
+    /// Gets the window title.
+    /// </summary>
+    const std::wstring& GetWindowTitle() const { return GetValue(WindowTitleProperty); }
+
+    /// <summary>
+    /// Sets the window title.
+    /// </summary>
+    void SetWindowTitle(const std::wstring& sTitle) { SetValue(WindowTitleProperty, sTitle); }
```
Jamiras: Three lines per property - the first declares the ModelProperty (implementation in cpp file). The second and third provide ways to read/write the value - these call into `ViewModelBase` for managing the buckets. To make a property read-only, the setter method can be made protected or private.

src/ui/viewmodels/MessageBoxViewModel.hh
```diff
@@ -0,0 +1,167 @@
...
+    RetryCancel,
+};
+
+enum class MessageBoxIcon
+{
+    None,
+    Info,
+    Warning,
+    Error,
+};
```
Jamiras: I couldn't decide if these should be members of the `MessageBoxViewModel` (i.e. `MessageBoxViewModel::Icon` instead of `MessageBoxIcon`). I'm probably going to move them in, but wanted a second opinion.

src/ui/viewmodels/MessageBoxViewModel.hh
```diff
@@ -0,0 +1,167 @@
...
+{
+public:
+    MessageBoxViewModel() noexcept {}
+
+    MessageBoxViewModel(const std::wstring& sMessage) noexcept
+    {
+        SetMessage(sMessage);
+    }
+
+#pragma push_macro("GetMessage") // windows stole my method name
```
Jamiras: once we fully decouple the UI from the ViewModels, the conflict should no longer exist.

src/ui/win32/Desktop.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <assert.h>
+
+namespace ra {
+namespace ui {
+namespace win32 {
+
+Desktop::Desktop() noexcept
+{
+    m_vDialogControllers.push_back(new Dlg_MessageBox::Controller());
```
Jamiras: A controller maps a ViewModel to an implementation-specific window. When `Show` or `ShowWindow` is called, the appropriate controller is called to generate the UI.

src/ui/win32/Dlg_MessageBox.cpp
```diff
@@ -0,0 +1,121 @@
...
+        {
+            default:
+            case MessageBoxButtons::OK: uType |= MB_OK; break;
+            case MessageBoxButtons::OKCancel: uType |= MB_OKCANCEL; break;
+            case MessageBoxButtons::YesNo: uType |= MB_YESNO; break;
+            case MessageBoxButtons::YesNoCancel: uType |= MB_YESNOCANCEL; break;
+            case MessageBoxButtons::RetryCancel: uType |= MB_RETRYCANCEL; break;
+        }
+
+        nButton = MessageBoxW(GetActiveWindow(), pMessage->c_str(), oMessageBoxViewModel.GetWindowTitle().c_str(), uType);
```
Jamiras: here's the actual Win32 API call. there's another one on line 101 for TaskDialog

tests/VM_MessageBoxViewModel_Tests.cpp
```diff
@@ -0,0 +1,235 @@
...
+    TEST_METHOD(TestShowMessage)
+    {
+        MockDesktop mockDesktop([](MessageBoxViewModel& vmMessageBox)
+        {
+            Assert::AreEqual(std::wstring(L"Hello, world!"), vmMessageBox.GetMessage());
+            Assert::AreEqual(std::wstring(), vmMessageBox.GetHeader());
+            Assert::AreEqual(MessageBoxIcon::None, vmMessageBox.GetIcon());
+            Assert::AreEqual(MessageBoxButtons::OK, vmMessageBox.GetButtons());
+            return DialogResult::OK;
+        });
```
Jamiras: This illustrates how to use a MockDesktop helper class to intercept the request to display UI associated to a ViewModel and validate properties of the ViewModel. In the case of intercepting ShowModal, it returns the DialogResult.

tests/VM_MessageBoxViewModel_Tests.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+        MessageBoxViewModel::ShowMessage(L"Hello, world!");
```
Jamiras: The calls through the above in the MockDesktop via the Service override.

tests/VM_MessageBoxViewModel_Tests.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+        mockDesktop.AssertDialogShown();
```
Jamiras: The asserts in the callback don't provide any value if the callback isn't actually called. This just asserts that they were processed.

SyrianBallaS:
This seems interesting. I'll review it once I'm at home. I've used WPF for desktop apps before so I have a basic idea how this should work.

src/ui/ViewModelBase.hh
```diff
@@ -0,0 +1,101 @@
...
+    StringModelProperty::ValueMap m_mStringValues;
+    IntModelProperty::ValueMap m_mIntValues;
+
+#ifdef _DEBUG
+    // complete list of values as strings for viewing in the debugger
+    std::map<std::string, std::wstring> m_mDebugValues;
+#endif
+
+    typedef std::set<NotifyTarget*> NotifyTargetSet;
+    NotifyTargetSet m_vNotifyTargets;
```
SyrianBallaS: From https://en.cppreference.com/w/cpp/container/set/~set (applies to all polymorphic containers)
`...Note, that if the elements are pointers, the pointed-to objects are not destroyed.`
I'll suggest making `value_type ``std::unique_ptr<NotifyTarget>`.

src/ui/WindowViewModelBase.hh
```diff
@@ -17,7 +19,65 @@ struct Size
     int Height{};
 };
 
+enum class DialogResult
+{
+    None = 0,
```
SyrianBallaS: If you felt it was necessary it's fine, but you don't have to do this.

src/ui/WindowViewModelBase.hh
```diff
@@ -17,7 +19,65 @@ struct Size
...
+    /// </summary>
+    void Show();
+
+    /// <summary>
+    /// Shows a modal window for this view model. Method will not return until the window is closed.
+    /// </summary>
+    DialogResult ShowModal();
+
+protected:
+    WindowViewModelBase() : ViewModelBase() {}
```
SyrianBallaS: Should be marked `noexcept`.

src/ui/viewmodels/MessageBoxViewModel.hh
SyrianBallaS: I think moving them in is fine.

src/ui/viewmodels/MessageBoxViewModel.hh
```diff
@@ -0,0 +1,167 @@
...
+    None,
+    Info,
+    Warning,
+    Error,
+};
+
+class MessageBoxViewModel : public WindowViewModelBase
+{
+public:
+    MessageBoxViewModel() noexcept {}
```
SyrianBallaS: mark this as defaulted.

src/ui/viewmodels/MessageBoxViewModel.hh
```diff
@@ -0,0 +1,167 @@
...
+
+    MessageBoxViewModel(const std::wstring& sMessage) noexcept
```
SyrianBallaS: Mark this as `explicit `to prevent implicit conversion.
And I think it would be better to give some default parameters such as `::GetActiveWindow` for `HWND `and `MessageBoxIcon`::Info for Icon, and `MessageBoxButtons::OK` for the button. Those are just examples, if there are better defaults that you can think of then be my guest.
I think the "specific" message boxes should be subclassed from this one but if you don't want to do it I guess that's fine too.

src/ui/viewmodels/MessageBoxViewModel.hh
```diff
@@ -0,0 +1,167 @@
...
+    {
+        SetMessage(sMessage);
+    }
+
+#pragma push_macro("GetMessage") // windows stole my method name
+#undef GetMessage
+    /// <summary>
+    /// The <see cref="ModelProperty" /> for the message.
+    /// </summary>
+    static StringModelProperty MessageProperty;
```
SyrianBallaS: It's not a big deal but I just want to know the reasoning why these properties are static instead of data members for future reference. I personally never used static properties and preferred auto properties, I can probably make something like later. static property just seems interesting.

src/ui/viewmodels/MessageBoxViewModel.hh
```diff
@@ -0,0 +1,167 @@
...
+
+    /// <summary>
+    /// Sets the buttons to display (default: OK).
+    /// </summary>
+    void SetButtons(MessageBoxButtons nValue) { SetValue(ButtonsProperty, static_cast<int>(nValue)); }
+
+    /// <summary>
+    /// Shows a generic message.
+    /// </summary>
+    static void ShowMessage(const std::wstring& sMessage)
```
SyrianBallaS: Should have mentioned previously in the review. If you really want these to be explicitly static functions I'd suggest putting them in a helper class since static methods aren't member functions. 

The more "correct" way (at least according to Stroustrup) is to make them helper (non-class, non-static) functions in the same namespace the class it's supposed to help.

src/ui/win32/Desktop.cpp
SyrianBallaS: I thought this was supposed to be MVVM and not MVC? Not that I'm complaining though, just a bit confused.

src/ui/win32/Desktop.cpp
```diff
@@ -0,0 +1,63 @@
...
+        if (pController->IsSupported(oViewModel))
+            return pController;
+    }
+
+    return nullptr;
+}
+
+void Desktop::Shutdown()
+{
+}
```
SyrianBallaS: Did you want this method to shutdown the desktop or the application? There are APIs for both.

src/ui/win32/Dlg_MessageBox.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+namespace ra {
+namespace ui {
+namespace win32 {
+
+using ra::ui::viewmodels::MessageBoxIcon;
+using ra::ui::viewmodels::MessageBoxButtons;
+using ra::ui::viewmodels::MessageBoxViewModel;
+
+static HRESULT(WINAPI *pTaskDialog)(HWND hwndParent, HINSTANCE hInstance, PCWSTR pszWindowTitle, PCWSTR pszMainInstruction,
```
SyrianBallaS: I strongly suggest using `std::function` for this callback, seems kind of strange making a global static but not too a big deal. It would look something like this. Adjust it however you like, it's just an example.
```
using TaskDialogCallbackType = HRESULT(WINAPI*)(HWND, HINSTANCE, PCWSTR, PCWSTR, TASKDIALOG_COMMON_BUTTON_FLAGS, PCWSTR, int);
std::function<TaskDialogCallbackType > TaskDialogCallback;
```

`TASKDIALOG_COMMON_BUTTON_FLAGS `is Vista+ so I'd suggest surround this callback with
```
#ifndef _WIN32_WINNT >= _WIN32_WINNT_LONGHORN
...
#endif // _WIN32_WINNT >= _WIN32_WINNT_LONGHORN
```
if supporting Windows XP is still considered a requirement. The "dynamically" finding doesn't really matter because the SDK is for Windows 7, the last platform SDK to support XP.

src/ui/win32/Dlg_MessageBox.cpp
```diff
@@ -0,0 +1,121 @@
...
+    PCWSTR pszContent, TASKDIALOG_COMMON_BUTTON_FLAGS dwCommonButtons, PCWSTR pszIcon, int *pnButton) = nullptr;
+
+Dlg_MessageBox::Controller::Controller() noexcept
+{
+    // TaskDialog isn't supported on WinXP, so we have to dynamically find it.
+    auto hDll = LoadLibraryA("comctl32");
+    if (hDll)
+        pTaskDialog = (HRESULT(WINAPI*)(HWND, HINSTANCE, PCWSTR, PCWSTR, PCWSTR, TASKDIALOG_COMMON_BUTTON_FLAGS, PCWSTR, int *))GetProcAddress(hDll, "TaskDialog");
+}
```
SyrianBallaS: TaskDialog doesn't exist before Vista. 
This function will probably fail on an XP device (which would be undefined by the standard). You should probably just get rid of this. and surround the whole class with a check to see if the platform is Vista or higher.

Jamiras: I _did_ test this in XP. That's the whole reason I'm using `GetProcAddress` instead of the #define already present in the Windows.h. `GetProcAddress` returns NULL on XP, so it'll go into the non-TaskDialog section of the code.

As mentioned above, we can't conditionally compile anything out of the library because the exact same library has to work on everything from XP to 10.

You're right - `FreeLibrary` _should_ be getting called, but as long as we're expecting to use `pTaskDialog`, we can't.  I guess I could make hDll a member variable and call FreeLibrary in the destructor.

src/ui/ModelProperty.cpp
SyrianBallaS: Ok, got you.

src/ui/win32/Desktop.cpp
SyrianBallaS: It sounds like a router. Not sure though.

src/RA_Core.cpp
```diff
@@ -1490,29 +1504,29 @@ BOOL _FileExists(const std::wstring& sFileName)
 std::string GetFolderFromDialog()
 {
     std::string sRetVal;
-	CComPtr<IFileOpenDialog> pDlg;
+    CComPtr<IFileOpenDialog> pDlg;
```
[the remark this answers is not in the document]
SyrianBallaS: Alright, dont worry about it.

src/ui/ModelProperty.hh
```diff
@@ -0,0 +1,107 @@
...
+    const char* GetPropertyName() const { return m_sPropertyName; }
+
+    /// <summary>
+    /// Gets the property for specified unique identifier.
+    /// </summary>
+    /// <returns>Associated property, <c>null</c> if not found.</returns>
+    static const ModelPropertyBase* GetPropertyForKey(int nKey);
+
+    bool operator==(const ModelPropertyBase& that) const { return m_nKey == that.m_nKey; }
+    bool operator!=(const ModelPropertyBase& that) const { return m_nKey != that.m_nKey; }
```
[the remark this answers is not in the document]
SyrianBallaS: Yeah, you make a good point. No difference functionality wise.

src/ui/ModelProperty.hh
```diff
@@ -0,0 +1,107 @@
...
+
+    ModelPropertyBase(int nKey) noexcept { m_nKey = nKey; } // for binary search
+
+private:
+    int m_nKey;
+    const char* m_sTypeName;
+    const char* m_sPropertyName;
+
+    static std::vector<ModelPropertyBase*> s_vProperties;
+    static int s_nNextKey;
```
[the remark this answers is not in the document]
SyrianBallaS: Thanks, I learned something new. Seems good to me.

src/ui/ModelProperty.hh
```diff
@@ -0,0 +1,107 @@
...
+        /// <summary>
+        /// The old value.
+        /// </summary>
+        const T& tOldValue;
+
+        /// <summary>
+        /// The new value.
+        /// </summary>
+        const T& tNewValue;
+    } ChangeArgs;
```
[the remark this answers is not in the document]
SyrianBallaS: It's ok, it happens to everyone.

src/ui/WindowViewModelBase.hh
SyrianBallaS: Seems it cant be helped. There're always exceptions to the rules I guess. I'm guessing its because they're protected but it's fine.

src/ui/viewmodels/MessageBoxViewModel.hh
SyrianBallaS: Valid point. It's mostly a design thing but what you have is fine.

src/ui/win32/Dlg_MessageBox.cpp
SyrianBallaS: Sounds good to me.

src/ui/win32/Desktop.cpp
Jamiras: I've renamed these to `Presenter`

src/ui/ViewModelBase.hh
```diff
@@ -0,0 +1,101 @@
...
+        virtual void OnViewModelBoolValueChanged([[maybe_unused]] const BoolModelProperty::ChangeArgs& args) {}
+        virtual void OnViewModelStringValueChanged([[maybe_unused]] const StringModelProperty::ChangeArgs& args) {}
+        virtual void OnViewModelIntValueChanged([[maybe_unused]] const IntModelProperty::ChangeArgs& args) {}
+    };
+
+    void AddNotifyTarget(NotifyTarget& pTarget) { m_vNotifyTargets.insert(&pTarget); }
+    void RemoveNotifyTarget(NotifyTarget& pTarget) { m_vNotifyTargets.erase(&pTarget); }
+
+protected:
+    ViewModelBase() = default;
```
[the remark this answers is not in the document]
SyrianBallaS: Adjustments are sometimes necessary. I think it's more important for it to be <code>noexcept</code> than default.

src/ui/viewmodels/MessageBoxViewModel.hh
SyrianBallaS: Full .NET WPF must be different than .NET Core's version. Thanks for the explanation.

src/ui/win32/Desktop.cpp
SyrianBallaS: Seems like a better name.

PR #121, opened 2018-09-30
ILogger service

### LINKED ISSUE #71: RA_Log has double newlines
image

Jamiras:
* add ILogger service
* add IClock service (and mock)
* strip newlines from end of log messages (fixes #71) - they're added automatically by the logger
* added timestamps and categories to log messages (existing messages default to Info category) - new categories and the ability to toggle categories will be added in the future.
* log file is now written to `RACache` directory (previously was `RACache\Data`) - old file will be deleted with a later PR that will delete files older than 30 days.

example from new log file:
```
120952.057|INFO| ================================================================================
120952.060|INFO| Log started at 09/30/18 12:09:52 Mountain Daylight Time
120952.060|INFO| BaseDirectory: E:\Games\RetroAchievements\RANES\
120952.061|INFO| Loading preferences...
120952.063|INFO| User-Agent: RANes/0.014.0.0 (WindowsNT 10.0) Integration/0.74.11.1-svc_log
120952.063|INFO| RAWeb::RA_InitializeHTTPThreads called
120952.063|INFO| RAWeb::RA_InitializeHTTPThreads Adding HTTP thread 0 (000038ec, 00000704)
120952.064|INFO| RAWeb::RA_InitializeHTTPThreads Adding HTTP thread 1 (00000d4c, 00000708)
120952.065|INFO| RAWeb::RA_InitializeHTTPThreads Adding HTTP thread 2 (00004ed8, 0000070c)
120952.065|INFO| RAWeb::RA_InitializeHTTPThreads Adding HTTP thread 3 (000032a0, 00000710)
```


src/services/impl/FileLogger.hh
```diff
@@ -0,0 +1,86 @@
...
+
+        std::tm tTimeStruct;
+        localtime_s(&tTimeStruct, &tTime);
+
+        char sBuffer[16];
+        strftime(sBuffer, sizeof(sBuffer), "%H%M%S", &tTimeStruct);
+        sprintf_s(&sBuffer[6], sizeof(sBuffer) - 6, ".%03u|", tMilliseconds);
+
+        {
+            std::lock_guard<std::mutex> oLock(m_oMutex);
```
SyrianBallaS: I strongly suggest using `std::scoped_lock` instead. `std::lock_guard` should only be used for backwards-compatibility (which locks haven't been used before in this project). `std::scoped_lock `doesn't require specifying template arguments at all since they are deduced via perferct forwarding. It also does a better job at preventing deadlocks as it can lock more than one mutex at once.

How this one would look...
```
...
{
  std::scoped_lock oLock{m_oMutex};
...
```

src/services/impl/FileTextWriter.hh
```diff
@@ -17,7 +17,12 @@ class FileTextWriter : public ra::services::TextWriter
 {
 public:
     explicit FileTextWriter(const std::wstring& sFilename) noexcept
-        : m_oStream(sFilename, std::ios::binary)
+        : FileTextWriter(sFilename, std::ios::out)
+    {
+    }
+
+    explicit FileTextWriter(const std::wstring& sFilename, std::ios_base::openmode nMode) noexcept
+        : m_oStream(sFilename, std::ios::binary | nMode)
```
SyrianBallaS: Not too sure how to adjust it but all the ostream classes have std::ios::out on by default and gets added to any flags input by the user.
Could be something like this:
```
...
explicit FileTextWriter(const std::wstring& sFilename, std::ios_base::openmode nMode = std::ios_base::binary) noexcept
...
```
Then you could probably get rid of the constructor above and leave the rest of this one the same. 

One other suggestion could be to subclass this class, friend the subclass, and specialize it for binary writing. You could call it `BinaryFileTextWriter` to make it clear it's for binary ops and not for formatted text. It seems kind of weird that this class has one constructor for formatted text and one for binary text.

From what I can see, it seems much more appropriate to make `std::ios::binary` the default open_mode because of `m_oStream.write` everywhere. Mixing a formatted stream and unformatted stream is a bad idea.

tests/mocks/MockClock.hh
```diff
@@ -0,0 +1,59 @@
...
+    MockClock() noexcept
+        : m_Override(this)
+    {
+        struct tm tNow;
+        tNow.tm_year = 118; // 2018
+        tNow.tm_mon = 7;    // August
+        tNow.tm_mday = 21;  // 21
+        tNow.tm_hour = 16;  // 4 pm
+        tNow.tm_min = 8;    // 4:08
+        tNow.tm_sec = 43;   // 4:08:43
```
SyrianBallaS: I'm going to suggest using aggregate initialization for speed and correctness. This will enforce that all members are initialized in-order. Designations for NRVO (named return value optimization/copy-epsilon) won't exist until C++20 so you'd have to put the object named before the dot. Since you have comments on the end of each line it shouldn't be necessary but might still make things clearer. Using `struct `like that looks kind of strange to me.
Should look like this...
```
std::tm tNow
{
        tNow.tm_sec = 43,   // 4:08:43
        tNow.tm_min = 8;    // 4:08
        tNow.tm_hour = 16;  // 4 pm
        tNow.tm_mday = 21;  // 21
        tNow.tm_mon = 7;    // August
        tNow.tm_year = 118; // 2018
       // The rest are automatically 0
};
```
As you can see, the order of initialization is different than what you had. Without the stuff on the left side, there would be no error, with it there should be (I say should because I haven't really checked if VC++ checks for it, GCC does though). The C++20 version wouldn't have the `tNow `before the dot.

tests/mocks/MockFileSystem.hh
```diff
@@ -65,6 +65,14 @@ public:
...
+    {
+        std::string& sContents = m_mFileContents[sPath];
+        sContents.clear();
+        auto pWriter = std::make_unique<ra::services::impl::StringTextWriter>(sContents);
+        return std::unique_ptr<TextWriter>(pWriter.release());
+    }
+
+    std::unique_ptr<TextWriter> AppendTextFile(const std::wstring& sPath) const override
     {
         std::string& sContents = m_mFileContents[sPath];
```
SyrianBallaS: I still think you should use `std::unordered_map::find` for security if `sPath `may not exist, but don't worry about it I can get to it later.

src/RA_Core.cpp
```diff
@@ -183,12 +183,6 @@ static void InitCommon(HWND hMainHWND, /*enum EmulatorID*/int nEmulatorID, const
             break;
     }
 
-    RA_LOG("=================================================");
-    if (g_sClientName != nullptr)
-        RA_LOG("Initializing for %s %s", g_sClientName, sClientVer);
-    else
-        RA_LOG("Initializing for unknown client %d %s", nEmulatorID, sClientVer);
-
     g_sROMDirLocation[0] = '\0';
```
SyrianBallaS: I don't think `std::string` needs to initialized like this since it's default constructor already does this. Unless I'm missing something.

src/RA_Log.h
```diff
@@ -2,9 +2,41 @@
...
 
-extern void RADebugLogNoFormat(const char* data);
-extern void RADebugLog(const char* sFormat, ...);
+#include "services\ILogger.hh"
 
-#define RA_LOG RADebugLog
+#ifdef RA_UTEST
+
+#define RA_LOG_LEVEL(lvl, ...)
+#define RA_LOG(...)
```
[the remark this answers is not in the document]
Jamiras: RA_LOG is already well-established in the codebase, and expects support for printf-style formatting: i.e. `%08x` and the number of characters following the '%' is not always 1: `%zu`

I don't believe those can be easily handled with your proposed "You don't need to put any letters or numbers around the specifier. Anything works as long as std::ostream can handle it.", and retrofitting all of the existing messages is more than I want to do for this PR.

`Format` serves as an upgrade to the buffer-based vsnprintf previously in `RADebugLog`: https://github.com/RetroAchievements/RAIntegration/pull/121/files#diff-50968615630ba5f4a431d984c17b1e31L43 and populates directly into a `std::string` object.

The macros all just forward the parameters to that.

src/RA_StringUtils.cpp
```diff
@@ -103,4 +105,31 @@ std::string& TrimLineEnding(std::string& str) noexcept
     return str;
 }
 
+_Use_decl_annotations_
+_NODISCARD std::string Format(const char* sFormat, ...)
```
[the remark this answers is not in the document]
Jamiras: Having read a couple pages on variadic templates, I agree that they are preferable to va_args, but in this case, we are forwarding the parameters on to vsnprintf for a specific set of functionality that's not easily reproduced. Perhaps that's something we can come back and attempt to address at a later point. I'm going to rename this method "StringPrintf" to make it clear we're leveraging printf semantics.

src/RA_Log.h
```diff
@@ -2,9 +2,41 @@
...
+
+#else
+
+#include "RA_StringUtils.h"
+
+#include "services\ServiceLocator.hh"
+
+#define RA_LOG_LEVEL(lvl, ...) \
```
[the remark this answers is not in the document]
Jamiras: Not using macros makes it harder to undefine these functions for the unit tests. However, it would be easy enough to provide a MockLogger or something similar to swallow the messages.

src/RA_Log.h
```diff
@@ -2,9 +2,41 @@
...
+        std::string __sFormatted = ra::Format(__VA_ARGS__); \
+        __pLogger.LogMessage(ra::services::LogLevel::Info, ra::TrimLineEnding(__sFormatted)); \
+    } \
+}
+
+#endif
+
+#define RA_LOG_INFO(...) RA_LOG_LEVEL(ra::services::LogLevel::Info, __VA_ARGS__)
+#define RA_LOG_WARN(...) RA_LOG_LEVEL(ra::services::LogLevel::Warn, __VA_ARGS__)
+#define RA_LOG_ERR(...) RA_LOG_LEVEL(ra::services::LogLevel::Error, __VA_ARGS__)
```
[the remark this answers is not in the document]
Jamiras: yep, helper macros to make the calls slightly less verbose.

src/RA_StringUtils.cpp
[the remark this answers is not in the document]
SyrianBallaS: That seems fine to mr.

src/RA_Log.h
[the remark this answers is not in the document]
SyrianBallaS: Dont worry about it. It's a lot of work, best saved for later.

src/services/impl/FileLogger.hh
SyrianBallaS: It is.

tests/mocks/MockFileSystem.hh
Jamiras: i've changed this to use `insert_or_assign`

PR #125, opened 2018-10-05
Add IThreadPool

Jamiras:
* extracts async logic from ra_httpthread for more generic consumption
* interface allows for mocking (not used currently as the only thing happening asynchronously is server calls)
* responses from downloading images are now handled in the background thread (instead of in `RA_HandleHttpResponses`) - greatly decreases flickering when first loading a new game.


src/services/ImageRepository.cpp
```diff
@@ -175,9 +188,17 @@ void ImageRepository::FetchImage(ImageType nType, const std::string& sName)
...
+    {
+        std::string sResponse;
+        if (RAWeb::DoBlockingRequest(nRequestType, args, sResponse))
+            _WriteBufferToFile(sFilename, sResponse);
+
+        {
+            std::lock_guard<std::mutex> lock(m_oMutex);
+            m_vRequestedImages.erase(sFilename);
+        }
+    });
```
Jamiras: Request the image and write it to disk all on the same background thread. No longer relies on the _RA_HandleHTTPResults method to write the image to disk.

Also co-locates the request with the response processing instead of making a roundabout loop through RA_Core.cpp

src/services/impl/FileLogger.hh
```diff
@@ -37,9 +37,19 @@ public:
...
+        {
+            auto tNow = ServiceLocator::Get<IClock>().Now();
+            tMilliseconds = static_cast<unsigned int>(std::chrono::time_point_cast<std::chrono::milliseconds>(tNow).time_since_epoch().count() % 1000);
+            tTime = std::chrono::system_clock::to_time_t(tNow);
+        }
+        else
+        {
+            tTime = time(nullptr);
+            tMilliseconds = 0;
+        }
```
Jamiras: prevents exception using RA_LOG before the IClock is registered or after it's been destroyed. I fixed the issue that caused me to run into this, so it may not be necessary in the long-term.

SyrianBallaS:
This one seems interesting. I'll probably look at this one tomorrow to see if it needs polish.

src/services/impl/ThreadPool.cpp
```diff
@@ -0,0 +1,82 @@
...
+            {
+                RA_LOG_ERR("Exception on background thread: %s", ex.what());
+            }
+        } while (!m_bShutdownInitiated);
+
+        // wait for work
+        if (!m_bShutdownInitiated)
+        {
+            std::unique_lock<std::mutex> lock(m_oMutex);
+            m_cvMutex.wait(lock);
```
SyrianBallaS: The way you have this currently, it might be better to use `std::scoped lock` and change `m_cvMutex` to `std::condition_variable_any` (since C++11, improved in C++14).

src/services/impl/ThreadPool.hh
```diff
@@ -0,0 +1,62 @@
...
+#pragma once
+
+#include "services\IThreadPool.hh"
+
+#include <assert.h>
+#include <functional>
+#include <mutex>
+#include <queue>
+#include <thread>
+#include <vector>
```
SyrianBallaS: Not a big deal since it'll probably change from the pch pr but `mutex `includes `functional `and `thread`. `queue `includes `vector`.

src/services/impl/ThreadPool.hh
```diff
@@ -0,0 +1,62 @@
...
+        return m_bShutdownInitiated;
+    }
+
+private:
+    void RunThread() noexcept;
+
+    std::vector<std::thread> m_vThreads;
+    std::queue<std::function<void()>> m_vQueue;
+    std::mutex m_oMutex;
+    std::condition_variable m_cvMutex;
```
SyrianBallaS: If you're not using a timer or moving the mutex around, locks should be std::scoped_lock.
`m_cvMutex `type should be `std::condition_variable_any`. 
If you still want it to be a `std::condition_variable `variant. You could just make it another mutex and put it in the scoped_lock. scoped_lock unlocks in reverse order as if `std::lock` was called (I don't think it's in VC++ yet though).

src/services/ImageRepository.h
```diff
@@ -104,12 +106,15 @@ class ImageRepository
         std::atomic<unsigned int> m_nReferences;
     };
 
-    typedef std::map<std::string, HBitmapReference> HBitmapMap;
+    typedef std::unordered_map<std::string, HBitmapReference> HBitmapMap;
```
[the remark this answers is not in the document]
Jamiras: ok

src/services/impl/ThreadPool.cpp
```diff
@@ -0,0 +1,82 @@
...
+void ThreadPool::RunThread() noexcept
+{
+    do
+    {
+        // check for work
+        do
+        {
+            std::function<void()> pNext;
+            {
+                std::unique_lock<std::mutex> lock(m_oMutex);
```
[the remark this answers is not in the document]
Jamiras: There isn't a timeout because this is specifically designed to wait until notified. `scoped_lock` doesn't support `condition` variables (`m_cvMutex`) which are used to pop out of the lock.

src/services/impl/ThreadPool.cpp
```diff
@@ -0,0 +1,82 @@
...
+                pNext = std::move(m_vQueue.front());
+                m_vQueue.pop();
+            }
+
+            // do work
+            try
+            {
+                pNext();
+            }
+            catch (std::exception ex)
```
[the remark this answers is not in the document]
Jamiras: Will make it `const&`. The general idea is that a thread exception shouldn't terminate the thread. More specific handling should be done in the worker code.

src/services/impl/ThreadPool.cpp
Jamiras: how is it better? there's plenty of examples of using `unique_lock` and `condition_variable` in this manner, but I can't find any for `scoped_lock` and `condition_variable_any`.

src/services/impl/ThreadPool.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+    m_bShutdownInitiated = true;
+    m_cvMutex.notify_all();
+
+    if (bWait && !m_vThreads.empty())
+    {
+        RA_LOG_INFO("Waiting for background threads");
+
+        for (size_t i = 0; i < m_vThreads.size(); ++i)
+            m_vThreads[i].join();
```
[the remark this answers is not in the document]
Jamiras: ok

src/services/impl/ThreadPool.hh
```diff
@@ -0,0 +1,62 @@
...
+#include <vector>
+
+namespace ra {
+namespace services {
+namespace impl {
+
+class ThreadPool : public IThreadPool
+{
+public:
+    ~ThreadPool();
```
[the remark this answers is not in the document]
Jamiras: ok

src/services/impl/ThreadPool.hh
Jamiras: this implementation seems to be the standard way to do notify/wait mutexes in c++. even the example in the `condition_variable_any` documentation page uses `unique_lock`: https://en.cppreference.com/w/cpp/thread/condition_variable_any/notify_one

src/services/impl/ThreadPool.hh
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+private:
+    void RunThread() noexcept;
+
+    std::vector<std::thread> m_vThreads;
+    std::queue<std::function<void()>> m_vQueue;
+    std::mutex m_oMutex;
+    std::condition_variable m_cvMutex;
+    size_t m_nThreads;
```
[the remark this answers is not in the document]
Jamiras: ok

Jamiras:
In regards to your "general observations". The intent is to have a pool of threads in a wait state. When work is queued, one thread will be awoken (notify_one). It will dequeue the work and execute it. When it finishes, if more work is available, it will do the next work, and so on until no more work is queued. Then the thread will go back into the wait state.

There is no need for timeouts as the threads will sleep until explicitly awoken.

I'm not sold on converting to `scoped_lock` and `condition_variable_any`. All the examples of similar functionality I can find are using `unique_lock` in a manner similar to this implementation.

src/services/impl/ThreadPool.cpp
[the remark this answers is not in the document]
SyrianBallaS: `std::condition_variable_any` supports any `scoped_lock`.

src/services/impl/ThreadPool.cpp
SyrianBallaS: Don't worry about it, performance-wise `unique_lock `and `condition_variable `are better. There's not much added benefit unless you need to lock more than one mutex.

src/services/impl/ThreadPool.hh
SyrianBallaS: Seems good to me.

PR #126, opened 2018-10-05
Enum to enum class (OverlayPage & TransitionState)

SyrianBallaS:
This PR mostly has to do with converting the `enum `s in the title to `enum class`. 
Did some much needed optimizations, and made a few more helper functions to reduce code bloat.
AchievementOverlay::Render in particular needed to be optimized, I ignored other methods that needed that treatment.
There will be some reviews comments after this PR is submitted.

P.S: 
I'm probably not going to rebase my master branch again because of how tedious it was, if the commit history is too large then it might better to just merge my master branch with this repositories master to get rid of those "new" commits (they're already merged but GitHub doesn't track them as merged or not), if it becomes a problem. Most of them are just merging this repo's master branch with mine.

I have two ideas for the next PR (other than `enum to enum class`):
**Precompiled header**
- So we can turn on /Wall for our code only 
     - It won't stay on until all the warnings have been addressed since there's a bunch (over 500)
     - Was thinking of doing one warning code at a time after the precompiled header PR.
- Also speeds up compilation time considerably 
- Emulator's won't depend on it

**Constructor optimizations**
- Unless copying is already specified, it will be deleted
- Will use the standard's core guidelines to help locate where it's needed (part of Visual Studio, not a library)

I'll put more details on discord but I don't know which one is more important here.


src/RA_AchievementOverlay.h
```diff
@@ -4,41 +4,9 @@
...
-enum TransitionState
-{
-    TS_OFF = 0,
-    TS_IN,
-    TS_HOLD,
-    TS_OUT,
-
-    TS__MAX
-};
+#include "services\ImageRepository.h" // map, string
```
SyrianBallaS: For source files it doesn't matter that much, but includes should be limited in headers. The comments are to show the removed includes are already included.

src/RA_AchievementOverlay.h
```diff
@@ -114,20 +82,39 @@ extern AchievementExamine g_AchExamine;
...
+    };
 
-    void Initialize(_UNUSED HINSTANCE);
+    enum class TransitionState
+    {
+        Off,
+        In,
+        Hold,
+        Out
+    };
```
SyrianBallaS: I've never seen these `enums `used outside of `AchievementOverlay` so they are private for now.

src/RA_AchievementOverlay.h
```diff
@@ -157,43 +144,46 @@ class AchievementOverlay
...
-    TransitionState			m_nTransitionState;
-    float					m_fTransitionTimer;
-
-    OverlayPage				m_Pages[5];
-    unsigned int			m_nPageStackPointer;
-
-    ra::services::ImageReference m_hOverlayBackground;
-    ra::services::ImageReference m_hUserImage;
+    static _CONSTANT_VAR PAGE_TRANSITION_IN{ -0.2F };
+    static _CONSTANT_VAR PAGE_TRANSITION_OUT{ 0.2F };
```
SyrianBallaS: Moved some of the constants in the source around. There are other files that have this variable name so it's a private static member. Data members can't be `constexpr `only so they're static as well.

src/RA_AchievementOverlay.h
```diff
@@ -157,43 +144,46 @@ class AchievementOverlay
...
+    BOOL					m_bInputLock{};	//	Waiting for pad release
+    std::vector<NewsItem>	m_LatestNews{};
+    TransitionState			m_nTransitionState{};
+    float					m_fTransitionTimer{ PAGE_TRANSITION_IN };
+
+    std::array<Page, 5>		m_Pages{ Page::Achievements };
+    unsigned int			m_nPageStackPointer{};
+
+    ra::services::ImageReference m_hOverlayBackground{ ra::services::ImageType::Local, "Overlay\\overlayBG.png" };
+    ra::services::ImageReference m_hUserImage{ ra::services::ImageType::UserPic, RAUsers::LocalUser().Username() };
```
SyrianBallaS: This way allows us to not have to define any constructors explicitly. Copying will be deleted in another PR but moving will be kept. 

The global is reinitialized using a move assign in `RA_Core.cpp (InitCommon)` after the image repository is initialized. It had to "reinitizlied" because the `AchievementOverlay `global is may or may not be initialized before the `ImageRepository` global.

The "`Initialize`" method is gone because it's initialized during construction.

src/RA_Core.cpp
```diff
@@ -69,11 +69,10 @@ const char* g_sClientDownloadURL = nullptr;
 const char* g_sClientEXEName = nullptr;
 bool g_bRAMTamperedWith = false;
 
-static const unsigned int PROCESS_WAIT_TIME = 100;
-static unsigned int g_nProcessTimer = 0;
+inline static constexpr unsigned int PROCESS_WAIT_TIME{ 100U };
+inline static unsigned int g_nProcessTimer{ 0U };
```
SyrianBallaS: Optimiztion

src/RA_Core.cpp
```diff
@@ -220,7 +219,7 @@ static void InitCommon(HWND hMainHWND, /*enum EmulatorID*/int nEmulatorID, const
     //////////////////////////////////////////////////////////////////////////
     //	Image rendering: Setup image factory and overlay
     ra::services::g_ImageRepository.Initialize();
-    g_AchievementOverlay.Initialize(g_hThisDLLInst);
+    g_AchievementOverlay = AchievementOverlay{};
```
SyrianBallaS: The global will mostly be the same except the `ImageReference `members will actually have something this time.

src/RA_Integration.vcxproj
```diff
@@ -353,11 +353,14 @@
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <DisableSpecificWarnings>4091</DisableSpecificWarnings>
       <ConformanceMode>true</ConformanceMode>
+      <LinkTimeCodeGeneration>UseFastLinkTimeCodeGeneration</LinkTimeCodeGeneration>
```
SyrianBallaS: During release mode, the linker keeps on turning this on. So I just turned it on explicitly.

src/ra_fwd.h
```diff
@@ -43,7 +43,9 @@ using LPVOID = void*;
 #endif // !_WINDEF_
 
 #ifndef _WINNT_
-using HANDLE = void*;
+using HANDLE  = void*;
+using LPCTSTR = const TCHAR*;
+using LPTSTR  = TCHAR* ;
```
SyrianBallaS: I moved these here to not obscure the type.
They were previously in the `ra `namespace, now they're in the global namespace.

src/ra_type_traits.h
```diff
@@ -29,14 +29,23 @@ namespace detail {
 /// <remarks><c>char16_t</c> and <c>char32_t</c> are not considered valid by this type_trait.</remarks>
 template<typename CharT>
 struct _NODISCARD is_char :
-    std::bool_constant<(std::_Is_character<CharT>::value || std::is_same_v<CharT, wchar_t>)>
+    std::bool_constant<((std::is_same_v<CharT, char> || std::is_same_v<CharT, signed char>)
```
SyrianBallaS: What I had before was non-standard (despite being in namespace `std`). plus now we don't have to worry about `CharT `being `unsigned char`. This `type_trait `was used mostly for conversions..

src/ra_utility.h
```diff
@@ -10,12 +10,40 @@
...
+template<
+    typename NarrowedType,
+    typename WideType,
+    class = std::enable_if_t<
+    std::is_arithmetic_v<NarrowedType> &&
+    std::is_arithmetic_v<WideType> &&
+    has_smaller_or_same_size_than_v<NarrowedType, WideType>>
+> _NODISCARD _CONSTANT_FN
+narrow_cast(_In_ WideType from) noexcept { return static_cast<NarrowedType>(static_cast<WideType>(from)); }
+
```
SyrianBallaS: Even though we currently don't have an x86_64 build, I remembered another comment in different PR about conditional compilation between architectures.

Originally, I was going to make the requirements `ra::has_smaller_size_than_v`, but most platform types are the same as the expected type in `x86 `based processors, they are usually different in `x86_64`. So I made a bit less strict to not contradict the purpose of reducing code-bloat.

This is usually used in `WinAPI `functions where the parameter type is `int `but the input is `size_t`. Would `boxed_cast `be a better name?

src/ra_utility.h
```diff
@@ -122,6 +147,29 @@ tstrtoul(_In_z_ LPCTSTR _String,
...
+    TBD: Annotate ANSI compatibility (509 bytes even after concatenation) or Microsoft's (2,048 bytes)?
+    Error Checking: First part might happen in MBCS mode but the standard does not reserve an error code for it.
+    Remarks: The second part is Microsoft Specific. A single null-terminated c string may not exceed 2048 bytes,
+             but may reach a total size of 65,535 bytes after concatenation.
+    This check will only occur during code analysis, the standard does not have data contracts yet,
+    i.e, expects, ensures... (GSL does though however).
+*/
+_Success_((return != to_unsigned(-1)) && (return <= 2048U))
+/* Returns the length of a null-terminated byte string or wide-character string depending on the character set specified */
+_NODISCARD inline auto __cdecl tstrlen(_In_z_ LPCTSTR _Str) noexcept
```
SyrianBallaS: I was just trying to explain what the annotation mean. Should I get rid of it?

src/ra_utility.h
```diff
@@ -131,40 +179,40 @@ _END_EXTERN_C
 namespace rel_ops {
 
 template<typename Enum, typename = std::enable_if_t<std::is_enum_v<Enum>>> _NODISCARD _CONSTANT_FN
-operator==(_In_ const Enum a, _In_ const std::underlying_type_t<Enum> b) noexcept { return (etoi(a) == b); }
+operator==(_In_ Enum a, _In_ std::underlying_type_t<Enum> b) noexcept { return (etoi(a) == b); }
```
SyrianBallaS: The `const `was probably redundant for a value type.

src/services/ImageRepository.h
```diff
@@ -20,37 +20,65 @@ enum class ImageType
...
         // don't copy m_hBitmap, it'll get initialized when GetHBitmap is called, which
         // will ensure the reference count is accurate.
     }
 
-    ~ImageReference();
-    
-    /// <summary>
-    /// Updates the referenced image.
-    /// </summary>
+    ImageReference& operator=(const ImageReference&) noexcept = delete;
```
SyrianBallaS: Wasn't really sure if we needed copy assign because of the comment regarding not copying `m_hBitmap`. Code still compiles and images still show.

src/services/ImageRepository.h
```diff
@@ -20,37 +20,65 @@ enum class ImageType
...
+
+        source.m_nType = ImageType{};
+        if (source.m_hBitmap != nullptr)
+        {
+            DeleteBitmap(source.m_hBitmap);
+            source.m_hBitmap = nullptr;
+        }
+
+        return *this;
+    }
```
SyrianBallaS: I needed to make sure `AchievementOverlay `wouldn't throw an exception during the move assign in `RA_Core.cpp`. From an end-user perspective nothing has changed, memory usage did go down a little bit though.

Images still show up.

src/RA_AchievementOverlay.cpp
```diff
@@ -15,30 +15,24 @@
...
+    _T(" Messages "),
+    _T(" News "),
+    _T(" Leaderboards "),
+    _T(" Achievement Info "),
+    _T(" Achievement Compare "),
+    _T(" Friend Info "),
+    _T(" Friend Add "),
+    _T(" Leaderboard Examine "),
+    _T(" Message Viewer ")
 };
```
SyrianBallaS: I moved all the other constants somewhere else, `NUM_MESSAGES_TO_DRAW `is gone because it wasn't used.

src/RA_AchievementOverlay.cpp
```diff
@@ -74,22 +68,22 @@ const unsigned int OVERLAY_HEIGHT = 1024;
 
 void AchievementOverlay::SelectNextTopLevelPage(BOOL bPressedRight)
 {
-    switch (m_Pages[m_nPageStackPointer])
+    switch (m_Pages.at(m_nPageStackPointer))
     {
-        case OP_ACHIEVEMENTS:
-            m_Pages[m_nPageStackPointer] = (bPressedRight ? OP_FRIENDS : OP_LEADERBOARDS);
+        case Page::Achievements:
+            m_Pages.at(m_nPageStackPointer) = (bPressedRight ? Page::Friends : Page::Leaderboards);
```
SyrianBallaS: `at `won't throw an exception with `std::array` because it is not a polymorphic container.

src/RA_AchievementOverlay.cpp
```diff
@@ -1123,192 +1078,188 @@ void AchievementOverlay::Render(HDC hRealDC, RECT* rcDest) const
...
-    const int nHeight = rcTarget.bottom - rcTarget.top;
-
+    RECT rcTarget{ *rcDest };
+    const auto lHeight{ rcTarget.bottom - rcTarget.top };
+    {
+        _CONSTANT_LOC nFontSize1{ 32 };
+        _CONSTANT_LOC nFontSize2{ 26 };
+        _CONSTANT_LOC nFontSize3{ 22 };
+        _CONSTANT_LOC nFontSize4{ 16 };
+        _CONSTANT_LOC FONT_TO_USE{ _T("Tahoma") };
```
SyrianBallaS: The sheer size of this method needed to be broken up. I didn't want to break anything right now so they are scoped. Those scopes could probably be member/helper functions depending on whether they need access to the class's private members.

src/RA_AchievementOverlay.cpp
```diff
@@ -1123,192 +1078,188 @@ void AchievementOverlay::Render(HDC hRealDC, RECT* rcDest) const
...
-        sprintf_s(buffer, 1024, " ->:%s ", "Next");
-        TextOut(hDC, nRightPx - nControlsX1, nControlsY1, NativeStr(buffer).c_str(), strlen(buffer));
-
-        sprintf_s(buffer, 1024, " <-:%s ", "Prev");
-        TextOut(hDC, nRightPx - nControlsX1, nControlsY2, NativeStr(buffer).c_str(), strlen(buffer));
+        {
+            ra::tstring stNext{ _T(" ->:") };
+            stNext += _T("Next ");
+            TextOut(hDC, nRightPx - nControlsX1, nControlsY1, stNext.c_str(), ra::narrow_cast<int>(stNext.length()));
+        }
```
SyrianBallaS: An example of where `ra::narrow_cast` was used. Boxing happens internally so we didn't have to do it here.

src/services/ImageRepository.h
```diff
@@ -20,37 +20,65 @@ enum class ImageType
...
+    ImageReference() noexcept = default;
+    explicit ImageReference(ImageType nType, const std::string& sName) noexcept
     {
         ChangeReference(nType, sName);
     }
 
-    ImageReference(const ImageReference& source)
-        : m_nType(source.m_nType), m_sName(source.m_sName)
+    ImageReference(const ImageReference& source) noexcept :
+        m_nType{ source.m_nType }, m_sName{ source.m_sName }
```
SyrianBallaS: Copy constructor/assignment isn't required to be `noexcept `by the standard but since the defaulted copy constructor was `noexcept`, this was `noexcept `as well.

This was all for `AchievementOverlay`. I ignored the other classes for later.

src/RA_AchievementOverlay.h
```diff
@@ -157,43 +144,46 @@ class AchievementOverlay
...
+    int	m_nMessagesSelectedItem{};
+    int	m_nNewsSelectedItem{};
+    int	m_nLeaderboardSelectedItem{};
+
+    mutable int m_nNumAchievementsBeingRendered{};
+    mutable int m_nNumFriendsBeingRendered{};
+    mutable int m_nNumLeaderboardsBeingRendered{};
+
+    BOOL                  m_bInputLock{}; // Waiting for pad release
+    std::vector<NewsItem> m_LatestNews{};
```
SyrianBallaS: In theory, this shouldn't be necessary. Since there was a clearmethod called in the removed Initializemethod, this guarantees it's in a default state.

src/RA_AchievementOverlay.h
Jamiras: Using `Username` in this manner seems dangerous. I expect the `g_AchievementOverlay` instance to be initialized before we're logged in.

Jamiras: Oh, I see what you're doing. "reinitializing" seems like a hack. Can you just add an `UpdateImages()` method that resets these two fields instead of the entire object?

Jamiras:
> I'm probably not going to rebase my master branch again because of how tedious it was

That's fine. I almost always squash when merging to master, so feel free to create as many merge commits as you need to keep your branch functioning.

> I have two ideas for the next PR
> I don't know which one is more important

I'd prefer the precompiled headers first, mostly because I have a better understanding of what that entails and implies.

src/RA_AchievementOverlay.cpp
```diff
@@ -1499,6 +1493,13 @@ void AchievementOverlay::InstallNewsArticlesFromFile()
     }
 }
 
+void AchievementOverlay::UpdateImages() noexcept
+{
+    using ra::services::ImageReference;
+    m_hOverlayBackground = ImageReference{ ra::services::ImageType::Local, "Overlay\\overlayBG.png" };
+    m_hUserImage         = ImageReference{ ra::services::ImageType::UserPic, RAUsers::LocalUser().Username() };
```
SyrianBallaS: These'll use `ra::services::ImageReference`'s `move assignment`, it moves it from a temporary (`rvalue reference`).

src/RA_AchievementOverlay.h
```diff
@@ -155,8 +160,8 @@ class AchievementOverlay
     };
 
 private:
-    static _CONSTANT_VAR PAGE_TRANSITION_IN{ -0.2F };
-    static _CONSTANT_VAR PAGE_TRANSITION_OUT{ 0.2F };
+    inline static constexpr auto PAGE_TRANSITION_IN{ -0.200F };
+    inline static constexpr auto PAGE_TRANSITION_OUT{ 0.2F };
```
SyrianBallaS: I thought the other way might've looked too confusing.

Issue #82, opened 2018-08-07
Achievement cannot be created with MemAddr larger than 2K


Jamiras:
As a developer, I want to create an achievement with many conditions. The UI doesn't limit me, but things start going back when the serialized string exceeds ~1900 bytes.

There's two major issues contributing to this.
1) `AchievementSet::SaveToFile` uses a 2K buffer to write each line of the XXX-User.txt file. With the additional space taken by the achievement title, description, and other attributes that leaves around 1900 bytes for the MemAddr string.
2) `AchievementSet::LoadFromFile` uses a 4K buffer read each line of the XXX-User.txt file, so even if you used another tool (RATools/Notepad) to generate a more complete achievement, it still couldn't be loaded.

These methods should be modified to support larger achievements now that the database does.

meleu:
@Jamiras is this still an issue on 0.73?

Jamiras:
Yes, I didn't get around to fixing it in the current implementation, and have some ideas that will simplify the solution after some other planned changes.

PR #127, opened 2018-10-06
Add ILocalStorage service

Jamiras:
* use ILocalStorage for game data, local achievements, code notes, and rich presence
* rename `RACache\Data\XXX.txt` to `RACache\Data\XXX.json`
* rename `RACache\Data\XXX-Notes2.txt` to `RACache\Data\XXX-Notes.json`
* support for serialized triggers greater than 2K (#82)
* restrict Achievement Editor to 200 conditions (per MAX_CONDITIONS define)

### DIFF
--- a/src/RA_AchievementSet.cpp
+++ b/src/RA_AchievementSet.cpp
@@ -359,8 +359,11 @@ bool AchievementSet::LoadFromFile(ra::GameID nGameID)
             return false;
         }
 
-        g_pCurrentGameData->ParseData(doc);
-        nGameID = g_pCurrentGameData->GetGameID();
+        if (m_nSetType == Type::Core)
+        {
+            g_pCurrentGameData->ParseData(doc);
+            nGameID = g_pCurrentGameData->GetGameID();
+        }
 
         const auto& AchievementsData{ doc["Achievements"] };
         for (const auto& achData : AchievementsData.GetArray())
@@ -382,14 +385,6 @@ bool AchievementSet::LoadFromFile(ra::GameID nGameID)
 
         if (m_nSetType != Type::Core)
             return true;
-
-        //	Rich Presence
-        {
-            auto pRich = pLocalStorage.WriteText(ra::services::StorageItemType::RichPresence, std::to_wstring(static_cast<unsigned int>(nGameID)));
-            pRich->Write(g_pCurrentGameData->RichPresencePatch());
-        }
-
-        g_RichPresenceInterpreter.ParseFromString(g_pCurrentGameData->RichPresencePatch().c_str());
 
         const auto& LeaderboardsData{ doc["Leaderboards"] };
         for (const auto& lbData : LeaderboardsData.GetArray())
--- a/src/RA_AchievementSet.cpp
+++ b/src/RA_AchievementSet.cpp
@@ -12,6 +12,7 @@
 
 #include "services\IConfiguration.hh"
 #include "services\ILeaderboardManager.hh"
+#include "services\ILocalStorage.hh"
 #include "services\ServiceLocator.hh"
 
 #include <array>
@@ -37,40 +38,6 @@ void RASetAchievementCollection(AchievementSet::Type Type) noexcept
 {
     g_nActiveAchievementSet = Type;
     g_pActiveAchievements = *ACH_SETS.at(ra::etoi(Type));
-}
-
-_Use_decl_annotations_
-std::wstring AchievementSet::GetAchievementSetFilename(ra::GameID nGameID) noexcept
-{
-    std::wstring str;
-    {
-        std::wostringstream oss;
-        oss << g_sHomeDir << RA_DIR_DATA << nGameID;
-        if (m_nSetType == Type::Local)
-            oss << L"-User";
-        oss << L".txt";
-        str = oss.str();
-    }
-    return str;
-}
-
-BOOL AchievementSet::DeletePatchFile(ra::GameID nGameID)
-{
-    if (nGameID == 0)
-    {
-        //	Nothing to do
-        return TRUE;
-    }
-
-    if (m_nSetType == Type::Local)
-    {
-        //	We do not automatically delete Local patch files
-        return TRUE;
-    }
-
-    //	Remove the text file
-    std::wstring sFilename = GetAchievementSetFilename(nGameID);
-    return RemoveFileIfExists(g_sHomeDir + L"\\" + sFilename);
 }
 
 //static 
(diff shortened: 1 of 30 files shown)


src/RA_AchievementSet.cpp
```diff
@@ -269,72 +236,43 @@ void AchievementSet::Reset()
...
-
-            fwrite(sNextLine, sizeof(char), strlen(sNextLine), pFile);
-        }
-
-        fclose(pFile);
-        return TRUE;
+        Achievement* pAch = &g_pLocalAchievements->GetAchievement(i);
+
+        pData->Write(std::to_string(pAch->ID()));
+        pData->Write(":");
```
SyrianBallaS: Not a big deal now but in the future we could probably write a parameter pack overload for this function but require that every argument is a string to not require a specifier. Might not be exactly what we need but it could look like this. I did look at the Write Methods, these are just examples to show a concept, some of it might not even exist.

```
// Default:
template<typename String, typename = std::enable_if_t<is_string_v<String>>>
void Write(const String& str) noexcept { /*Do the default thing*/ }

// Variadic Version
template<typename String, typename ...Strings>
void Write(const String& str, String&&... strs) noexcept 
{ 
   static_assert(std::is_convertible_v<String, Strings>);
   for(auto i = 0U; i < sizeof...(strs); i++) // sizeof... reports the number of elements in the expansion
      Write(str, std::forward<Strings>(strs)...); // write str, forward the next strs as str   
}
```
I'm going to try something like this myself and see if I can suggest something more practical.
The intention is to have it write out like this...
`pData->Write(std::to_string(pAch->ID()), ":", pAch->CreateMemString(), ":", .... /*the other ones*/);`

src/RA_AchievementSet.cpp
```diff
@@ -353,27 +291,15 @@ BOOL AchievementSet::FetchFromWebBlocking(ra::GameID nGameID)
...
-        {
-            std::wostringstream oss;
-            oss << g_sHomeDir << RA_DIR_DATA << nGameID << L".txt";
-            sAchSetFileName = oss.str();
-        }
-
-        std::ofstream ofile{ sAchSetFileName };
-        if (!ofile.is_open())
+        auto& pLocalStorage = ra::services::ServiceLocator::GetMutable<ra::services::ILocalStorage>();
+        auto pData = pLocalStorage.WriteText(ra::services::StorageItemType::GameData, std::to_wstring(static_cast<unsigned int>(nGameID)));
```
SyrianBallaS: Not a big deal but the `static_cast<unsigned int>` is redundant. `ra::GameID` _uses_ the`std::size_t` definition, if `std::to_wstring` works for `std::size_t` it should work for `ra::GameID`.
With a `using `(type alias) the alias is identical to the original type, not different.

src/RA_AchievementSet.cpp
```diff
@@ -388,97 +314,56 @@ BOOL AchievementSet::FetchFromWebBlocking(ra::GameID nGameID)
...
+        return true;
 
-    //	Store this: we are now assuming this is the correct checksum if we have a file for it
-    g_pCurrentGameData->SetGameID(nGameID);
+    auto& pLocalStorage = ra::services::ServiceLocator::GetMutable<ra::services::ILocalStorage>();
 
     if (m_nSetType == Type::Local)
     {
-        auto buffer{ std::make_unique<char[]>(4096U) };
+        auto pData = pLocalStorage.ReadText(ra::services::StorageItemType::UserAchievements, std::to_wstring(static_cast<unsigned int>(nGameID)));
```
SyrianBallaS: redundant cast around `nGameID`.

src/RA_AchievementSet.cpp
```diff
@@ -388,97 +314,56 @@ BOOL AchievementSet::FetchFromWebBlocking(ra::GameID nGameID)
...
+        pData->GetLine(sLine); // version used to create the file
+        pData->GetLine(sLine); // game title
 
-        //	Get game title:
-        ifile.getline(buffer.get(), 4096LL);
+        while (pData->GetLine(sLine))
         {
-            const auto nCharsRead{ ifile.gcount() };
-            if (nCharsRead > 0)
+            if (!sLine.empty() && isdigit(sLine[0]))
```
SyrianBallaS: Prefer bounds checking whenever possible.
`sLine[0]` could be` sLine.at(0)` or `sLine.front()`.
Sometimes we can't but that'll get solved over time.

src/RA_AchievementSet.cpp
```diff
@@ -388,97 +314,56 @@ BOOL AchievementSet::FetchFromWebBlocking(ra::GameID nGameID)
...
 
-        return TRUE;
+        return true;
     }
     else
     {
-        rapidjson::Document doc;
-        rapidjson::IStreamWrapper isw{ ifile };
-        doc.ParseStream(isw);
+        auto pData = pLocalStorage.ReadText(ra::services::StorageItemType::GameData, std::to_wstring(static_cast<unsigned int>(nGameID)));
```
SyrianBallaS: Redundant cast around `nGameID`.

src/RA_CodeNotes.cpp
```diff
@@ -9,27 +9,25 @@
...
 
-size_t CodeNotes::Load(const std::wstring& sFile)
+size_t CodeNotes::Load(ra::GameID nID)
 {
     Clear();
 
-    std::ifstream ifile{ sFile };
-    if (!ifile.is_open())
+    auto& pLocalStorage = ra::services::ServiceLocator::GetMutable<ra::services::ILocalStorage>();
+    auto pData = pLocalStorage.ReadText(ra::services::StorageItemType::CodeNotes, std::to_wstring(static_cast<unsigned int>(nID)));
```
SyrianBallaS: Redundant cast around `nID`.

src/RA_CodeNotes.cpp
```diff
@@ -65,7 +63,14 @@ void CodeNotes::OnCodeNotesResponse(rapidjson::Document& doc)
     //	Persist then reload
     const ra::GameID nGameID = doc["GameID"].GetUint();
 
-    _WriteBufferToFile(g_sHomeDir + RA_DIR_DATA + std::to_wstring(nGameID) + L"-Notes2.txt", doc);
+    auto& pLocalStorage = ra::services::ServiceLocator::GetMutable<ra::services::ILocalStorage>();
+    auto pData = pLocalStorage.WriteText(ra::services::StorageItemType::CodeNotes, std::to_wstring(static_cast<unsigned int>(nGameID)));
```
SyrianBallaS: Redundant cast around `nID`.

src/RA_GameData.cpp
```diff
@@ -1,12 +1,25 @@
...
-    m_sRichPresencePatch = doc["RichPresencePatch"].IsNull() ? "" : doc["RichPresencePatch"].GetString();
+
+    if (doc.HasMember("RichPresencePatch"))
+    {
+        std::string sRichPresence;
+        if (!doc["RichPresencePatch"].IsNull())
+            sRichPresence = doc["RichPresencePatch"].GetString();
+
+        auto& pLocalStorage = ra::services::ServiceLocator::GetMutable<ra::services::ILocalStorage>();
+        auto pRich = pLocalStorage.WriteText(ra::services::StorageItemType::RichPresence, std::to_wstring(static_cast<unsigned int>(m_nGameID)));
```
SyrianBallaS: Redundant cast around `m_nGameID`.

src/RA_RichPresence.cpp
```diff
@@ -177,7 +181,24 @@ static bool GetLine(std::stringstream& stream, std::string& sLine)
...
+bool RA_RichPresenceInterpreter::Load()
+{
+    m_vLookups.clear();
+    m_vDisplayStrings.clear();
+
+#ifdef RA_UTEST
+    return false;
+#else
+    auto& pLocalStorage = ra::services::ServiceLocator::GetMutable<ra::services::ILocalStorage>();
+    auto pRich = pLocalStorage.ReadText(ra::services::StorageItemType::RichPresence, std::to_wstring(static_cast<unsigned int>(g_pCurrentGameData->GetGameID())));
```
SyrianBallaS: Redundant cast around `g_pCurrentGameData->GetGameID()`.

src/RA_RichPresence.cpp
```diff
@@ -261,17 +281,19 @@ void RA_RichPresenceInterpreter::ParseFromString(const char* sRichPresence)
...
+    if (sDisplayString.empty())
+        return false;
 
-        auto& displayString = m_vDisplayStrings.emplace_back();
-        displayString.InitializeParts(sDisplayString, mFormats, m_vLookups);
+    for (std::vector<std::pair<std::string, std::string>>::const_iterator iter = mDisplayStrings.begin(); iter != mDisplayStrings.end(); ++iter)
+    {
+        auto& displayString = m_vDisplayStrings.emplace_back(iter->first);
+        displayString.InitializeParts(iter->second, mFormats, m_vLookups);
     }
```
SyrianBallaS: This for loop can probably look cleaner as a ranged for. Kind of like this...
```
for(auto& sPair : mDisplayStrings)
{
   auto& displayString = m_vDisplayStrings.emplace_back(sPair.first);
   displayString.InitializeParts(sPair.second, mFormats, m_vLookups);
}
```
The only time I can think of to not use a ranged for loop is if you need rewind or advance the iterator, or if the reference passed isn't used.

src/RA_RichPresence.h
```diff
@@ -5,18 +5,22 @@
 #include "RA_Condition.h"
 #include "RA_MemValue.h"
 
+#include "services\TextReader.hh"
+
 class RA_RichPresenceInterpreter
 {
 public:
     RA_RichPresenceInterpreter() {}
```
SyrianBallaS: Not a big deal, but since you're already here this should be marked as `default `and `noexcept`. Another option could be just to get rid of it for now.

src/services/impl/FileLocalStorage.cpp
```diff
@@ -0,0 +1,81 @@
...
+_CONSTANT_VAR RA_DIR_USERPIC = L"RACache\\UserPic\\";
+_CONSTANT_VAR RA_DIR_BOOKMARKS = L"RACache\\Bookmarks\\";
+
+static void EnsureDirectoryExists(const ra::services::IFileSystem& pFileSystem, const std::wstring& sDirectory)
+{
+    if (!pFileSystem.DirectoryExists(sDirectory))
+        pFileSystem.CreateDirectory(sDirectory);
+}
+
+FileLocalStorage::FileLocalStorage(IFileSystem& pFileSystem)
```
SyrianBallaS: This should be `noexcept `and also `explicit `in the header.

src/RA_AchievementSet.cpp
Jamiras: Changed. Though I was sure I had to put that in for some reason - it's not something I would have thought to do without a warning/error.

src/RA_AchievementSet.cpp
Jamiras: ok

src/RA_AchievementSet.cpp
Jamiras: `empty` handles the bounds checking. 

Why is either of the alternatives better?

src/RA_AchievementSet.cpp
Jamiras: ok

src/RA_CodeNotes.cpp
Jamiras: ok

src/RA_CodeNotes.cpp
Jamiras: ok

src/RA_GameData.cpp
Jamiras: ok

src/RA_RichPresence.cpp
Jamiras: ok

src/RA_RichPresence.cpp
Jamiras: not new code - indention changed - but will modify as suggested.

src/RA_RichPresence.h
Jamiras: done

src/services/impl/FileLocalStorage.hh
```diff
@@ -0,0 +1,31 @@
...
+#include "services\ILocalStorage.hh"
+
+namespace ra {
+namespace services {
+namespace impl {
+
+class FileLocalStorage : public ILocalStorage
+{
+public:
+    FileLocalStorage(IFileSystem& pFileSystem);
```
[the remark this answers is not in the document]
Jamiras: done

src/services/impl/FileLocalStorage.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    std::wstring sPath = pFileSystem.BaseDirectory();
+
+    switch (nType)
+    {
+        case StorageItemType::GameData:
+            return sPath + RA_DIR_DATA + sKey + L".json";
+
+        case StorageItemType::CodeNotes:
+            return sPath + RA_DIR_DATA + sKey + L"-Notes.json";
```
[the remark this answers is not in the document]
Jamiras: You'd expect that by now a compiler would understand string concatenation and optimize for it. It disappoints me that things like `append("foo")` aren't automatically converted to `append("foo", 3)` by the compiler. The internal call to `strlen` could be done at compile time.

I'll change to use `append`.

tests/mocks/MockFileSystem.hh
```diff
@@ -41,16 +41,16 @@ public:
...
 
-    std::string& GetFileContents(const std::wstring& sPath)
+    const std::string& GetFileContents(const std::wstring& sPath)
     {
         auto pIter = m_mFileContents.find(sPath);
         if (pIter != m_mFileContents.end())
             return pIter->second;
 
-        static std::string sEmpty;
+        static const std::string sEmpty;
```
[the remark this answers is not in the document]
Jamiras: This method is returning a reference. All of your proposals would result in a temporary, and referencing a temporary is bad, isn't it?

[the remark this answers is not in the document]
SyrianBallaS: You are totally right, I must've been half asleep.

src/RA_AchievementSet.cpp
SyrianBallaS: All I know is the index operator should be avoided unless you have no choice.
If you don't want to change it that's fine as well.

In this case it's not a big deal, but indexing like this can cause loads of security issues in other cases.

src/RA_RichPresence.cpp
SyrianBallaS: Sorry I hadn't realized. Thanks anyway.

src/RA_AchievementSet.cpp
Jamiras: will change to `front`.

PR #131, opened 2018-10-08
Rich Presence view model

Jamiras:
* add RichPresenceMonitorViewModel and RichPresenceDialog
* add DialogBase for routing Win32 messages
* add WindowBinding for connecting viewmodel to dialog
* hook up position/size remembering through binding
* renamed `Dlg_MessageBox` to `MessageBoxDialog`

NOTE: the timer is still in RichPresenceDialog. I plan to be move it to RichPresenceInterpretter once ITimerService supports scheduled callbacks, which would remove the DialogProc and OnDestroy overrides, StartTimer, StopTimer, and UpdateDisplayString from RichPresenceDialog.


src/RA_Core.cpp
```diff
@@ -1220,18 +1212,11 @@ API void CCONV _RA_InvokeDialog(LPARAM nID)
...
-                g_RichPresenceDialog.InstallHWND(CreateDialog(g_hThisDLLInst, MAKEINTRESOURCE(IDD_RA_RICHPRESENCE), g_RAMainWnd, &Dlg_RichPresence::s_RichPresenceDialogProc));
-            if (g_RichPresenceDialog.GetHWND() != nullptr)
-                ShowWindow(g_RichPresenceDialog.GetHWND(), SW_SHOW);
-
-            if (bRichPresenceExists)
-                g_RichPresenceDialog.StartMonitoring();
-            else
-                g_RichPresenceDialog.ClearMessage();
+            auto& pWindowManager = ra::services::ServiceLocator::GetMutable<ra::ui::viewmodels::WindowManager>();
+            pWindowManager.RichPresenceMonitor.Show();
```
Jamiras: `WindowManager` service serves as a container for the global viewmodels.

src/ui/BindingBase.hh
```diff
@@ -0,0 +1,91 @@
+#ifndef RA_UI_BINDINGBASE_H
+#define RA_UI_BINDINGBASE_H
+#pragma once
+
+#include "ui/ViewModelBase.hh"
+
+namespace ra {
+namespace ui {
+
+class BindingBase : protected ViewModelBase::NotifyTarget
```
Jamiras: `BindingBase` attaches to a `ViewModelBase` through the `NotifyTarget` callback. It also has methods for reading data from the `ViewModelBase` generically using a `ModelProperty`

src/ui/viewmodels/RichPresenceMonitorViewModel.cpp
```diff
@@ -0,0 +1,37 @@
...
+    else if (!g_RichPresenceInterpreter.Enabled())
+    {
+        SetDisplayString(DisplayStringProperty.GetDefaultValue());
+    }
+    else
+    {
+        std::wstring sDisplayString = ra::Widen(g_RichPresenceInterpreter.GetRichPresenceString());
+        SetDisplayString(sDisplayString);
+    }
+}
```
Jamiras: `UpdateDisplayString` is called every minute to update the view model message. When it changes, the binding is notified, which will update the UI.

src/ui/win32/RichPresenceDialog.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// ------------------------------------
+
+RichPresenceDialog::RichPresenceDialog(ra::ui::viewmodels::RichPresenceMonitorViewModel& vmRichPresenceDisplay)
+    : DialogBase(vmRichPresenceDisplay)
+{
+    m_hFont = CreateFont(15, 0, 0, 0, FW_NORMAL, FALSE, FALSE, FALSE, DEFAULT_CHARSET, OUT_DEFAULT_PRECIS, CLIP_DEFAULT_PRECIS, CLEARTYPE_QUALITY, VARIABLE_PITCH, nullptr);
+
+    m_bindWindow.SetInitialPosition(ra::ui::win32::bindings::RelativePosition::After, ra::ui::win32::bindings::RelativePosition::Far, "Rich Presence Monitor");
```
Jamiras: Specifies that the window size/position should be persisted between sessions via the "Rich Presence Monitor" key. If a previous value is not found, it should appear to the right of the main window (After) aligned with the bottom of the main window (Far).

src/ui/win32/RichPresenceDialog.cpp
```diff
@@ -0,0 +1,113 @@
...
+    m_bindWindow.BindLabel(IDC_RA_RICHPRESENCERESULTTEXT, ra::ui::viewmodels::RichPresenceMonitorViewModel::DisplayStringProperty);
```
Jamiras: Binds the `DisplayString` property of the viewmodel to the Text control identified as `IDC_RA_RICHPRESENCERESULTTEXT`.

src/ui/win32/RichPresenceDialog.cpp
```diff
@@ -0,0 +1,113 @@
...
+RichPresenceDialog::~RichPresenceDialog()
+{
+    DeleteFont(m_hFont);
+}
+
+static void UpdateDisplayString(ra::ui::WindowViewModelBase& vmWindow)
+{
+    auto& vmRichPresence = dynamic_cast<ra::ui::viewmodels::RichPresenceMonitorViewModel&>(vmWindow);
+    vmRichPresence.UpdateDisplayString();
+}
```
Jamiras: `UpdateDisplayString` is called by the timer - it (and all the timer related code) will be moved once a scheduled callback service is available.

src/ui/win32/bindings/WindowBinding.cpp
```diff
@@ -0,0 +1,237 @@
...
+        // capture position relative to main window
+        RECT rcMainWindow;
+        GetWindowRect(g_RAMainWnd, &rcMainWindow);
+        ra::ui::Position oRelativePosition{ rcDialog.left - rcMainWindow.left, rcDialog.top - rcMainWindow.top };
+
+        ra::services::ServiceLocator::GetMutable<ra::services::IConfiguration>().SetWindowPosition(m_sSizeAndPositionKey, oRelativePosition);
+    }
+}
+
+void WindowBinding::OnViewModelStringValueChanged(const StringModelProperty::ChangeArgs& args)
```
Jamiras: this is the `NotifyTarget` callback. The first statement handles the window title, then the labels are managed by an unordered_map.

tests/UI_WindowViewModelBase_Tests.cpp
```diff
@@ -21,30 +24,6 @@ TEST_CLASS(UI_WindowViewModelBase_Tests)
...
-        {
-            vmModalWindow = &vmViewModel;
-            return DialogResult::OK;
-        }
-
-        void Shutdown() override {}
-
-        mutable WindowViewModelBase* vmShownWindow = nullptr;
-        mutable WindowViewModelBase* vmModalWindow = nullptr;
-    };
```
Jamiras: extracted into a separate file, merged with second MockDesktop in VM_MessageBoxViewModel_Tests.cpp

src/ui/win32/Desktop.cpp
```diff
@@ -12,43 +13,69 @@ namespace win32 {
...
+            m_oWorkAreaPosition = { 0, 0 };
+            m_oWorkAreaSize = { 800, 600 };
+        }
+    }
+
+    oUpperLeftCorner = m_oWorkAreaPosition;
+    oSize = m_oWorkAreaSize;
+}
+
+ra::ui::win32::IDialogPresenter* Desktop::GetDialogPresenter(ra::ui::WindowViewModelBase& oViewModel) const
```
SyrianBallaS: This right here looks way too verbose (the `ra::ui::win32::IDialogPresenter*` part). I'd probably pull my hair out if I had to the write the fully qualified name each time when it could've been avoided. It's too bad Visual Studio only has built-in ReSharper-like capabilities for .NET Framework languages.

src/ui/win32/DialogBase.cpp
```diff
@@ -0,0 +1,128 @@
...
+static INT_PTR CALLBACK StaticDialogProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
+{
+    DialogBase* pDialog = reinterpret_cast<DialogBase*>(GetWindowLongPtr(hDlg, DWLP_USER));
+    if (pDialog == nullptr)
+        return DefWindowProc(hDlg, uMsg, wParam, lParam);
+
+    INT_PTR result = pDialog->DialogProc(hDlg, uMsg, wParam, lParam);
+
+    if (uMsg == WM_DESTROY)
+        SetWindowLongPtr(hDlg, DWLP_USER, 0L);
```
SyrianBallaS: In `WM_DESTROY `you're supposed to call `::PostQuitMessage(0)`.
What you have here is for `WM_NCDESTROY`.

src/ui/win32/DialogBase.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    return m_hWnd;
+}
+
+INT_PTR CALLBACK DialogBase::DialogProc(_UNUSED HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
+{
+    switch (uMsg)
+    {
+        case WM_INITDIALOG:
+            OnInitDialog();
```
SyrianBallaS: If you want to get all the values you can get from the kernel there are some message crackers. I'll do em one by one. If you don't want to to them now that's fine. Google "Message Cracker Wizard" if you want an easier time. I was thinking of doing something like this later as well but need to make sure the core is "strong" enough to support the revised UI layer. 

These prototypes will allow you to use them in `HANDLE_MSG`.
`BOOL OnInitDialog(HWND hwnd, HWND hwndFocus, LPARAM lParam);`

src/ui/win32/DialogBase.cpp
```diff
@@ -0,0 +1,128 @@
...
+            return TRUE;
+
+        case WM_DESTROY:
+            OnDestroy();
```
SyrianBallaS: `void OnDestroy(HWND hwnd)`

src/ui/win32/DialogBase.cpp
```diff
@@ -0,0 +1,128 @@
...
+            return 0;
+
+        case WM_COMMAND:
+            return OnCommand(LOWORD(wParam));
```
SyrianBallaS:
```
HANDLE_MSG (hwnd, WM_COMMAND, OnCommand);
void OnCommand(HWND hwnd, int id, HWND hwndCtl, UINT codeNotify)
id is LOWORD(wParam)
```

src/ui/win32/DialogBase.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+        case WM_MOVE:
```
SyrianBallaS:
```
void OnMove(HWND hwnd, int x, int y)
x is LOWORD(lParam) and y is HIWORD(lParam)
```

src/ui/win32/DialogBase.cpp
```diff
@@ -0,0 +1,128 @@
...
+        {
+            ra::ui::Position oPosition{ LOWORD(lParam), HIWORD(lParam) };
+            OnMove(oPosition);
+            return 0;
+        }
+
+        case WM_SIZE:
```
SyrianBallaS: `void OnSize(HWND hwnd, UINT state, int cx, int cy)`
For the rest of these if you don't know which proc argument each one is for look at WindowsX.h.

src/ui/win32/DialogBase.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    switch (nCommand)
+    {
+        case IDOK:
+            m_vmWindow.SetDialogResult(ra::ui::DialogResult::OK);
+            DestroyWindow(m_hWnd);
+            return TRUE;
+
+        case IDCLOSE:
+        case IDCANCEL:
```
SyrianBallaS: VC++ currently doesn't warn on fallthrough, but you should put [[fallthrough]]; after each unhandled case label.

src/ui/win32/DialogBase.hh
```diff
@@ -0,0 +1,95 @@
...
+
+        ShowWindow(m_hWnd, SW_SHOW);
+        return true;
+    }
+
+protected:
+    DialogBase() = delete;
+
+    DialogBase(ra::ui::WindowViewModelBase& vmWindow) noexcept;
+    virtual ~DialogBase();
```
SyrianBallaS: Deleted default constructor is fine.
Mark the one with a param as explicit.
Mark the destructor as `noexcept`, even if it might throw, 
The only ones that it's OK to not have `noexcept `are copy constructor/assignment.

src/ui/win32/DialogBase.hh
```diff
@@ -0,0 +1,95 @@
...
+    /// </summary>
+    /// <param name="oNewPosition">The new size of the client area.</param>
+    virtual void OnSize(ra::ui::Size& oNewSize);
+
+    ra::ui::win32::bindings::WindowBinding m_bindWindow;
+
+    ra::ui::WindowViewModelBase& m_vmWindow;
+
+private:
+    HWND m_hWnd;
```
SyrianBallaS: Even though the default constructor is deleted, you should still initialize this here to `nullptr`. Last time I checked default initializing with a deleted defaulted constructor doesn't cause any side affects.

src/ui/win32/RichPresenceDialog.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+    m_hFont = CreateFont(15, 0, 0, 0, FW_NORMAL, FALSE, FALSE, FALSE, DEFAULT_CHARSET, OUT_DEFAULT_PRECIS, CLIP_DEFAULT_PRECIS, CLEARTYPE_QUALITY, VARIABLE_PITCH, nullptr);
+
+    m_bindWindow.SetInitialPosition(ra::ui::win32::bindings::RelativePosition::After, ra::ui::win32::bindings::RelativePosition::Far, "Rich Presence Monitor");
+    m_bindWindow.BindLabel(IDC_RA_RICHPRESENCERESULTTEXT, ra::ui::viewmodels::RichPresenceMonitorViewModel::DisplayStringProperty);
+}
+
+RichPresenceDialog::~RichPresenceDialog()
+{
+    DeleteFont(m_hFont);
```
SyrianBallaS: Nothing's wrong here, but I just wanted to bring that I'll eventually finalize the RAII-style wrappers for GDI handles (`HDC `will have it's own one) so we don't have to remember to delete these.

I have a few sitting around on my computer, I already made some for `HWND `(app window, client window, modal, modeless), `HANDLE `(just for a few of the kernel objects),` HGLOBAL, HFILE, HINSTANCE/HMODULE, GDI` ones, HDC (was too complicated to put with the other `GDI `ones), `HICON`, `HINTERNET`. I think the last time I was working on it was for `HWND`, haven't touched it in a while.

src/ui/win32/RichPresenceDialog.cpp
```diff
@@ -0,0 +1,113 @@
...
+    StopTimer();
+
+    DialogBase::OnDestroy();
+}
+
+INT_PTR CALLBACK RichPresenceDialog::DialogProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lParam)
+{
+    switch (nMsg)
+    {
+        case WM_TIMER:
```
SyrianBallaS: `WM_TIMER `message cracker for future reference, if there's only timer shouldn't be `TimerProc`?
`void OnTimer(HWND hwnd, UINT id)`

src/ui/win32/RichPresenceDialog.hh
```diff
@@ -0,0 +1,51 @@
...
+#include <memory>
+
+namespace ra {
+namespace ui {
+namespace win32 {
+
+class RichPresenceDialog : public DialogBase
+{
+public:
+    RichPresenceDialog(ra::ui::viewmodels::RichPresenceMonitorViewModel& vmRichPresenceDisplay);
```
SyrianBallaS: needs to be `explicit `and `noexcept`.

src/ui/win32/RichPresenceDialog.hh
```diff
@@ -0,0 +1,51 @@
...
+    ~RichPresenceDialog();
```
SyrianBallaS: needs to be `noexcept`.

src/ui/win32/bindings/WindowBinding.hh
```diff
@@ -0,0 +1,89 @@
...
+    Near,      // Aligned with the target's left or top side
+    Center,    // Centered relative to the target
+    Far,       // Aligned with the target's right or bottom side
+    After,     // To the right of, or below the target
+};
+
+class WindowBinding : protected BindingBase
+{
+public:
+    WindowBinding(WindowViewModelBase& vmWindowViewModel) noexcept
```
SyrianBallaS: Should be `explicit`.

tests/mocks/MockDesktop.hh
```diff
@@ -0,0 +1,55 @@
...
+namespace ra {
+namespace ui {
+namespace mocks {
+
+using DialogHandler = std::function<ra::ui::DialogResult(WindowViewModelBase&)>;
+
+class MockDesktop : public IDesktop
+{
+public:
+    MockDesktop(DialogHandler pHandler)
```
SyrianBallaS: Should be `explicit `and `noexcept`.
Also `pHandler `should either be passed by `const reference` or `rvalue reference.`

src/ui/win32/Desktop.cpp
```diff
@@ -12,43 +13,69 @@ namespace win32 {
...
+        pPresenter->ShowWindow(vmViewModel);
     }
     else
     {
-        assert("!No controller for view model");
+        assert("!No presenter for view model");
     }
 }
 
 ra::ui::DialogResult Desktop::ShowModal(WindowViewModelBase& vmViewModel) const
```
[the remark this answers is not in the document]
Jamiras: Yea, i'm very inconsistent with this.

src/ui/win32/DialogBase.cpp
Jamiras: By the time `WM_NCDESTROY` is called, the `DWLP_USER` value is invalid which causes `pDialog->` to fail. This just says "after the WC_DESTROY message has been handled, stop listening to the remaining events." `DWLP_USER` will be nullptr, so line 26-27 will kick in for the remaining messages.

Jamiras: I thought `PostQuitMessage` only applied to the main window. Doesn't it signal that the application is about to terminate?

src/ui/win32/DialogBase.cpp
Jamiras: I'm torn between exposing everything and exposing only what's needed. For simplicity, I'm going to choose the latter. If we find the need for these parameters later, we can change the signature.

`hWnd` for this, and the other functions will be accessible through `m_hWnd`

src/ui/win32/DialogBase.cpp
Jamiras: `m_hWnd`

src/ui/win32/DialogBase.cpp
Jamiras: yes, converted those to a `Position` structure on line 67

src/ui/win32/DialogBase.cpp
Jamiras: again, prefer to focus on the important parameters until the others become necessary.

src/ui/win32/DialogBase.cpp
Jamiras: I wasn't sure about including state. None of these windows can be minimized/maximized.

src/ui/win32/DialogBase.cpp
Jamiras: I agree when there's code for a case. Multiple case labels immediately adjacent to each other are clearly meant to be handled by the same code.

src/ui/win32/DialogBase.hh
Jamiras: ok

src/ui/win32/DialogBase.hh
Jamiras: ok

src/ui/win32/RichPresenceDialog.cpp
Jamiras: The timer code is temporary. The plan is to move this to the RichPresenceInterpretter using something along these lines:
```
ITimerService.Schedule(std::chrono::minutes(2), UpdateDisplayText);
```
where UpdateDisplayText schedules another callback for itself two minutes later.

src/ui/win32/RichPresenceDialog.hh
Jamiras: done

src/ui/win32/RichPresenceDialog.hh
Jamiras: done

src/ui/win32/bindings/WindowBinding.hh
Jamiras: ok

tests/mocks/MockDesktop.hh
Jamiras: ok

Jamiras:
> This took a long time to go through but that's OK.

And thank you for doing so. I appreciate having a second set of eyes to make sure I'm not doing anything stupid.

src/ui/win32/DialogBase.cpp
SyrianBallaS: Thanks for that. I'm not certain but I think you're right. Dialogs have different rules than other window types.

src/ui/win32/DialogBase.cpp
SyrianBallaS: That sounds like a good approach to me.

PR #134, opened 2018-10-09
Precompiled Header

SyrianBallaS:
This PR mostly has to do w/ using a precompiled header with RA_Integration only. The unit test project does not use it. To save time, I did code analysis using `Microsoft Native Recommended Rules`. Only a small portion of what would show up with `/Wall` was addressed (from the analyzer, the rest will be addressed later. Some code analysis warnings were "unresolvable" so they are temporarily disabled when it happens.

`685 additions and 562 deletions`, review at your leisure.
There will be review comments after this PR is submitted,


src/RA_Dlg_AchievementsReporter.cpp
```diff
@@ -172,25 +171,36 @@ INT_PTR CALLBACK Dlg_AchievementsReporter::AchievementsReporterProc(HWND hDlg, U
...
+                            oss << _T("Bug report is too long, it needs to be ")
+                                << sBugReportInFull.length() - 8192U
+                                << _T(" characters shorter.\n\n"
+                                    "Consider breaking up this bug report and try again!");
+                            stMsg = oss.str();
+                        }
+                        if (::MessageBox(nullptr, stMsg.c_str(), _T("Error"), MB_RETRYCANCEL | MB_ICONERROR) == IDRETRY)
+                            return TRUE; // User wants to try again
+                        FORWARD_WM_CLOSE(hDlg, ::PostMessage); // User doesn't want to retry
+                        return FALSE;
```
SyrianBallaS: I've never actually gotten the bug report to exceed `8192` characters but wanted to see what we could do if that did happen. I just used `100 `instead of `8192`. If it's too long, app will ask if you want to try again and return you to the dialog. If the user doesn't want try again, the dialog is closed going back to the app window. If it succeeded it'll ask you if it was OK like usual, I didn't send the report though.

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -297,25 +299,21 @@ void Dlg_GameLibrary::ThreadedScanProc()
...
             // obtain file size:
             fseek(pf, 0, SEEK_END);
             DWORD nSize = ftell(pf);
             rewind(pf);
 
-            static BYTE* pBuf = nullptr;	//	static (optimisation)
-            if (pBuf == nullptr)
-                pBuf = new BYTE[6 * 1024 * 1024];
+            // May have caused a buffer overrun, this is way to big to be on the stack
+            auto pBuf{ std::make_unique<unsigned char[]>(6 * 1024 * 1024) };
```
SyrianBallaS: Something about a possible buffer overrun.

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -475,7 +473,7 @@ BOOL Dlg_GameLibrary::LaunchSelected()
     const int nSel = ListView_GetSelectionMark(hList);
     if (nSel != -1)
     {
-        TCHAR buffer[1024];
+        TCHAR buffer[1024]{};
```
SyrianBallaS: Arrays changed like this were to address a "string may not be null-terminated" warning.

src/RA_Dlg_GameLibrary.h
```diff
@@ -2,25 +2,23 @@
...
-
 #include "RA_Defs.h"
 
 class GameEntry
 {
 public:
-    GameEntry(const std::string& sTitle, const std::string& sFile, unsigned int nGameID) :
-        m_sTitle(sTitle), m_sFilename(sFile), m_nGameID(nGameID)
-    {
-    }
```
SyrianBallaS: With brace initialization we don't need this. There were other cases like that.

src/RA_Dlg_RomChecksum.cpp
```diff
@@ -27,16 +27,23 @@ INT_PTR CALLBACK RA_Dlg_RomChecksum::RA_Dlg_RomChecksumProc(HWND hDlg, UINT nMsg
...
-                    EmptyClipboard();
-                    SetClipboardData(CF_TEXT, hMem);
-                    CloseClipboard();
-
-                    //MessageBeep( 0xffffffff );
+                    auto hMem{ ::GlobalAlloc(GMEM_MOVEABLE, g_sCurrentROMMD5.length() + 1) };
+                    if (hMem)
+                    {
+                        const auto& lockResult{ ::GlobalLock(hMem) };
+                        if (lockResult)
```
SyrianBallaS: Warning about `hMem `and the return of `GlobalLock `possibly being `nullptr`.

src/RA_Dlg_RomChecksum.cpp
```diff
@@ -27,16 +27,23 @@ INT_PTR CALLBACK RA_Dlg_RomChecksum::RA_Dlg_RomChecksumProc(HWND hDlg, UINT nMsg
...
+                    else
+                        return ra::to_signed(::GetLastError());
+
+                    ::GlobalUnlock(hMem);
+                    ::OpenClipboard(nullptr);
+                    ::EmptyClipboard();
+                    ::SetClipboardData(CF_TEXT, hMem);
+                    ::CloseClipboard();
+
+                    ::GlobalFree(hMem);
```
SyrianBallaS: Leak protection.

src/RA_MemValue.cpp
```diff
@@ -25,12 +26,12 @@ double MemValue::Clause::GetValue() const
...
     if (m_nSecondAddress != 0)
     {
-        unsigned int nSecondVal = g_MemManager.ActiveBankRAMRead(m_nSecondAddress, m_nSecondVarSize);
+        auto nSecondVal = static_cast<std::uint64_t>(g_MemManager.ActiveBankRAMRead(m_nSecondAddress, m_nSecondVarSize));
 
         if (m_bInvertBit && m_nSecondVarSize >= ComparisonVariableSize::Bit_0 && m_nSecondVarSize <= ComparisonVariableSize::Bit_7)
             nSecondVal ^= 1;
 
-        return nRetVal * nSecondVal;
+        return static_cast<double>(nRetVal * nSecondVal);
```
SyrianBallaS: Warning about the result of a `unsigned int` being cast to `double`, made them `unsigned long long `instead and then cast to `double`.

src/RA_RichPresence.cpp
```diff
@@ -221,7 +219,9 @@ void RA_RichPresenceInterpreter::ParseFromString(const char* sRichPresence)
                 newLookup.AddLookupData(nVal, sLabel);
             } while (true);
 
+#ifndef RA_UTEST
             RA_LOG("RP: Adding Lookup %s (%zu items)\n", sLookupName.c_str(), newLookup.NumItems());
+#endif /* !RA_UTEST */
```
SyrianBallaS: Unit test project kept on giving errors about `RA_LOG(...)`

src/RA_httpthread.cpp
```diff
@@ -255,31 +257,39 @@ void RAWeb::SetUserAgentString()
...
-    using fnRtlGetVersion = NTSTATUS(NTAPI*)(PRTL_OSVERSIONINFOEXW lpVersionInformation);
-    auto RtlGetVersion = (fnRtlGetVersion)GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlGetVersion");
-    if (RtlGetVersion)
     {
-        RTL_OSVERSIONINFOEXW osVersion = {};
-        RtlGetVersion(&osVersion);
-        if (osVersion.dwMajorVersion > 0)
+        using fnRtlGetVersion = NTSTATUS(NTAPI*)(PRTL_OSVERSIONINFOEXW);
+        const auto& ntModule{ ::GetModuleHandleW(L"ntdll.dll") };
+        if (ntModule)
```
SyrianBallaS: Warning about the result of `GetModuleHandleW `possibly being `nullptr`.

src/RA_httpthread.cpp
```diff
@@ -255,31 +257,39 @@ void RAWeb::SetUserAgentString()
...
+            auto RtlGetVersion{
+                reinterpret_cast<fnRtlGetVersion>(::GetProcAddress(ntModule, "RtlGetVersion"))
+            };
+            RTL_OSVERSIONINFOEXW osVersion{};
+            RtlGetVersion(&osVersion);
+            if (osVersion.dwMajorVersion > 0)
+            {
+                std::ostringstream oss;
+                oss << "WindowsNT " << osVersion.dwMajorVersion << '.' << osVersion.dwMinorVersion
+                    << '.' << osVersion.dwBuildNumber;
```
SyrianBallaS: Put the build number for good measure, even XP has build numbers. Plus there's numbers so `std::ostringstream` was used.

src/RA_httpthread.cpp
```diff
@@ -255,31 +257,39 @@ void RAWeb::SetUserAgentString()
...
-              RA_INTEGRATION_VERSION_REVISION, RA_INTEGRATION_VERSION_MODIFIED);
-    sUserAgent.append(buffer.release());
-
-    const char* ptr = strchr(RA_INTEGRATION_VERSION_PRODUCT, '-');
-    if (ptr != nullptr)
-        sUserAgent.append(ptr);
+    {
+        const auto szPosFound{ std::string{RA_INTEGRATION_VERSION_PRODUCT}.rfind('-') };
+        if (szPosFound != std::string::npos)
+            sUserAgent.append(std::string{RA_INTEGRATION_VERSION_PRODUCT }, szPosFound);
```
SyrianBallaS: Remembered something about not allocating on the heap for small arrays, this should be slightly faster/safer anyway.

src/base.props
```diff
@@ -25,6 +25,11 @@
       <AdditionalIncludeDirectories>$(RA_IncludePath)</AdditionalIncludeDirectories>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
       <WarningLevel>Level4</WarningLevel>
+      <PrecompiledHeader>Use</PrecompiledHeader>
+      <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
+      <ForcedIncludeFiles>pch.h</ForcedIncludeFiles>
+      <PreprocessorDefinitions>_ARCH=$(PlatformArchitecture);%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
SyrianBallaS: I ended up not using `_ARCH`. I was going to put in in `ra_fwd.h` but decided it would be best saved for later. `/W4` doesn't check for correctness inside macros but `/Wall `does.

I figured it would be better than using `_WIN32`/`_WIN64`.

src/pch.h
```diff
@@ -0,0 +1,75 @@
...
+
+/*
+    Only put files that never change or are rarely changed in this file.
+    An include graph will be provided for some headers that were not explicit
+    added. VC++ is a requirement.
+
+    We don't care about warnings for library files.
+*/
+#pragma warning(push)
+#pragma warning(disable : 4191 4365 4464 4571 4619 4623 4625 4626 4768 4774 5026 5027 5039 5045)
```
SyrianBallaS: `/Wall `is currently off but these are all the warnings the build would have reported with it on.

src/pch.h
```diff
@@ -0,0 +1,75 @@
...
+#include <map>
+#include <memory>
+#include <mutex> // chrono (time.h), functional, thread, utility
+#include <queue> // deque, vector, algorithm
+#include <set>
+#include <sstream> // string
+#include <stack>
+
+#pragma warning(push)
+#pragma warning(disable : 26451 26495)
```
SyrianBallaS: We shouldn't have to deal with code analysis warnings for library files.

src/services/ImageRepository.cpp
```diff
@@ -241,7 +243,7 @@ static HRESULT ConvertBitmapSource(_In_ RECT rcDest, _In_ IWICBitmapSource* pOri
     return hr;
 }
 
-static HRESULT CreateDIBFromBitmapSource(_In_ IWICBitmapSource *pToRenderBitmapSource, _Out_ HBITMAP& hBitmapInOut)
+static HRESULT CreateDIBFromBitmapSource(_In_ IWICBitmapSource *pToRenderBitmapSource, _Inout_ HBITMAP& hBitmapInOut)
```
SyrianBallaS: You can't really `new` a GDI Handle, plus it already says `InOut`,
You can do something similar with `std::unique_ptr `but that'll be saved for later.

src/services/ImageRepository.cpp
```diff
@@ -300,7 +302,7 @@ static HRESULT CreateDIBFromBitmapSource(_In_ IWICBitmapSource *pToRenderBitmapS
         hr = UIntMult(cbStride, nHeight, &cbImage);
 
     // Extract the image into the HBITMAP    
-    if (SUCCEEDED(hr))
+    if (SUCCEEDED(hr) && pvImageBits)
```
SyrianBallaS: Something about `pvImageBits `not being allowed to be `nullptr `in a API call below.

src/services/ImageRepository.h
```diff
@@ -26,15 +30,19 @@ class ImageReference
         ChangeReference(nType, sName);
     }
 
+#pragma warning(push)
+#pragma warning(disable : 26495)
     ImageReference(const ImageReference& source)
         : m_nType(source.m_nType), m_sName(source.m_sName)
```
SyrianBallaS: Because of the comment in this constructor, I disabled the warning.

src/services/SearchResults.h
```diff
@@ -82,57 +86,60 @@ class SearchResults
...
 
-        ~MemBlock()
+        ~MemBlock() noexcept
         {
-            if (nSize > sizeof(m_vBytes))
-                delete m_pBytes;
+            if (m_nSize > sizeof(m_vBytes))
+            {
+                delete[] m_pBytes;
+                m_pBytes = nullptr;
```
SyrianBallaS: `m_pBytes `was allocated using `new[]`. We could make it `std::unique_ptr` later to not have to worry about deallocation here.

src/services/SearchResults.h
```diff
@@ -82,57 +86,60 @@ class SearchResults
...
         {
-            unsigned char m_vBytes[8];
+            unsigned char m_vBytes[8]{};
             unsigned char* m_pBytes;
         };
 
-        unsigned int nAddress; // 4 bytes
-        unsigned int nSize;    // 4 bytes
+        unsigned int m_nAddress{}; // 4 bytes
+        unsigned int m_nSize{};    // 4 bytes
```
SyrianBallaS: I am not certain of what was here before was intentional, but these variables never got initialized in the parameter constructors.

src/services/TextReader.hh
```diff
@@ -2,7 +2,7 @@
 #define RA_SERVICES_TEXTREADER
 #pragma once
 
-#include <string>
+#include <xstring>
```
SyrianBallaS: We just need the using for the string. forward declaring `std::basic_string `is kind of difficult, so this should be good enough.

src/windows_nodefines.h
```diff
@@ -0,0 +1,59 @@
...
+#define NOHELP              1
+#define NOPROFILER          1
+#define NODEFERWINDOWPOS    1
+#define NOMCX               1
+
+/*
+    Problem struct, this file should get included before any windows header
+    file anyway
+*/          
+struct IUnknown;
```
SyrianBallaS: This file should be included before any Windows header in RA_Integration.

src/services/SearchResults.h
```diff
@@ -82,57 +86,60 @@ class SearchResults
...
-        unsigned int GetAddress() const { return nAddress; }
-        unsigned int GetSize() const { return nSize; }
+        unsigned int GetAddress() const { return m_nAddress; }
+        unsigned int GetSize() const { return m_nSize; }
 
     private:
         union // 8 bytes
         {
-            unsigned char m_vBytes[8];
+            unsigned char m_vBytes[8]{};
```
SyrianBallaS: Only one initializer was allowed, this array wouldn't get initialized otherwise. The pointer magically got initialized.

src/RA_Achievement.h
```diff
@@ -148,6 +148,9 @@ class Achievement
 
     unsigned short m_nUpvotes{};
     unsigned short m_nDownvotes{};
+#pragma warning(push)
+#pragma warning(disable : 26495) // "variable" uninitialized
 };
+#pragma warning(pop)
```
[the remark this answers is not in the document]
Jamiras: If I uncomment this, i don't see this error building your branch (with or without /Wall). With /Wall, I get a lot of unreferenced inline functions and some padding warnings.

Please explain what is causing this issue. I'd rather not put this sort of workaround into the code.

src/RA_AchievementOverlay.cpp
```diff
@@ -11,9 +11,11 @@
 #include "services\ILeaderboardManager.hh"
 #include "services\ServiceLocator.hh"
 
+#ifndef PCH_H
 #include <fstream>
 #include <memory>
-#include <ctime>
+#include <ctime>  
+#endif /* !PCH_H */
```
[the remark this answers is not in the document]
Jamiras: Why can't the test project just pull in `pch.h`? It should be able to use it as its own PCH file.

src/RA_AchievementSet.cpp
```diff
@@ -295,7 +292,7 @@ BOOL AchievementSet::SaveToFile()
             Achievement* pAch = &g_pLocalAchievements->GetAchievement(i);
 
             ZeroMemory(sMem, 2048);
-            pAch->CreateMemString();
+            const auto _ = pAch->CreateMemString();
```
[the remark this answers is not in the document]
Jamiras: not seeing this error either. in this case, the line should just be removed. the return value is ignored, and the function is called again on line 300 where the return value is used.  I've already corrected that in PR127.

src/RA_Core.cpp
```diff
@@ -1445,34 +1429,22 @@ bool _ReadBufferFromFile(_Out_ std::string& buffer, const wchar_t* sFile)
...
-    char* pRawFileOut = (char*)malloc((nFileSizeOut + 1) * sizeof(char));
-    ZeroMemory(pRawFileOut, nFileSizeOut + 1);
-
-    fread(pRawFileOut, nFileSizeOut, sizeof(char), pf);
-    fclose(pf);
-
-    return pRawFileOut;
+    auto pRawFileOut = std::make_unique<char[]>(nFileSizeOut + 1);
+    ifile.read(pRawFileOut.get(), nFileSizeOut);
+    return pRawFileOut.release(); // callers needs to manage it
```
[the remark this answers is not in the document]
Jamiras: long term, this function should go away and be handled by `IFileSystem` and `TextReader`

SyrianBallaS:
Like I said in the op, I used the Microsoft native recommend rules. To use it you have to go to ra_integration properties, go to code analysis, check enable code analysis on build, temporarily change the toolset to use v141 w/ windows 8.1/10  SDK. Once I did them I changed the settings back to see if everything still worked.

src/RA_AchievementOverlay.cpp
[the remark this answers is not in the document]
SyrianBallaS: Sorry I misunderstood. I thought you didn't want it to be shared. I can do that  without making it a shared project.

src/RA_Achievement.h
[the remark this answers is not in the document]
SyrianBallaS: That's a code analysis warning. There are warnings in /Wall that address the same thing but with a different code.

src/RA_AchievementSet.cpp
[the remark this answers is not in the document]
SyrianBallaS: If its resolved already, I'll just overwrite with ra's changes when I merge with this branch.

src/RA_Core.cpp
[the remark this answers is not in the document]
SyrianBallaS: Alright, it can go away when the time comes.

src/RA_AchievementOverlay.cpp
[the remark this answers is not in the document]
SyrianBallaS: Did this

src/RA_AchievementSet.cpp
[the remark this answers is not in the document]
SyrianBallaS: Used what was in PR127 instead.

src/RA_AchievementOverlay.cpp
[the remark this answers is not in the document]
SyrianBallaS: Test project has it's own pch but uses `pch.h` and `pch.cpp` from RA_Integration.

SyrianBallaS:
Gonna revert a few commits and open this back up

Jamiras:
I'm still having trouble getting the code analysis to work. I've enabled "Microsoft Native Recommended Rules" for the project, and changed the toolset to "v141" - I don't have an option for  "v141 with windows SDK", which might be related to my other errors, the first being:

> Cannot open include file: 'sdkddkver.h': No such file or directory

Google indicated I needed to have my sdk in the include path, so I did that. There doesn't seem to be a variable for it, so I hardcoded it to "C:\Program Files (x86)\Microsoft SDKs\Windows\v7.1A\Include" - something I'd rather not commit - and then ran into this error:

> 1>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.15.26726\atlmfc\include\atldef.h(69): fatal error C1083: Cannot open include file: 'errno.h': No such file or directory

Google's solution for that was to reinstall VS with the UCRT SDK module - I haven't tried that, and I'm not sure that I want to. If we have to change the toolset to get these warnings, are we actually fixing the warnings correctly for the toolset that we're using?

I did try building the project without the code analysis rules or the v141_xp toolset (but with the new include path) and did notice a significant improvement in compile time, so I reiterate my earlier question: why are you trying to fix the code analysis warnings in a PR that's supposed to just be enabling the PCH functionality?

As a final test, I tried without the additional include, and it did seem to find the sdkddkver.h file when targeting the 141_xp toolset.

SyrianBallaS:
This branch is too difficult to revert, I already made a new one that has pch stuff only w/o code analysis.

PR #136, opened 2018-10-11
replace g_pCurrentData and g_sCurrentROMMD5 with IGameContext

Jamiras:
* g_pCurrentData replaced with IGameContext for game id and title
* g_pCurrentROMMD5 moved into IGameContext
* RichPresencePatch removed from g_pCurrentData and loaded directly into RichPresenceInterpreter
* eliminated ra::GameId alias


src/services/IGameContext.hh
```diff
@@ -0,0 +1,42 @@
...
+#pragma once
+
+#include <string>
+
+namespace ra {
+namespace services {
+
+class IGameContext
+{
+public:
```
SyrianBallaS: You definitely need a pure virtual destructor. 

The rest would be nice to put it in.
I'd normally say save this for later but since the impl class seems like it's supposed to be ID'd.
Delete copying, then put in default moving and default construction as protected.

Normally you're supposed to delete copying anyway with abstract classes to eliminate splicing but the unique constraint that an ID implies seems to warrant this more. To save time you can just copy and paste this. I was thinking of doing this for all our classes later, except the ones that already have them implemented.

```
...
public:
  virtual ~IGameContext() noexcept = default;
  IGameContext(const IGameContext&) noexcept = delete;
  IGameContext& operator=(const IGameContext&) noexcept = delete;
...
protected:
  IGameContext() noexcept = default;
  IGameContext(IGameContext&&) noexcept = default;
  IGameContext& operator=(IGameContext&&) noexcept = default;
...
```

Then you can override the non-deleted ones in implementations if you need non-default behavior.
Some people say you can just make the ID const and it won't change, that is a myth, it's still possible for duplicate entries to be made resulting in exceptions being thrown. Plus it won't be move assignable. Deleting copying will guarantee it won't be duplicated.

src/services/impl/GameContext.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "RA_StringUtils.h"
+
+#include "services\impl\FileTextReader.hh"
+#include "services\impl\FileTextWriter.hh"
+
+namespace ra {
+namespace services {
+namespace impl {
+
+void GameContext::LoadGame(unsigned int nGameId, const std::wstring& sGameTitle)
```
SyrianBallaS: It's up to you but this looks like a constructor.

src/services/impl/GameContext.cpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    if (nGameId == 0)
```
SyrianBallaS: It's up to you, but this what I think about this. An exception handling strategy can come later.

I think an ID of 0 should be considered undefined. If you used this class with iterators and the ID didn't exist, it'll throw an exception. Is it ever allowed for GameID to be 0? If not, you should probably throw an exception and add a new entry when it's caught as it's an exceptional circumstance that only occurs at runtime.

Yes, above I did say this looks like a constructor. Parameterized default constructors can throw, a parameterless/compiler-generated one should not.

src/services/impl/GameContext.hh
```diff
@@ -0,0 +1,32 @@
...
+    unsigned int GameId() const override { return m_nGameId; }       
+    const std::wstring& GameTitle() const override { return m_sGameTitle; }
+
+    const std::string& GameHash() const override { return m_sGameHash; }
+    void SetGameHash(const std::string& sGameHash) override { m_sGameHash = sGameHash; }
+
+private:
+    unsigned int m_nGameId = 0;
+    std::wstring m_sGameTitle;
+    std::string m_sGameHash;
```
SyrianBallaS: Just my opinion, not imperative. There are many ways to about this.

If you're going to have getters and setters for every field you might as well make this a struct. Another thing you could do is till keep it a class and only friend classes/functions that need access.

src/services/IGameContext.hh
```diff
@@ -0,0 +1,42 @@
...
+
+    /// <summary>
+    /// Gets the hash of the currently loaded game.
+    /// </summary>
+    virtual const std::string& GameHash() const = 0;
+    
+    /// <summary>
+    /// Sets the game hash.
+    /// </summary>
+    virtual void SetGameHash(const std::string& sGameHash) = 0;
```
SyrianBallaS: I don't think getters and setters should be part of an interface, if you absolutely need them, they should be in the implementation. One of the main reasons is so you can make them a constant expression if the field used is a literal type (able to be used in constant expressions like primitives).

src/services/IGameContext.hh
Jamiras: As a service, it shouldn't be moveable either. I've added the suggested constructors/destructors with the move functions also marked as deleted.

src/services/IGameContext.hh
Jamiras: The getters are the interface. This object stores data about the currently loaded game. `LoadGame` and `SetGameHash` are the initialization methods. At some point in the future, `LoadGame` will make the API call to fetch the patch data, which will populate the other getters. Similarly, resolving the game hash to an ID might get moved into this service as well.

Another thing that I plan to do in a future PR is add a callback so viewmodel can subscribe to when the game data changes, which is currently handled by multiple locations explicitly informing each window that cares.

As such, this class behaves more like a view model than a service, though it won't actually be bound to any views. I'll eliminate the interface and move the implementation out of impl and register the instance directly (not through the interface). Not having the interface makes it slightly harder to mock, but leaving the backing fields as protected should allow for direct setting via a subclass instead of having to mock the entire Load process.

src/services/impl/GameContext.cpp
Jamiras: For now. In the future, it'll make the server hit to acquire the data and populate the fields from the response.

Jamiras: Also, it has to be able to change when the player switches from one game to another.

src/services/impl/GameContext.cpp
Jamiras: This is to handle the case where a hash doesn't resolve, or the user unloads the game without closing the emulator. It might make more sense if I had moved the server API call into this class, but I wanted to save that for a separate PR.

src/services/impl/GameContext.hh
Jamiras: Not every field, most of the fields should be set by the data returned from the API call. 

Long term, I expect one input (the hash), which will make a server call to get the ID, which will make another server call to get the rest of the data, which will include the achievements - I plan to move g_pCoreAchievements, g_pUnofficialAchievements, and g_pLocalAchievements into this class eventually. I'll probably end up moving the code notes into here too at some point.

The hashing resolution might end up separated due to the logic around popping the dialog when a hash doesn't resolve. I'm still trying to decide the best way to handle that.

SyrianBallaS:
It really depends. I know about Web API controllers as a service which is similar to what you have but the controller isn't an interface. What you have is similar to `ControllerContext` (`Microsoft.AspNetCore.Mvc`). You could just make it as straight up model for state and view-model for input but it'll probably cause some inconsistency in design.

From all your responses, for now, it seems better to keep it how it was but do the constructor thing. It can change later when appropriate.

Jamiras:
I've decided to make it a singleton model without an interface (`ra::data::GameContext`). It feels cleaner from an implementation perspective, but does feel a little weird to access via `ServiceLocator`.

SyrianBallaS:
I think it's fine, but since there isn't any inheritance anymore your should probably make the destructor in `GameContext `non-virtual.

It might seem weird but it's done all the time. What you're trying to do is similar to Autofac, and you can get transients, singletons and services even though its supposed to be a service container..

Jamiras:
Eventually, a `MockGameContext` will be created, which will subclass and expose setters for the remaining getters. I'm going to leave the virtual destructor.

PR #137, opened 2018-10-11
Precompiled header

SyrianBallaS:
Basically the same as #134 but got rid of all the controversial code analysis stuff and will probably do that way later. Left very few in, mostly where stuff could go terribly wrong if not checked.
`394 additions and 259 deletions.`


src/RA_AchievementSet.cpp
```diff
@@ -353,8 +350,7 @@ bool AchievementSet::LoadFromFile(ra::GameID nGameID)
         g_pCurrentGameData->SetGameID(nGameID);
                
         rapidjson::Document doc;
-        LoadDocument(doc, *pData.get());
-        if (doc.HasParseError())
+        if (LoadDocument(doc, *pData.get()) == doc.HasParseError())
```
Jamiras: This logic is incorrect. `LoadDocument` may return `false` without setting `doc.HasParseError`, which would bypass the assert/return. If it makes it all the way to the end, `LoadDocument` return `!doch.HasParseError`, so you can eliminate that and just say:
```
if (!LoadDocument(doc, *pData.get()))
```

src/RA_AchievementOverlay.cpp
```diff
@@ -11,9 +11,11 @@
 #include "services\ILeaderboardManager.hh"
 #include "services\ServiceLocator.hh"
 
+#ifndef PCH_H
 #include <fstream>
 #include <memory>
-#include <ctime>
+#include <ctime>  
+#endif /* !PCH_H */
```
Jamiras: Are these chunks still necessary if `pch.h` is being included by both projects?

src/RA_CodeNotes.cpp
```diff
@@ -23,8 +21,7 @@ size_t CodeNotes::Load(ra::GameID nID)
         return 0U;
 
     rapidjson::Document doc;
-    LoadDocument(doc, *pData.get());
-    if (doc.HasParseError())
+    if (LoadDocument(doc, *pData.get()) == doc.HasParseError())
```
Jamiras: Same comment as above.

src/RA_Dlg_RomChecksum.cpp
```diff
@@ -27,16 +27,23 @@ INT_PTR CALLBACK RA_Dlg_RomChecksum::RA_Dlg_RomChecksumProc(HWND hDlg, UINT nMsg
...
-                    //MessageBeep( 0xffffffff );
+                    auto hMem{ ::GlobalAlloc(GMEM_MOVEABLE, g_sCurrentROMMD5.length() + 1) };
+                    if (hMem)
+                    {
+                        const auto& lockResult{ ::GlobalLock(hMem) };
+                        if (lockResult)
+                            std::memcpy(lockResult, g_sCurrentROMMD5.c_str(), g_sCurrentROMMD5.length() + 1);
+                    }
+                    else
+                        return ra::to_signed(::GetLastError());
```
Jamiras: I believe WM_COMMAND should still return TRUE even on error, to indicate that we _tried_ to handle it.

src/RA_Integration.vcxproj
```diff
@@ -158,7 +168,7 @@
     <ProjectName>RA_Integration</ProjectName>
     <ProjectGuid>{825CB292-F069-4B27-82CF-3417746ACDF3}</ProjectGuid>
     <RootNamespace>RA_Integration</RootNamespace>
-    <WindowsTargetPlatformVersion>7.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
```
Jamiras: Why did this change?

src/RA_RichPresence.cpp
```diff
@@ -241,7 +239,9 @@ bool RA_RichPresenceInterpreter::Load(ra::services::TextReader& pReader)
                 newLookup.AddLookupData(nVal, sLabel);
             } while (true);
 
+#ifndef RA_UTEST
             RA_LOG("RP: Adding Lookup %s (%zu items)\n", sLookupName.c_str(), newLookup.NumItems());
+#endif /* !RA_UTEST */
```
Jamiras: `RA_Log.h` already undefines these for `RA_UTEST`. You shouldn't have to do this here, or we'd have to do it for all the files that get pulled into the Tests project.

src/RA_httpthread.cpp
```diff
@@ -256,9 +258,10 @@ void RAWeb::SetUserAgentString()
     #define NTSTATUS long
 #endif
     using fnRtlGetVersion = NTSTATUS(NTAPI*)(PRTL_OSVERSIONINFOEXW lpVersionInformation);
-    auto RtlGetVersion = (fnRtlGetVersion)GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlGetVersion");
-    if (RtlGetVersion)
+    const auto& ntModule{ ::GetModuleHandleW(L"ntdll.dll") };
+    if (ntModule)
     {
+        auto RtlGetVersion = (fnRtlGetVersion)GetProcAddress(ntModule, "RtlGetVersion");
```
Jamiras: It's probably more important to null check the function than the dll. The dll probably exists on all machines.

src/pch.h
```diff
@@ -0,0 +1,73 @@
...
+#include <map>
+#include <memory>
+#include <mutex> // chrono (time.h), functional, thread, utility
+#include <queue> // deque, vector, algorithm
+#include <set>
+#include <sstream> // string
+#include <stack>
+
+/* RapidJSON Stuff */
+#if (RA_EXPORTS || RA_UTEST)
```
Jamiras: this seems unnecessary. `RA_EXPORTS` is the dll, `RA_UTEST` is the Tests project. The only way either of these wouldn't be defined is in the emulators, which shouldn't be using the `pch.h` file.

src/services/TextWriter.hh
```diff
@@ -2,7 +2,7 @@
 #define RA_SERVICES_TEXTWRITER
 #pragma once
 
-#include <string>
+#include <xstring>
```
Jamiras: xstring is a microsoft specific header, please use the standard string header so we're ready when we eventually get to supporting other platforms.

src/RA_AchievementSet.cpp
SyrianBallaS: I just saw this, I got you.

src/RA_AchievementOverlay.cpp
SyrianBallaS: Well not really, I was afraid if a different project in the future didn't have pch we wouldn't have to adjust for it. I'll get rid of them.

src/RA_Integration.vcxproj
SyrianBallaS: That is really weird, it says 7.0 on my computer.

src/pch.h
SyrianBallaS: good point

src/services/TextWriter.hh
SyrianBallaS: I got you. Non standard stuff should be avoided for easier transition.

src/RA_Integration.vcxproj
SyrianBallaS: Yeah no problem on the document, I'll probably do another configuration later.

SyrianBallaS:
There's a few things have to do with this, closing it for now.

RA_Integration.sln
```diff
@@ -14,17 +14,27 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "RA_Integration.Tests", "tes
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Analysis|Win32 = Analysis|Win32
+		AnalysisUnicode|Win32 = AnalysisUnicode|Win32
```
SyrianBallaS: Put in new Analysis Configuration, Test projects is Unicode only. They are both debug, doesn't make sense to use it for release mode.

src/RA_Integration.vcxproj
```diff
@@ -179,6 +191,7 @@
     <ProjectName>RA_Integration</ProjectName>
     <ProjectGuid>{825CB292-F069-4B27-82CF-3417746ACDF3}</ProjectGuid>
     <RootNamespace>RA_Integration</RootNamespace>
+    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
```
SyrianBallaS: This is the lowest version of the Windows SDK I have in `Windows Kits`. Someone else might need a different version. Recommend using 10.0.15063+. It can possibly work with 8.1 or 7.1 but haven't tested it.

src/RA_Integration.vcxproj
```diff
@@ -210,11 +223,21 @@
     <CharacterSet>MultiByte</CharacterSet>
     <PlatformToolset>v141_xp</PlatformToolset>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Analysis|Win32'" Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <CharacterSet>MultiByte</CharacterSet>
+    <PlatformToolset>v141</PlatformToolset>
```
SyrianBallaS: Will not work targeting the XP compatible tool-set as the SAL annotations and runtime libraries between XP and 7 are incompatible.

src/RA_httpthread.cpp
```diff
@@ -247,32 +247,55 @@ void RAWeb::SetUserAgentString()
...
+            return;
         {
-            sUserAgent.append("WindowsNT ");
-            sUserAgent.append(std::to_string(osVersion.dwMajorVersion));
-            sUserAgent.append(".");
-            sUserAgent.append(std::to_string(osVersion.dwMinorVersion));
+            std::ostringstream oss;
+            oss << "WindowsNT " << osVersion.dwMajorVersion << '.' << osVersion.dwMinorVersion
+                << '.' << osVersion.dwBuildNumber;
+            sUserAgent.append(oss.str());
```
SyrianBallaS: Applied previous feedback but expanded on it. Build numbers can help us know what version of Windows someone is using more accurately.

src/md5.c
```diff
@@ -54,6 +54,10 @@
 #include "md5.h"
 #include <string.h>
 
+/* Really didn't want to have to do this but can't find a better way to do it - SyrianBallaS */
+#include <codeanalysis\warnings.h>
+#pragma warning(disable : ALL_CPPCORECHECK_WARNINGS)
```
SyrianBallaS: Didn't want to do this but this is included as a project file, there's no other way around as we don't use a .lib/.dll for it.

src/pch.h
```diff
@@ -52,6 +52,11 @@
 #include <stack>
 #include <unordered_map>
 
+#pragma warning(push)
+// Look at this file if you want to know what they are and what they mean
+#include "pch_cppcorecheck_suppressions.h"
+#include "pch_microsoft_suppressions.h"
```
SyrianBallaS: These have the suppression in greater detail in case someone doesn't know what they are.

src/RA_Dlg_RichPresence.h
```diff
@@ -0,0 +1,41 @@
+#ifndef RA_DLG_RICHPRESENCE_H
+#define RA_DLG_RICHPRESENCE_H
+#pragma once
```
Jamiras: This file should no longer exist. I'm guessing it got readded to resolve a conflict.

src/RA_Dlg_RomChecksum.cpp
```diff
@@ -27,16 +27,29 @@ INT_PTR CALLBACK RA_Dlg_RomChecksum::RA_Dlg_RomChecksumProc(HWND hDlg, UINT nMsg
                 case IDC_RA_COPYCHECKSUMCLIPBOARD:
                 {
                     //	Allocate memory to be managed by the clipboard
-                    HGLOBAL hMem = GlobalAlloc(GMEM_MOVEABLE, g_sCurrentROMMD5.length() + 1);
-                    memcpy(GlobalLock(hMem), g_sCurrentROMMD5.c_str(), g_sCurrentROMMD5.length() + 1);
-                    GlobalUnlock(hMem);
+                    // wrapped just in case it goes out of scope early
+                    using GlobalOwner = std::unique_ptr<std::remove_pointer_t<HGLOBAL>, decltype(&::GlobalFree)>;
```
Jamiras: This will cause the memory to get released when leaving the function, even if we hand it off to the clipboard. You probably need to change line 49 to `hMem.release()`.

src/RA_httpthread.cpp
```diff
@@ -252,31 +247,74 @@ void RAWeb::SetUserAgentString()
 #ifndef NTSTATUS
     #define NTSTATUS long
 #endif
-    using fnRtlGetVersion = NTSTATUS(NTAPI*)(PRTL_OSVERSIONINFOEXW lpVersionInformation);
-    auto RtlGetVersion = (fnRtlGetVersion)GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlGetVersion");
-    if (RtlGetVersion)
     {
-        RTL_OSVERSIONINFOEXW osVersion = {};
+        // needs to be XP or higher no point in continuing
```
Jamiras: We still need to construct a User Agent regardless of the Windows version. 

What was wrong with the code before? On XP (or before XP), the GetModuleHandle or the GetProcAddress would fail and we'd simply skip over the NT version section of the User Agent.

src/RA_httpthread.cpp
Jamiras: This was modeled after IE11 and Chrome user agent strings, neither of which include the build number.

Jamiras: Why use `osstringstream` to create an intermediate string instead of just appending directly to the `sUserAgent` string as was previously done before?

src/RA_httpthread.cpp
```diff
@@ -252,31 +247,74 @@ void RAWeb::SetUserAgentString()
...
     }
 
     sUserAgent.append(") Integration/");
+    {
+        std::ostringstream oss;
+        // Compile-time string concatenation will need some user-defined stuff/work.
+        oss << "WindowsNT " << RA_INTEGRATION_VERSION_MAJOR << '.' << RA_INTEGRATION_VERSION_MINOR
+            << '.' << RA_INTEGRATION_VERSION_REVISION << '.' << RA_INTEGRATION_VERSION_MODIFIED;
+        sUserAgent.append(oss.str());
+    }
```
Jamiras: Instead of using `sprintf` once, you're relying on the `<<` operator to call `std::to_string` four times and using an `ostringstream` to build an intermediate string. How is this better?

src/RA_httpthread.cpp
```diff
@@ -252,31 +247,74 @@ void RAWeb::SetUserAgentString()
...
 
-    auto buffer{ std::make_unique<char[]>(64U) };
-    sprintf_s(buffer.get(), 64U, "%d.%d.%d.%d", RA_INTEGRATION_VERSION_MAJOR, RA_INTEGRATION_VERSION_MINOR,
-              RA_INTEGRATION_VERSION_REVISION, RA_INTEGRATION_VERSION_MODIFIED);
-    sUserAgent.append(buffer.release());
```
Jamiras: This is a memory leak. Luckily the method is only called once.
https://github.com/RetroAchievements/RAIntegration/commit/59acb109e85b1f63cdcd5803a503fac8d699effd#diff-5f269daa1d43eae5f0106751651f3a31R274

src/RA_httpthread.cpp
```diff
@@ -252,31 +247,74 @@ void RAWeb::SetUserAgentString()
...
+        // compile-time finding, intellisense will tell you the number is already
+        // Only put C-Strings in string_view, they must also live longer than the string_view
+        _CONSTANT_LOC posFound{ std::string_view{ RA_INTEGRATION_VERSION_PRODUCT }.find('-') };
+        if constexpr (posFound != std::string_view::npos)
+        {
+            // appending still done at runtime but the values for it are retrieved at compile-time
+            // needs to use the const T& overload
+            std::string_view sAppend{ RA_INTEGRATION_VERSION_PRODUCT };
+            sUserAgent.append(sAppend, posFound);
+        }
```
Jamiras: I like the idea of resolving this at compile time, but how is it related to the PCH?

src/RA_Dlg_RichPresence.h
SyrianBallaS: I'll get rid of it

src/RA_Dlg_RomChecksum.cpp
SyrianBallaS: Ok

src/RA_httpthread.cpp
SyrianBallaS: Nothing was wrong but if we don't care about the windows version here it can go back

src/RA_httpthread.cpp
SyrianBallaS: I need to check again but I thought the macros were numbers except the last one. It was for speed. It go back but using a static array instead

src/RA_httpthread.cpp
SyrianBallaS: It's not, I think too far ahead sometimes. I'll try not to do it again.

src/RA_httpthread.cpp
SyrianBallaS: The build number will go away but I still think `ostringstream `should be used.

Basically what I've been repeating several times, there is more sources than this one
https://www.thecodingdelight.com/string-cplusplus/

Some Key Points if it's too long to read, I do research all the time. It's really important and feel like it's been quite understated on how important it really is.
- `Do not concatenate using `+ `operator. Use `+=` or `append()`.`
- Only use `stringstream `when you need to convert `other data types` `(E.g. int)` to `strings`
- Replace all instances of `+ `with `+=` where applicable.

src/RA_httpthread.cpp
SyrianBallaS: I don't know what code you're talking about but `operator<<` does not call `to_string` it uses `ostream`. So I still think this is better. I will run some benchmarks just to be sure. Theoretically this should be faster.

SyrianBallaS:
My benchmark results are copied from a different file. Tested the narrowing to make sure it always it worked. Made sure there were no leaks in the areas discussed.
```
All these benchmarks are the average times over 1 million executions
None of these include object creation, just the operation to write the string and/or append
ostringstream vs append (not all numbers)
-----------------------------------------------
ostringstream: 137.830222 microseconds
append: 28.534731 microseconds

Conclusion: Use append instead of ostringstream even if there are numbers.

OK, we know append is faster ostringstream let's compare append to sprintf_s
append vs sprintf_s (with static array and std::string respectively) (not all numbers)
------------------------------------------------------------------------------------------
sprintf_s: 3.604900 microseconds
sprintf_s with std::string: 3.230776 microseconds // even faster than static array
append: 62.642287 microseconds

Conclusion: Use sprintf_s here with std::string instead for now, and eventually use a checked version, like the ra::stprintf_s wrapper if it gets accepted.
```

src/RA_httpthread.cpp
SyrianBallaS: And theory was disproved, implementations are not made equal I guess. It was faster last time I bench-marked using GCC (for research not this project).

SyrianBallaS:
Something strange happened, closing this.

SyrianBallaS:
Problem solved, reopening.

src/RA_LeaderboardPopup.cpp
```diff
@@ -290,14 +290,20 @@ void LeaderboardPopup::Render(HDC hDC, RECT& rcDest)
...
-                    ra::tstring str;
+                    
                     {
-                        std::basic_ostringstream<TCHAR> oss;
-                        oss << " " << lbInfo.m_nRank << " " << lbInfo.m_sUsername << " ";
-                        str = oss.str();
+                        std::string str{ "" };
+                        const auto nCharsWritten
+                        {
+                            std::snprintf(str.data(), BUFSIZ, " %u %s ", lbInfo.m_nRank,
```
Jamiras: You need to `str.reserve(BUFSIZ)`. The initial capacity of a `std::string` is much smaller than `BUFSIZ`.

Jamiras: Pretty sure you also need to do a `str.resize(nCharsWritten)` afterwards to update the length.

src/RA_httpthread.cpp
```diff
@@ -252,31 +247,51 @@ void RAWeb::SetUserAgentString()
...
-    using fnRtlGetVersion = NTSTATUS(NTAPI*)(PRTL_OSVERSIONINFOEXW lpVersionInformation);
-    auto RtlGetVersion = (fnRtlGetVersion)GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlGetVersion");
-    if (RtlGetVersion)
     {
-        RTL_OSVERSIONINFOEXW osVersion = {};
+        using fnRtlGetVersion = NTSTATUS(NTAPI*)(PRTL_OSVERSIONINFOEXW lpVersionInformation);
+
+        const auto ntModule{ ::GetModuleHandleW(L"ntdll.dll") };
+        if (!ntModule)
+            return;
```
Jamiras: Need to not bail from this function if these fail. The User-Agent string should still be created, even if the exact NT version cannot be determined.

src/RA_httpthread.cpp
```diff
@@ -252,31 +247,51 @@ void RAWeb::SetUserAgentString()
...
+                sUserAgent.append(str.c_str());
         }
     }
 
     sUserAgent.append(") Integration/");
+    {
+        std::string str{ "" };
+        const auto needed
+        {
+            std::snprintf(str.data(), BUFSIZ, "%d.%d.%d.%d", RA_INTEGRATION_VERSION_MAJOR,
```
Jamiras: Same comment(s) about `sprintf`ing into a `std::string`. You need to make sure the string actually has the capacity specified as the buffer size, and need to update the length of the string afterwards.

src/services/ImageRepository.cpp
```diff
@@ -262,7 +258,7 @@ static HRESULT ConvertBitmapSource(_In_ RECT rcDest, _In_ IWICBitmapSource* pOri
     return hr;
 }
 
-static HRESULT CreateDIBFromBitmapSource(_In_ IWICBitmapSource *pToRenderBitmapSource, _Out_ HBITMAP& hBitmapInOut)
+static HRESULT CreateDIBFromBitmapSource(_In_ IWICBitmapSource *pToRenderBitmapSource, _Inout_ HBITMAP& hBitmapInOut)
```
Jamiras: This is only called in a situation where hBitmapInOut is truly just an out parameter. The name should be fixed, and the "DeleteBitmap" check on line 299 is unnecessary.

I'm fine with leaving this for now.

tests/RA_Integration.Tests.vcxproj
```diff
@@ -43,32 +54,55 @@
...
       <AdditionalIncludeDirectories>$(RA_IncludePath);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>WIN32;RA_UTEST;_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <UseFullPaths>true</UseFullPaths>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
+      <ShowIncludes>false</ShowIncludes>
+    </ClCompile>
+    <Link>
+      <SubSystem>Windows</SubSystem>
+      <AdditionalLibraryDirectories>$(VCInstallDir)UnitTest\lib;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalDependencies>windowscodecs.lib;winmm.lib;Winhttp.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
Jamiras: Where did all these additional dependencies come from? If they're really necessary, they should be added to all builds, not just Debug.

SyrianBallaS: All I did was copy the configuration from the Debug version and adjusted it a little bit for Analysis Version. I'll look into more.

src/RA_LeaderboardPopup.cpp
SyrianBallaS: Sure.

src/RA_httpthread.cpp
SyrianBallaS: OK.

src/RA_LeaderboardPopup.cpp
SyrianBallaS: The resize thing didn't work, but doing a "self" assign did.
`str = str.c_str();`

Jamiras: Here's an example of using snprintf with a string buffer, in a function called "StringPrintf": https://github.com/RetroAchievements/RAIntegration/blob/master/src/RA_StringUtils.cpp#L129

Maybe you should just call that?

SyrianBallaS: Seems like a good approach

SyrianBallaS: I'll still use your suggestion but note it is slower than before (as well as not safe with the C varargs).
average over 1 million times.
`with ra::StringPrintf: 13.280193 microseconds`
instead of the 3.xxxxx microseconds.
It's not that big a deal, 1 millisecond = 1000 microseconds.

PR #138, opened 2018-10-13
wsprintf to ra::stprintf_s (new, wrapper-ish)

SyrianBallaS:
OK, so I noticed some `wsprintf`. Standard C I/O streams are bad enough, but `wsprintf `is so dangerous even Microsoft doesn't encourage using it. It doesn't even support proper encoding or some format specs.

I could've just replaced the `wsprintf`s with `_stprintf_s `but that is still not good enough. It doesn't have the bounds/type checking of the C11 versions. So I designed a wrapper as a parameter pack to handle multibyte or wide-character strings, even if a different character set was specified. 

The functions wrapped where the C++11 versions (it's C equivalent is C99) (not those so called "safe" functions) to somewhat imitate the checks of the C11 versions based off of their specification. It does possibly explicitly throw exceptions at runtime in release mode, but with the stuff I've changed it never did. The compiler generated messages are usually too cryptic, "Segmentation Fault" or "Stack/Heap Corruption" but tells you nothing on what you should do. The exceptions thrown will.

I can actually see the 32-bit view now without the stack getting corrupted like before.


src/ra_utility.h
```diff
@@ -172,6 +177,146 @@ _NODISCARD inline auto __cdecl tstrlen(_In_z_ LPCTSTR _Str) noexcept
...
+            throw std::out_of_range{ "Stack or heap corruption detected, increase bufsz!" };
+        else if (ret < 0)
+            throw std::runtime_error{ "An encoding error has occurred, try using Unicode!" };
+#endif /* _DEBUG */
+
+        return ret;
+    }
+    if constexpr (std::is_same_v<CharT, wchar_t>)
+    {
+        const auto ret{ std::swprintf(buffer, bufsz, format, std::forward<Args>(args)...) };
```
Jamiras: why go through all this effort if it just boils down to changing `wsprintf` to `swprintf`?

src/ra_utility.h
```diff
@@ -172,6 +177,146 @@ _NODISCARD inline auto __cdecl tstrlen(_In_z_ LPCTSTR _Str) noexcept
...
+    // MSVC doesn't seem to have constraint handlers like C11 we could throw an exception...
+    if constexpr (std::is_same_v<CharT, char>)
+    {
+        const auto ret{ std::snprintf(buffer, bufsz, format, std::forward<Args>(args)...) };
+#if _DEBUG
+        assert(std::strlen(buffer) < bufsz);
+        assert(ret > 0);
+#else
+        if (std::strlen(buffer) > bufsz)
+            throw std::out_of_range{ "Stack or heap corruption detected, increase bufsz!" };
```
Jamiras: I have two issues with this:
1) calling strlen on every resulting string is unnecessary overhead.
2) if a problem is found, throwing an exception will kill the app. I don't believe we have sufficient handlers to even capture/report the exception in a release build.

src/ra_utility.h
```diff
@@ -172,6 +177,146 @@ _NODISCARD inline auto __cdecl tstrlen(_In_z_ LPCTSTR _Str) noexcept
...
+        const auto num_specs{ std::count(test.begin(), test.end(), '%') };
+        assert(num_specs == sizeof...(args));
+    }
+    else if constexpr(std::is_same_v<CharT, wchar_t>)
+    {
+        assert(std::wstring{ format }.find(L"%n") == std::wstring::npos);
+        std::wstring test{ format };
+        const auto num_specs{ std::count(test.begin(), test.end(), L'%') };
+        assert(num_specs == sizeof...(args));
+    }
```
Jamiras: It's generally bad practice to have different behavior in release vs debug builds. if one gets modified and the other doesn't, it's hard to diagnose the difference. Since we don't handle exceptions very well, I recommend limiting the problems to the asserts. However, as these are catastrophic errors, the exceptions would be justified.

SyrianBallaS: For now I'll just leave the asserts

src/ra_utility.h
SyrianBallaS: It's for flexibility so it doesn't strictly have to be TCHAR since there's explicit use of unicode APIs even in a mbcs build. It can be used as a replacement for sprintf in general but didn't want to change all of them at this time.

src/ra_utility.h
SyrianBallaS: I got you.

src/ra_utility.h
```diff
@@ -172,6 +176,77 @@ _NODISCARD inline auto __cdecl tstrlen(_In_z_ LPCTSTR _Str) noexcept
...
+/// </param>
+/// <param name="...args">
+///   Forwarded arguments specifying data to print. If the number of arguments is more than the
+///   number of format specifiers present, the operation is undefined.
+/// </param>
+/// <returns>
+///   The number of characters written to <paramref name="buffer" /> excluding the null character.
+/// </returns>
+template<typename CharT = TCHAR, typename ...Args, typename = std::enable_if_t<is_char_v<CharT>>>
+_Success_(return >= 0) /*_NODISCARD*/ inline auto __cdecl
```
SyrianBallaS: Wasn't sure if I should've made it required to check the return value or not.

src/ra_utility.h
```diff
@@ -172,6 +176,77 @@ _NODISCARD inline auto __cdecl tstrlen(_In_z_ LPCTSTR _Str) noexcept
...
+stprintf_s(_Inout_updates_bytes_(bufsz)  CharT*       const __restrict buffer,
+           _In_range_(1, RSIZE_MAX)      rsize_t      const            bufsz,
+           _In_z_ _Printf_format_string_ const CharT* const __restrict format,
+           _In_                          Args&&...                     args)
```
SyrianBallaS: These pre/post conditions are only checked during code analysis, this is based off the the specification for C11's `sprintf_s`. MSVC version is not this strict.

src/ra_utility.h
```diff
@@ -172,6 +176,77 @@ _NODISCARD inline auto __cdecl tstrlen(_In_z_ LPCTSTR _Str) noexcept
...
+{
+    assert((buffer != nullptr) && (format != nullptr));
+    assert((bufsz > 0) && (bufsz <= RSIZE_MAX));
+    
+    if constexpr(std::is_same_v<CharT, char>)
+    {
+        [[maybe_unused]] std::string_view test{ format };
+        assert(test.find("%n") == std::string::npos);
+        [[maybe_unused]] const auto num_specs{ std::count(test.begin(), test.end(), '%') };
+        assert(num_specs == sizeof...(args));
```
SyrianBallaS: variables are sometimes unused.
Everything in here is a constant expression except `std::count` but should still iterate fast from the `string_view`.

src/ra_utility.h
```diff
@@ -172,6 +176,77 @@ _NODISCARD inline auto __cdecl tstrlen(_In_z_ LPCTSTR _Str) noexcept
...
+        [[maybe_unused]] const auto num_specs{ std::count(test.begin(), test.end(), L'%') };
+        assert(num_specs == sizeof...(args));
+    }
+    
+    // MSVC doesn't seem to have constraint handlers like C11
+    if constexpr (std::is_same_v<CharT, char>)
+    {
+        const auto ret{ std::snprintf(buffer, bufsz, format, std::forward<Args>(args)...) };
+        assert(std::char_traits<char>::length(buffer) < bufsz);
+        assert(ret > 0);
```
SyrianBallaS: `std::char_traits<char_type>::length` is a constant expression, it shouldn't give any overhead.

SyrianBallaS:
We probably don't need this right now, but will keep it locally. Figured out another way for the task at hand in another PR.

PR #139, opened 2018-10-13
rotate log file if larger than 1MB

Jamiras:


src/services/IFileSystem.hh
```diff
@@ -35,6 +37,25 @@ public:
     /// <returns><c>true</c> if successful, <c>false</c> if not.</returns>
     virtual bool CreateDirectory(const std::wstring& sDirectory) const = 0;
 
+    /// <summary>
+    /// Gets the size of the file (in bytes).
+    /// </summary>
+    /// <remarks>Returns -1 if the file does not exist.</remarks>
+    virtual int64_t GetFileSize(const std::wstring& sPath) const = 0;
```
SyrianBallaS: What you have is fine but just wanted to comment on this.

I have a function similar to this in `ra_utility.h` that attempts to get the file size at compile-time, should we just keep that one as an alternative or get rid of it? At worst it's constant-time, as the value is retrieved on construction. It also supports any character type specified the `ra::is_char` `type_trait `constraint.

It looks like this:
```
template<typename CharT, class = std::enable_if_t<is_char_v<CharT>>> _NODISCARD _CONSTANT_FN
filesize(_In_ const std::basic_string<CharT>& filename) noexcept
{
    using file_type = std::basic_fstream<CharT>;
    file_type file{ filename, std::ios::in | std::ios::ate | std::ios::binary };
    return file.tellg();
} // end function filesize
```
`file.tellg()` will also return -1 on error. Auto deduced return type is `std::basic_fstream<CharT>::pos_type` which is `std::streamoff` which is implementation defined but usually `long long.`

There's also an` rvalue reference` version if the string is a temporary or no longer needs to be used. I guess the main difference is that mine is assumed not to fail as only RA filenames should be passed to it.

`tellg `can throw `std::ios_base::failur`e but is usually handled internally.

src/services/impl/FileLogger.hh
```diff
@@ -21,7 +21,18 @@ class FileLogger : public ra::services::ILogger
 public:
     explicit FileLogger(const ra::services::IFileSystem& pFileSystem) noexcept
     {
-        m_pWriter = pFileSystem.AppendTextFile(pFileSystem.BaseDirectory() + L"RACache\\RALog.txt");
+        std::wstring sLogFilePath = pFileSystem.BaseDirectory() + L"RACache\\RALog.txt";
+
+        // if the file is over 1MB, rename it and start a new one
+        int64_t nLogSize = pFileSystem.GetFileSize(sLogFilePath);
```
SyrianBallaS: It's up to you but this variable can be destroyed earlier, either by scope or selection statement.

src/services/impl/FileLogger.hh
```diff
@@ -21,7 +21,18 @@ class FileLogger : public ra::services::ILogger
 public:
     explicit FileLogger(const ra::services::IFileSystem& pFileSystem) noexcept
     {
-        m_pWriter = pFileSystem.AppendTextFile(pFileSystem.BaseDirectory() + L"RACache\\RALog.txt");
+        std::wstring sLogFilePath = pFileSystem.BaseDirectory() + L"RACache\\RALog.txt";
```
SyrianBallaS: Speed could be optimized more using `+=` or `append`. `L"RACache\\RALog.txt"` seems it can be a `constexpr `variable.
I was thinking of making something that attempts compile-time string concatenation. As far as I'm aware, none of the standard library does this.

tests/mocks/MockFileSystem.hh
```diff
@@ -53,6 +59,55 @@ public:
...
+        m_mFileSizes.insert_or_assign(sPath, nFileSize);
+    }
+
+    int64_t GetFileSize(const std::wstring& sPath) const override
+    {
+        auto pIter = m_mFileSizes.find(sPath);
+        if (pIter != m_mFileSizes.end())
+            return pIter->second;
+
+        auto pIter2 = m_mFileContents.find(sPath);
```
SyrianBallaS: It's up to you but `pIter `and `pIter2 `can be destroyed earlier if the conditions failed.

tests/mocks/MockFileSystem.hh
```diff
@@ -53,6 +59,55 @@ public:
...
+    bool MoveFile(const std::wstring& sOldPath, const std::wstring& sNewPath) const override
+    {
+        auto hNode = m_mFileContents.extract(sOldPath);
+        if (hNode.empty())
+            return false;
+
+        hNode.key() = sNewPath;
+        m_mFileContents.insert(std::move(hNode));
+
+        auto hNode2 = m_mFileSizes.extract(sOldPath);
```
SyrianBallaS: This seems like it can destroyed early if the operation failed.

tests/services/FileLogger_Tests.cpp
```diff
@@ -20,6 +20,7 @@ TEST_CLASS(FileLogger_Tests)
 {
 private:
     const std::wstring mockLogFileName = L".\\RACache\\RALog.txt";
+    const std::wstring mockOldLogFileName = L".\\RACache\\RALog-old.txt";
```
SyrianBallaS: If you want you can make these `static constexpr`.

tests/mocks/MockFileSystem.hh
Jamiras: Why are you so obsessed about destroying things? None of these methods are critical paths, and until a profiler proves otherwise, not worth optimizing. You might even argue that constantly shifting the stack around is detrimental to performance.

In this particular case, destroying it "early" will only exclude the return statement.

tests/services/FileLogger_Tests.cpp
Jamiras: Apparently I can't.
```
Error 	E2402	a constexpr variable must have a literal type or a reference type	RA_Integration.Tests
Error	C2131	expression did not evaluate to a constant	RA_Integration.Tests
```	
However, I can make a `constexpr` point to the string itself:
```
    static constexpr const wchar_t* mockLogFileName = L".\\RACache\\RALog.txt";
```
But then a new string object is created every time it's used. The current implementation creates a single string object for the life of the test fixture.

src/services/impl/FileLogger.hh
SyrianBallaS: Good point

PR #141, opened 2018-10-17
ComparisonVariableSize to MemSize

SyrianBallaS:
Basically mad the `enum `into an `enum class`. Its associated array was made into a compile-time constant of `std::array`. Some loops were changed where the arrays were. It's possible to make `CompVariable `used in constant expressions so it's a true literal type now. 

Majority of the lines changed were just changing the enum's name and adjusting for that.


src/RA_Condition.h
```diff
@@ -60,46 +74,75 @@ class CompVariable
...
-        m_nVal = nVal;
-        m_nPreviousVal = nPrevVal;
+        this->swap(CompVariable{ nSize, nType, nInitialValue });
     }
 
-    void ResetDelta()
+    _CONSTANT_FN SetValues(_In_ unsigned int nVal, _In_ unsigned int nPrevVal) noexcept
     {
-        m_nPreviousVal = m_nVal;
+        this->swap(CompVariable{ nVal, nPrevVal });
```
Jamiras: `swap` might work for `Set` because previous value is unimportant when `Set` is called, but this will not work here. The size and type variables must not be overridden.

src/RA_Condition.h
```diff
@@ -60,46 +74,75 @@ class CompVariable
...
 
     _CONSTANT_FN SetType(_In_ Type nType) noexcept { m_nVarType = nType; }
     _NODISCARD _CONSTANT_FN GetType() const noexcept { return m_nVarType; }
 
-    inline unsigned int RawValue() const { return m_nVal; }
-    inline unsigned int RawPreviousValue() const { return m_nPreviousVal; }
+    _NODISCARD _CONSTANT_FN RawValue() const noexcept { return m_nVal; }
+    _NODISCARD _CONSTANT_FN RawPreviousValue() const noexcept { return m_nPreviousVal; }
+
+    inline void swap(_Inout_ CompVariable& a) noexcept
```
Jamiras: I'm not convinced of the merit of these methods, but as they're only used by local functions, they should be private.

src/services/SearchResults.cpp
```diff
@@ -103,36 +103,40 @@ static const char* ComparisonString(ComparisonType nCompareType)
     }
 }
 
-static unsigned int GetValue(const unsigned char* pBuffer, unsigned int nOffset, ComparisonVariableSize nSize)
+_Success_(return)
+_NODISCARD _CONSTANT_FN GetValue(_In_z_ const unsigned char* const pBuffer,
+                                 _In_   unsigned int nOffset,
+                                 _In_   MemSize nSize) noexcept
```
Jamiras: You lost the `static unsigned int`, which makes it unclear that this method is only used by this file, and what the return type is. I see the `_CONSTANT_FN` expands to `constexpr auto`. `constexpr` is supposed to be used to indicate it's compile-time constant, but this method is called with three variables, so nothing about this function is compile-time constant. Besides confusing me, what is the purpose of this change?

Also note that `pBuffer` is not a null terminated string so the `_In_z_` attribute may be inappropriate. It's a pointer to a block of memory.

src/services/SearchResults.cpp
```diff
@@ -103,36 +103,40 @@ static const char* ComparisonString(ComparisonType nCompareType)
...
 {
+    // TBD: See if std::byte fits are purposes better
```
Jamiras: are => our

src/services/SearchResults.cpp
```diff
@@ -103,36 +103,40 @@ static const char* ComparisonString(ComparisonType nCompareType)
...
+    [[maybe_unused]] const auto uItemAtOffs{ static_cast<unsigned int>(pBuffer[nOffset]) };
```
Jamiras: This feels like unnecessary code obfuscation. The duplication made it clear exactly what was happening in each case, and didn't require declaring the variable as sometimes unused.

tests/RA_UnitTestHelpers.h
```diff
@@ -14,9 +14,15 @@ namespace CppUnitTestFramework {
...
+template<> static std::wstring ToString<MemSize>(const MemSize& t)
 {
-    return ra::Widen(COMPARISONVARIABLESIZE_STR[(int)t]);
+#if _MBCS
+    return ra::Widen(MEMSIZE_STR.at(ra::etoi(t)));
+#elif _UNICODE
+    return MEMSIZE_STR.at(ra::etoi(t));
+#else
+#error Unsupported character set detected
+#endif /* _MBCS */
```
Jamiras: don't the `ra::Widen` overloads handle these `#if`s?

src/RA_Condition.h
SyrianBallaS: I got you. I'll get rid of swap while I'm at it.

src/services/SearchResults.cpp
SyrianBallaS: It'll go back to being just static, and the annotation for pBuffer will change to just _In_ since it's not a null terminated string.

src/services/SearchResults.cpp
SyrianBallaS: The return type is different in the first case. I'll try to do it with more clarity.

tests/RA_UnitTestHelpers.h
SyrianBallaS: You are totally right. Have to look at again but we probably don't even need the widen.

PR #142, opened 2018-10-17
Enum to Enum Class (PopupState)

SyrianBallaS:
This one was as minimal as possible, just changed `PopupState `into an `enum class` and it's a private type member of `LeaderboardPopup `because it wasn't used outside the class.


SyrianBallaS:
The line count is inaccurate. reopening soon.

PR #146, opened 2018-10-19
ViewModel for Get ROM Checksum

Jamiras:
* added support for modal dialog to DialogBase
* illustrates button click redirect to ViewModel


src/services/impl/WindowsClipboard.hh
```diff
@@ -0,0 +1,47 @@
...
+            // assign the memory to the clipboard
+            if (OpenClipboard(nullptr))
+            {
+                EmptyClipboard();
+                SetClipboardData(CF_UNICODETEXT, hMem);
+                CloseClipboard();
+            }
+            else
+            {
+                GlobalFree(hMem);
```
SyrianBallaS: If you aren't gonna make hMem a const pointer I think you should set it to nullptr after freeing its memory to prevent pointer dangling.

src/services/impl/WindowsClipboard.hh
```diff
@@ -0,0 +1,47 @@
...
+{
+public:
+    void SetText(const std::wstring& sValue) const override
+    {
+        // allocate memory to be managed by the clipboard
+        SIZE_T nSize = (sValue.length() + 1) * 2; // 16-bit characters, with null terminator
+        HGLOBAL hMem = GlobalAlloc(GMEM_MOVEABLE, nSize);
+        if (hMem)
+        {
+            LPVOID pMem = GlobalLock(hMem);
```
SyrianBallaS: Seems this should be a const pointer since the pointer shouldn't be modified. It's kinda weird but putting const in front will do just that. As if you did "void* const".

src/ui/viewmodels/GameChecksumViewModel.cpp
```diff
@@ -0,0 +1,30 @@
...
+namespace ra {
+namespace ui {
+namespace viewmodels {
+
+const StringModelProperty GameChecksumViewModel::ChecksumProperty("GameChecksumViewModel", "Checksum", L"No game loaded.");
+
+GameChecksumViewModel::GameChecksumViewModel() noexcept
+{
+    const auto& pGameContext = ra::services::ServiceLocator::Get<ra::data::GameContext>();
+    auto& sGameHash = pGameContext.GameHash();
```
SyrianBallaS: This should probably be const.

src/ui/viewmodels/GameChecksumViewModel.hh
```diff
@@ -0,0 +1,41 @@
...
+#include "ui/WindowViewModelBase.hh"
+
+namespace ra {
+namespace ui {
+namespace viewmodels {
+
+class GameChecksumViewModel : public WindowViewModelBase
+{
+public:
+    explicit GameChecksumViewModel() noexcept;
```
SyrianBallaS: You dont need the explicit for parameterless constructors.

src/ui/win32/DialogBase.cpp
```diff
@@ -99,13 +139,23 @@ BOOL DialogBase::OnCommand(WORD nCommand)
...
 
         case IDCLOSE:
         case IDCANCEL:
             m_vmWindow.SetDialogResult(ra::ui::DialogResult::Cancel);
-            DestroyWindow(m_hWnd);
+
+            if (m_bModal)
+                EndDialog(m_hWnd, nCommand);
+            else
+                DestroyWindow(m_hWnd);
```
SyrianBallaS: Not sure if it should be done here but the handle should be set to nullptr after being destroyed since the handle us invalid now.

src/services/impl/WindowsClipboard.hh
Jamiras: This seems like a weak argument. If I make it const, it'll still dangle. Additionally, there isn't any code after this line before it goes out of scope.

I'll make it const to match the changes made for`pMem`.

src/ui/viewmodels/GameChecksumViewModel.cpp
SyrianBallaS: Sure that's fine.

PR #148, opened 2018-10-20
Extract WinHttp code into mockable service, add Request/Response data structures

### LINKED ISSUE #114: Passwords are not URL-escaped

Jamiras:
* `IHttpRequester` contains the actual logic for making the HTTP call
* `Http::Request` defines the parameters to construct the HTTP request
* `Http::Response` contains the response and may be returned directly via `Http::Request::Call` or through an asynchronous callback via `Http::Request::CallAsync`.
* Url parameters are now URL encoded (fixes #114)
* Image downloads are no longer routed through RAWeb. They now use `IHttpRequester::Download`, which writes the image directly to disk without first buffering it in RAM.


src/RA_httpthread.cpp
```diff
@@ -327,242 +204,49 @@ BOOL RAWeb::DoBlockingRequest(RequestType nType, const PostArgs& PostData, rapid
...
     }
 
-    //  Debug logging...
-    if (ResponseOut.size() > 0)
-    {
-        rapidjson::Document doc;
-        doc.Parse(ResponseOut.c_str());
+    ra::services::Http::Request request(sUrl);
+    request.SetPostData(sPostData);
+    auto response = request.Call();
```
Jamiras: Illustrates synchronous call

src/services/ImageRepository.cpp
```diff
@@ -164,31 +165,36 @@ void ImageRepository::FetchImage(ImageType nType, const std::string& sName)
...
+    RA_LOG_INFO("Downloading %s", sUrl.c_str());
 
-    ra::services::ServiceLocator::GetMutable<ra::services::IThreadPool>().RunAsync([this, nRequestType, args, sFilename]()
+    Http::Request request(sUrl);
+    request.DownloadAsync(sFilename, [this,sFilename,sUrl](const Http::Response& response)
     {
-        std::string sResponse;
-        if (RAWeb::DoBlockingRequest(nRequestType, args, sResponse))
-            _WriteBufferToFile(sFilename, sResponse);
+        if (response.StatusCode() == 200)
```
Jamiras: Illustrates asynchronous download call.

src/services/Http.hh
```diff
@@ -0,0 +1,155 @@
...
+    class Response
+    {
+    public:
+        explicit Response(int nStatusCode, const std::string& sResponse) noexcept
+            : m_nStatusCode(nStatusCode), m_sResponse(sResponse)
+        {
+        }
+
+        Response(const Response&) noexcept = default;
+        Response(Response&&) noexcept = default;
```
SyrianBallaS: You should probably default the other operations too. Destructor and the assignments. If you didn't want the assignments you should delete them.

src/services/Http.hh
```diff
@@ -0,0 +1,155 @@
...
+                m_sUrl = sUrl;
+            }
+            else
+            {
+                m_sUrl.assign(sUrl, 0, nIndex);
+                m_sQueryString.assign(sUrl, nIndex + 1, std::string::npos);
+            }
+        }
+
+        Request(const Request&) = default;
```
SyrianBallaS: You should default and/or delete all of them if you're going to do this. It'll also make it clear what's to be used so we don't have to guess what's implicitly deleted. If you want copying, moving can't be deleted.

src/services/IHttpRequester.hh
```diff
@@ -0,0 +1,33 @@
...
+#include "services\Http.hh"
+#include "services\TextWriter.hh"
+
+namespace ra {
+namespace services {
+
+class IHttpRequester
+{
+public:
+    virtual ~IHttpRequester() noexcept = default;
```
SyrianBallaS: You should delete/default all of them.

PR #150, opened 2018-10-21
expire cache files older than 30 days

Jamiras:


src/services/impl/FileLocalStorage.cpp
```diff
@@ -17,21 +19,47 @@ _CONSTANT_VAR RA_DIR_DATA = L"RACache\\Data\\";
...
+    else if (bExpireOldFiles)
+    {
+        std::vector<std::wstring> vFiles;
+        if (pFileSystem.GetFilesInDirectory(sDirectory, vFiles) > 0)
+        {
+            const auto& pClock = ServiceLocator::Get<IClock>();
+            const auto tExpire = pClock.Now() - std::chrono::hours(24 * 30); // 30 days
+            std::wstring sPath;
+
+            for (auto sFile : vFiles)
```
SyrianBallaS: The `sFile `should be a reference.

Jamiras: fixed.

PR #155, opened 2018-10-24
Spectre Mitigations

SyrianBallaS:
This PR mostly has to do with making are code as immune as possible to the **Spectre security vulnerability**. 

### tl;dr:
- Applied code-based spectre mitigation, so we don't have to rely on `/QSpectre` as it is known to slow down running-time significantly 
  - i.e. A software based CPU microcode Windows Update a few months back (which only worked on Intel processors btw)
- Applied` /W4 `to test project
   - Fixed the unit tests that were failing after addressing the issues here
- All unit tests pass

It usually happens when you assign a variable from an unbounded array index (`[]`) and then assign another variable to the one that was just assigned, if the array was to go out-of-bounds memory corruption takes place and an attacker could inject malicious code to be executed remotely. If you want to see where they were without turning on /Wall, you'll have to put `/w45045 `as a command-line argument for the compiler and it should trigger 15 times. This also deals with other security issues as well but tried to limit this scope of the PR as much as possible.

I forgot to turn on `/W4` for the test project and it did come up with warnings, mostly about non-standard extensions that have been discussed before such as using an `rvalue reference as an lvalue reference`. Unless the compiler adjusts for it, it will result in `undefined behavior`. After making the areas C++ standards compliant, some tests were failing (unsurprisingly), so they had to be adjusted as well.

I'll leave some review comments for areas that be confusing for some people.


src/RA_Leaderboard.cpp
```diff
@@ -176,17 +168,7 @@ void RA_Leaderboard::SubmitRankInfo(unsigned int nRank, const std::string& sUser
...
-            if (m_RankInfo.at(i).m_nRank > m_RankInfo.at(j).m_nRank)
-            {
-                Entry temp = m_RankInfo[i];
-                m_RankInfo[i] = m_RankInfo[j];
-                m_RankInfo[j] = temp;
-            }
-        }
-    }
+    static_assert(ra::is_nothrow_lessthan_comparable_v<decltype(m_RankInfo)>);
+    std::sort(m_RankInfo.begin(), m_RankInfo.end());
```
SyrianBallaS: `N^2 logN` to `N logN`. The static_assert should be self-explanatory, but the test asserts at compile-time that `RA_Leaderboard::Entry` is less than comparable. Meaning it can be compared with the less than operator (`operator<`)
Since std::sort uses `std::less()(value_type, value_type)`, which is a function object of `operator<`.

src/RA_Leaderboard.h
```diff
@@ -51,22 +51,63 @@ class RA_Leaderboard
...
+    MemValue                m_progress;     //  A collection of memory addresses, used to show progress towards completion.
+    MemValue::Format        m_nFormat;      //  A format to output. Typically "%d" for score or "%02d:%02d.%02d" for time
 
-    std::string				m_sTitle;		//	The title of the leaderboard
-    std::string				m_sDescription;	//	
+    std::string             m_sTitle;       //  The title of the leaderboard
+    std::string             m_sDescription; //  
 
-    std::vector<Entry>	    m_RankInfo;		//	Recent users ranks
+    std::vector<Entry>      m_RankInfo;     //  Recent users ranks
```
SyrianBallaS: Turned on a setting that lets me look at the spacing visually. The were all tabs.

src/RA_StringUtils.cpp
```diff
@@ -85,14 +85,16 @@ std::string Narrow(const std::string& str)
...
+    auto dCount = std::count_if(str.begin(), str.end(), [](const char c) noexcept
+    {
+        return (c == '\r' || c == '\n');
+    });
+
+    while (dCount != 0)
+    {
+        str.pop_back();
+        dCount--;
+    }
```
SyrianBallaS: I could have made this logarithmic but because some of tests had line endings in the middle it couldn't be helped. It's still pretty fast.

```
Test                      | Time
‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
TestTrimLineEnding        | 0:00:00.0001801
TestThreeLinesEmptyMiddle | 0:00:00.0001469
TestOneLineWithLineEnding | 0:00:00.00005
TestTwoLines              | 0:00:00.0001768
```
The way it was before made the tests fail after correcting the `TestTrimLineEnding` test.

src/pch.h
```diff
@@ -68,6 +69,15 @@
...
 
 #include <md5.h>
+
+#if RA_UTEST
+#pragma warning(push)
+#include <CodeAnalysis\Warnings.h>
+#pragma warning(disable : ALL_CPPCORECHECK_WARNINGS)
+#include <CppUnitTest.h>
+#pragma warning(pop)
+#endif /* RA_UTEST */
```
SyrianBallaS: This file gives off `/W4` errors as well so it's here.

src/services/SearchResults.cpp
```diff
@@ -231,15 +231,21 @@ void SearchResults::ProcessBlocksNibbles(const SearchResults& srSource, unsigned
...
-        const unsigned char* pPrev = block.GetBytes();
+        std::vector<unsigned char> vMemory2(vMemory.capacity());
+        _CONSTANT_LOC pPrevSize {
+            sizeof(block.GetBytes()) / sizeof(*block.GetBytes())
+        };
+        std::array<unsigned char, pPrevSize> pPrev{};
 
-        g_MemManager.ActiveBankRAMRead(pMemory, block.GetAddress(), block.GetSize());
+        for (auto i = 0U; i < pPrevSize; i++)
+            pPrev.at(i) = block.GetBytes()[i];
```
SyrianBallaS: Honestly couldn't find a better way to do this.

src/services/SearchResults.cpp
```diff
@@ -231,15 +231,21 @@ void SearchResults::ProcessBlocksNibbles(const SearchResults& srSource, unsigned
         if (block.GetSize() > vMemory.capacity())
             vMemory.reserve(block.GetSize());
 
-        unsigned char* pMemory = vMemory.data();
-        const unsigned char* pPrev = block.GetBytes();
+        std::vector<unsigned char> vMemory2(vMemory.capacity());
+        _CONSTANT_LOC pPrevSize {
+            sizeof(block.GetBytes()) / sizeof(*block.GetBytes())
+        };
+        std::array<unsigned char, pPrevSize> pPrev{};
```
SyrianBallaS: The compiler may not warn with C arrays, but it will warn if `std::array` is uninitialized, it'll fill all the slots with '\0' instead of garbage, leading possible stack corruption.

src/services/SearchResults.h
```diff
@@ -126,7 +126,7 @@ class SearchResults
     private:
         union // 8 bytes
         {
-            unsigned char m_vBytes[8];
+            unsigned char m_vBytes[8]{};
```
SyrianBallaS: This kept on giving be garbage, this is a lot simpler than doing...
` unsigned char m_vBytes[8] = {static_cast<unsigned char>('\0'), static_cast<unsigned char>('\0'), static_cast<unsigned char>('\0'), static_cast<unsigned char>('\0'), static_cast<unsigned char>('\0'), static_cast<unsigned char>('\0'), static_cast<unsigned char>('\0'), static_cast<unsigned char>('\0'), };`
... in class/structure **initializers** is the standard way of **initializing** something in C++. It makes perfect sense, **initialize** with an **initializer**. The `={}` syntax is not **direct initialization**, that's **copy-initialization**. Why copy when we don't have too? Basically using `={}` means copy the right side into to left instead of just initializing directly.

Whether this is intended to be a block of memory or not, the compiler still thinks it's an array of characters. _unsigned char_ is a _character_ type. We could use std::byte if we aren't to assume it's unsigned char, it has all bit-wise and relational operator overloads already.

tests/RA_StringUtils_Tests.cpp
```diff
@@ -46,12 +44,22 @@ TEST_CLASS(RA_StringUtils_Tests)
...
+            "test\n\n",
+            "test\r\n\r\n"
+        };
+
+        Assert::AreEqual(test_strings.at(0), TrimLineEnding(test_strings.at(0)));
+        Assert::AreEqual(test_strings.at(0), TrimLineEnding(test_strings.at(1)));
+        Assert::AreEqual(test_strings.at(0), TrimLineEnding(test_strings.at(2)));
+        Assert::AreEqual(test_strings.at(0), TrimLineEnding(test_strings.at(3)));
+        Assert::AreEqual(test_strings.at(2), TrimLineEnding(test_strings.at(4)));
+        Assert::AreEqual(test_strings.at(3), TrimLineEnding(test_strings.at(5)));
```
SyrianBallaS: This is where the` /W4` warnings were triggered. The param is supposed be `std::string&` but previously it was `std::string&&`, temporary variables do not go out, they get destroyed. Scopes are very important in object-oriented languages. You could use an `rvalue reference` instead of a `const lvalue reference`, but if it's not `const `it's undefined unless an extension adjusts for it.

tests/RA_UnitTestHelpers.h
```diff
@@ -17,33 +15,34 @@ namespace VisualStudio {
 namespace CppUnitTestFramework {
 
 // converters for asserting enum values
-
-template<> static std::wstring ToString<MemSize>(const MemSize& t)
+#pragma warning(push)
+#pragma warning(disable : 4505) // unreferenced inline functions
+template<> std::wstring ToString<MemSize>(const MemSize& t)
```
SyrianBallaS: `static template` overloads are undefined, unless the compiler adjusts for it. I've never seen static template overloads before either, it doesn't make any sense.

Anyway these functions were considered unreferenced inline functions but didn't want to deal with that now.

SyrianBallaS:
It's a bit difficult to fix this branch, I'll make a new one.

src/services/SearchResults.cpp
Jamiras: `block.GetBytes()` returns a pointer, not an array. why are you trying to calculate size by dividing the size of the pointer by the size of an item in the pointer? 

pPrev is either going to point to the buffer in the block, or to memory on the heap. It's functionality similar to SSO. If the block of memory is smaller than the size of a pointer, just write it where the pointer would go. I don't even want to try to fathom what this change is trying to do. pPrev is not local data.

src/services/SearchResults.h
Jamiras: This is just 8 bytes of RAM. it should be a `uint8_t`.

PR #157, opened 2018-10-24
CppCoreCheck Const Rules

SyrianBallaS:
Basically a const correctness PR but used the analyzer to find them faster. If a parameter or variable should be `const` it was was marked as `const`. Tried to keep the focus to task at hand.

Some export functions from the overlay and popup APIs were changed if they should've been const as well, as far as I can tell, the overlay still renders the same, the examine pages still render the same, and the popups still render and animate the same.

API violations and invalid annotation syntax was addressed as well. One of them was left alone because it was causing some Unit Tests to fail. We should probably mark non const references that utilize RAII as an "in-out" parameter instead of "out" because it's already initialized at construction (`CompVariable::SerializeAppend`). Normally we shouldn't be using uninitialized variables in the first place.


src/RA_Dlg_AchEditor.cpp
```diff
@@ -1897,7 +1897,7 @@ void Dlg_AchievementEditor::UpdateSelectedBadgeImage(const std::string& sBackupB
         sAchievementBadgeURI = sBackupBadgeToUse;
 
     m_hAchievementBadge.ChangeReference(ra::services::ImageType::Badge, sAchievementBadgeURI);
-    HBITMAP hBitmap = m_hAchievementBadge.GetHBitmap();
+    const HBITMAP__* hBitmap = m_hAchievementBadge.GetHBitmap();
```
Jamiras: Why can't this be `const HBITMAP hBitmap`? I'd prefer to keep the exposed aliases, rather than rely on the underlying type that might get renamed at some point in the future.

src/RA_Integration.vcxproj
```diff
@@ -329,7 +329,7 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Analysis|Win32'">
     <TargetName>RA_Integration</TargetName>
-    <CodeAnalysisRuleSet>CppCoreCheckClassRules.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRuleSet>CppCoreCheckConstRules.ruleset</CodeAnalysisRuleSet>
```
Jamiras: Is there some way to enforce both? i.e. `CppCoreCheckClassRules.ruleset,CppCoreCheckConstRules.ruleset`?

src/RA_PopupWindows.h
```diff
@@ -41,11 +42,11 @@ class PopupWindows
...
-extern "C"
-{
-    API extern int _RA_UpdatePopups(ControllerInput* input, float fDTime, bool Full_Screen, bool Paused);
-    API extern int _RA_RenderPopups(HDC hDC, RECT* rcSize);
-}
+_EXTERN_C
+_NODISCARD API int _RA_UpdatePopups(_In_ const ControllerInput* const __restrict input, _In_ float fDTime,
+                                    _In_ bool Full_Screen, _In_ bool Paused);
+_NODISCARD API int _RA_RenderPopups(_In_ HDC hDC, _In_ const RECT* const __restrict rcSize);
+_END_EXTERN_C
```
Jamiras: These cannot be `_NODISCARD`. The calling methods are currently ignoring the return value: https://github.com/RetroAchievements/RAIntegration/blob/master/src/RA_Interface.cpp#L114-L118

src/RA_PopupWindows.h
```diff
@@ -41,11 +42,11 @@ class PopupWindows
...
 
 //	Exposed to DLL
-extern "C"
-{
-    API extern int _RA_UpdatePopups(ControllerInput* input, float fDTime, bool Full_Screen, bool Paused);
-    API extern int _RA_RenderPopups(HDC hDC, RECT* rcSize);
-}
+_EXTERN_C
+_NODISCARD API int _RA_UpdatePopups(_In_ const ControllerInput* const __restrict input, _In_ float fDTime,
+                                    _In_ bool Full_Screen, _In_ bool Paused);
```
Jamiras: Does changing the const-ness affect the ability to resolve the method via GetProcAddress?
https://github.com/RetroAchievements/RAIntegration/blob/master/src/RA_Interface.cpp#L453

I assume the annotations do not.

src/RA_PopupWindows.h
SyrianBallaS: Alright. Just for future reference you can use (void) to suppress a nodiscard warning.

src/RA_Integration.vcxproj
SyrianBallaS: You can, but it'd have to be a custom ruleset. The "CppCoreCheckRules" one shows too much (over 1000). I can try doing that.

src/RA_Dlg_AchEditor.cpp
SyrianBallaS: Well I can try suppressing it. Using `const HBITMAP` is the same as `HBITMAP__* const` which isn't what we need.

src/RA_PopupWindows.h
SyrianBallaS: The annotations don't. I think it might affect GetProcAddress. I'll suppress the const rule for the overlay and popup exports.

src/RA_PopupWindows.h
SyrianBallaS: Just to note, we don't actually need gsl to use the suppression because only CppCoreCheck understands it.

PR #158, opened 2018-10-28
Enum to Enum Class (EmulatorID & ConsoleID) w/ Backwards Compatibility 

SyrianBallaS:
I got to thinking, C++ is extremely flexible so with some conditions it should be able to work with targeted platforms/toolsets. I did not touch anything in the `#ifndef RA_EXPORTS` blocks.
There will be some review comments from myself to explain some of the changes.

Basically the `enum`s in the title are `enum class` for C++11 and up and are unscoped `enum` pre-C++11.


src/RA_Interface.h
```diff
@@ -2,102 +2,139 @@
...
 {
-    BOOL m_bUpPressed;
-    BOOL m_bDownPressed;
-    BOOL m_bLeftPressed;
-    BOOL m_bRightPressed;
-    BOOL m_bConfirmPressed;	//	Usually C or A
-    BOOL m_bCancelPressed;	//	Usually B
-    BOOL m_bQuitPressed;	//	Usually Start
+#if (defined(__cplusplus) || (defined(__STDC__) || defined(__STDC_VERSION__)))
+#if ((__cplusplus < 201103L) || (__STDC__ || (__STDC_VERSION__ >= 199901L)))
```
Jamiras: instead of repeating this logic, could you put something at the top of the file that defines capabilities and use those?
```
#if (__STDC_VERSION__ >= 199901L)
  #define HAS_C99 1
  #if (defined(__cplusplus) && (__cplusplus >= 201103L))
    #define HAS_CPP11 1
  #endif
#endif
```
This logic may not be perfect, but I think you get the idea.

Then you could make the processing logic easier to read:
```
#if HAS_CPP11
  #define _ENUM_CLASS enum class
#elif HAS_C99
  #define _ENUM_CLASS enum
#else
  #define _ENUM_CLASS typedef enum
#endif
```

SyrianBallaS: alright, I'm going to try doing your suggestions. Gonna close this PR for now.

SyrianBallaS:
I got RALibretro to compile with this, gonna try with the other emus. Remember `/Zc:__cplusplus` has to be added as an additional command line argument for MSVC or `__cplusplus `will show `199711L `even though it's wrong (`C++98`). 
https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2017

If using MSVC it'll compile as C++14, I saw the makefile and it said `CXXFLAGS=$(CCFLAGS) -std=c++11`.
There's a few changes you'll need to do there.

`Application.cpp`
image

`Emulator.h`
image

Noticed some warnings there too. I think they should be addressed or suppressed, but w/e.

Jamiras:
> Remember /Zc:__cplusplus has to be added as an additional command line argument

That's a deal-breaker. If you can't do this without requiring emulator changes, I don't want to do it.

SyrianBallaS:
I can do it w/o `/Zc:__cplusplus`, the enums will still need to be `static_cast` and use their scoped names. If you don't want to do that either then that's fine.

Jamiras:
Yea, lets hold off on this. It feels like a breaking API change. 

We probably need to consider how this will be affected by the `rcheevos` integration, which also defines these constants (https://github.com/RetroAchievements/rcheevos/blob/master/include/rcheevos.h#L39-L71). It's going to become the source of truth, so internally, we'll probably end up using those enums, or something that wraps those enums. We can't directly assign those enum values to this enum without creating a dependency on `rcheevos.h`, which is also a deal-breaker as it would require the emulators know about the rcheevos repository. Additionally, `rcheevos.h` has to be C99 compliant, so we're still stuck using an old-style enum for these constants. Using this enum at least gets us PascalCase.

SyrianBallaS:
I thought it was C89 (ANSI C), if it was we should probably disable language extensions by file if using visual studio so it'll put `__STDC__` as 1 instead of 0. It's either for C89 if C or C++11 but because we use Windows API it will not build, it might for rcheevos. Some of the stuff in rcheevos is nonstandard even for C, I'm trying to figure out how to filter them out from the warnings from spectre.

That's fine about using plain enum if it comes from C, just use the best tools available. If that's what it is then that's what it is.

Unrelated: I'm "done" with the spectre mitigation thing but can't figure out a few things. I'll explain more when I PR it.

PR #160, opened 2018-10-30
Generic ra::StringPrintf and minor improvements to other helpers

SyrianBallaS:
In my other branches I've been using `ra::StringPrintf `a lot. It got kind of annoying to call `NativeStr `almost every-time. So I made it Generic. The only thing you have to  is either put no prefix, `u8 `for `UTF-8`, or `L `for wide-character string in the format string. You could put `_T(string)` if you want if it's supposed to be a "type-character string". No need to specify template parameter.

Example:
```
const auto str{ra::StringPrintf("This is a %s.", "string")};         // std::string (ASCII)
const auto str{ra::StringPrintf(u8"This is a %s.", u8"string")};     // std::string (UTF-8)
const auto str{ra::StringPrintf(L"This is a %s.", L"string")};       // std::wstring
const auto str{ra::StringPrintf(_T("This is a %s."), _T("string"))}; // ra::tstring
```
As you can see, the only thing that matters is the prefix before the format string (or none).
I was originally going to make one for `char16_t `but that's not possible without making a user-defined implementation for it (even though it exists in C11).

Also moved some stuff around. I'll go in detail in some review comments.
There was one precheck and one post check that I didn't do because I was afraid it would cause some overhead. One of them was to check that `%n` (dangerous!) was not in the format string and that the format string was smaller than the capacity (capacity is always larger than size for containers).


src/ra_math.h
```diff
@@ -0,0 +1,43 @@
...
+_NODISCARD inline auto ftol(_In_ FloatingPoint fp) noexcept { return std::lround(fp); }
+
+template<typename FloatingPoint, class = std::enable_if_t<std::is_floating_point_v<FloatingPoint>>>
+_NODISCARD inline auto ftoll(_In_ FloatingPoint fp) noexcept { return std::llround(fp); }
+
+template<typename FloatingPoint, class = std::enable_if_t<std::is_floating_point_v<FloatingPoint>>>
+_NODISCARD inline auto ftoul(_In_ FloatingPoint fp) noexcept { return to_unsigned(std::lround(fp)); }
+
+template<typename FloatingPoint, class = std::enable_if_t<std::is_floating_point_v<FloatingPoint>>>
+_NODISCARD inline auto ftoull(_In_ FloatingPoint fp) noexcept { return to_unsigned(std::llround(fp)); }
```
SyrianBallaS: One of the unit tests was using a `long long` as an argument and threw an exception because it was too big in another branch. So if we still want to return `long`, we'd have to down cast after calling `ra::ftoull`, not all at once.
That test was not touched in this PR.

Also these functions can't be constant expressions so the `_CONSTANT_FN `was taken out.

src/ra_utility.h
```diff
@@ -21,6 +21,29 @@ to_unsigned(_In_ SignedType st) noexcept { return static_cast<std::make_unsigned
...
+{
+    if constexpr (std::is_integral_v<Arithmetic>)
+    {
+        if constexpr (std::is_signed_v<Arithmetic>)
+            return to_unsigned(a);
+        else if constexpr (std::is_unsigned_v<Arithmetic>)
+            return to_signed(a);
+    }
+    if constexpr (std::is_floating_point_v<Arithmetic>)
+        return -a;
```
SyrianBallaS: I made this convenience wrapper so we don't have to remember what the sign is. A "one-size-fits-all" function so to speak.

Jamiras: This doesn't appear to be used, and has different behavior depending on if it's called with an integer or a float. 

I think it has very little benefit. I would prefer the user explicitly called `to_signed` or `to_unsigned`, or `-a` to make it clear what he's trying to do.

src/RA_StringUtils.h
```diff
@@ -24,9 +26,118 @@ _NODISCARD std::string Narrow(_In_ const std::string& wstr);
...
+/// </summary>
+template<typename CharT, typename = std::enable_if_t<is_char_v<CharT>>>
+_NODISCARD inline auto StringPrintf(_In_z_ _Printf_format_string_ const CharT* const __restrict sFormat, ...)
+{
+    assert(sFormat != nullptr);
+    std::basic_string<CharT> sFormatted;
+    const auto post_check =[](int nChars) noexcept
+    {
+        assert(nChars >= 0);
+        assert(nChars < RSIZE_MAX/sizeof(CharT));
```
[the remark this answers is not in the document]
Jamiras: You've eliminated one copy/paste in favor of causing the reader to following the code path into a function defined within the scope of another function. And I'm not entirely sure the compiler is smart enough to eliminate it in a release build where the asserts go away. I'd rather have the two asserts in the two places you call post_check.

Alternately, the second call to post_check should probably be `assert(nCheck == nNeeded)`, in which case there wouldn't be any copy pasting.

src/RA_StringUtils.h
```diff
@@ -24,9 +26,118 @@ _NODISCARD std::string Narrow(_In_ const std::string& wstr);
...
+template<typename CharT, typename = std::enable_if_t<ra::is_char_v<CharT>>>
+_NODISCARD _Success_(0 < return < strsz) inline auto __cdecl
+tcslen_s(_In_reads_or_z_(strsz) const CharT* const str, _In_ std::size_t strsz = UINT16_MAX) noexcept
+{
+    assert(str != nullptr);
+    const auto post_check = [&](std::size_t retVal) noexcept
+    {
+        assert(retVal != 0); // somehow str was a nullptr
+        assert(retVal != strsz); // str isn't null-terminated
+    };
```
Jamiras: similar comment here. this method could be:
```
assert(str != nullptr);
int ret;
if constexpr (std::is_same_v<CharT, char>) {
    ret = strnlen_s(str, strsz);
} else if constexpr (std::is_same_v<CharT, wchar_t>) {
    ret = wcsnlen_s(str, strsz);
} else {
    #error Unsupported CharT
}
assert(ret != 0);
assert(ret != strsz);
return ret;
```
or, you could just have two methods and avoid the templating altogether. it's not gaining you anything in this particular case.
```
_NODISCARD int __cdecl tcslen_s(const char* str, std::size_t strsz = UINT16_MAX) noexcept
{
    int ret = strnlen_s(str, strsz);
    assert(ret != 0);
    assert(ret != strsz);
    return ret;
}
_NODISCARD int __cdecl tcslen_s(const wchar_t* str, std::size_t strsz = UINT16_MAX) noexcept
{
    int ret = wscnlen_s(str, strsz);
    assert(ret != 0);
    assert(ret != strsz);
    return ret;
}
```

src/ra_utility.h
SyrianBallaS: That sounds good to me. Thought it would make things easier. We could just have it for Integrals because sign flipping between integer types and floating-points _are_ different. It can be gone though.

src/RA_StringUtils.h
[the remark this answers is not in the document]
SyrianBallaS: I'll be sure to keep in mind to not put asserts in lambdas are any other function types. Normally an exception should be thrown, caught, handled but that should be for another PR.

SyrianBallaS:
This one shouldn't take too long but still closing it until it's ready.

src/RA_StringUtils.h
SyrianBallaS: I already have std::enable_if_t, the only character types that are allowed are `char `and `wchar_t`. If that condition failed, the dll won't even compile because it's a compile-time test.
They have something similar in C# but with different syntax.
`where T : struct`

Would it be easier for you to follow if I did `static_assert `instead of std::enable_if_t.

detail version:
```
template<typename CharT>
struct _NODISCARD is_char :
    std::bool_constant<(std::is_same_v<CharT, char> || std::is_same_v<CharT, wchar_t>)>
{
};
```
User version:
```
template<typename CharacterType> _CONSTANT_VAR
is_char_v{ detail::is_char<CharacterType>::value };
```

PR #161, opened 2018-11-03
Spectre Mitigation

SyrianBallaS:
Almost identical to #155 but some things are different which I will leave review comments for. I figured out how to make the unit test work unchanged, the trim line ending needed a rvalue reference overload with a different way to find the characters.

There is one spectre mitigation that I couldn't do because of prior feedback and not sure how to do it.
rcheevos uses an unnamed struct and since we have the source files in the test project, the warning suppression won't have an effect (similar situation with the `md5.c`). I tried disabling warnings by file but since `rcheevos.h `is not a source file, it couldn't be configured.


src/RA_StringUtils.h
```diff
@@ -45,9 +48,21 @@ _NODISCARD inline auto StringPrintf(_In_z_ _Printf_format_string_ const CharT* c
...
+        // starting capacity to reduce the need to reallocate, it could still be too small but is handled below
+        sFormatted.reserve(64U);
+        nNeeded = -1; // assume it already failed (usually does)
         assert(std::wstring_view{ sFormat }.find(L"%n") == std::wstring_view::npos);
         assert(sFormatted.capacity() > 0U && (sFormatted.capacity() < RSIZE_MAX/sizeof(wchar_t)));
-        nNeeded = std::vswprintf(sFormatted.data(), sFormatted.capacity(), sFormat, pArgs);
+
+        while(nNeeded < 0) /*constant at best, logk N at worst; where k is the amount of times doubling capacity*/
+        {
+            nNeeded = std::vswprintf(sFormatted.data(), sFormatted.capacity(), sFormat, pArgs);
```
SyrianBallaS: Using `vswprintf_s `will not work, it will just automatically fail if the buffer is too small. With the test in `RA_AchievementPopup.cpp`, it had to double 3 times to 63 (kind of weird, `7->15->31->63`).

So I just put 64 as a default, but keep on doubling it below if it failed. From what I've read, `std::vsnprintf` automatically adjusts the buffer size but `std::vswprintf` does not.

src/api/impl/ConnectedServer.cpp
```diff
@@ -26,19 +26,19 @@ _NODISCARD static bool HandleHttpError(_In_ const ra::services::Http::Response&
...
+        pDocument.SetArray(); // Two Debug Assertion Errors, one from RapidJSON and one from Kernel
 
         RA_LOG_ERR("-- %s: %s", sApiName, pResponse.ErrorMessage.c_str());
         return false;
     }
 
     if (httpResponse.Content().empty())
     {
-        pDocument.Clear();
+        pDocument.SetArray(); // Debug Assertion Error
```
SyrianBallaS: These two where causing debug assertion failures. According the documentation, `rapidjson::Document::SetArray`, sets the Document as an empty array.

src/services/SearchResults.cpp
```diff
@@ -234,17 +234,18 @@ void SearchResults::ProcessBlocksNibbles(const SearchResults& srSource, unsigned
...
     {
         if (block.GetSize() > vMemory.capacity())
-            vMemory.reserve(block.GetSize());
+            vMemory.resize(block.GetSize());
 
-        unsigned char* pMemory = vMemory.data();
-        const unsigned char* pPrev = block.GetBytes();
+        std::array<unsigned char, sizeof(MemBlock)> pPrev{};
+        for (auto i = 0U; i < sizeof(MemBlock); i++)
+            pPrev.at(i) = block.GetBytes()[i];
```
SyrianBallaS: This was the only reasonable way I could do this, if it's still not good enough, we might have to get the Spectre Mitigated compilers and libraries. If there's a better way to do this without resorting to that, please come forward.

src/RA_Dlg_Memory.cpp
```diff
@@ -615,23 +614,23 @@ void MemoryViewerControl::RenderMemViewer(HWND hTarget)
                             g_MemBookmarkDialog.WriteFrozenValue(*bm);
                     }
                 }
+                std::array<unsigned char, 16> data{};
```
Jamiras: this should be allocated once outside the loop

src/RA_StringUtils.cpp
```diff
@@ -85,18 +85,23 @@ std::string Narrow(const std::string& str)
...
-    if (nIndex > 0 && str[nIndex - 1] == '\n')
-        nIndex--;
-    if (nIndex > 0 && str[nIndex - 1] == '\r')
-        nIndex--;
-
-    if (nIndex != str.length())
-        str.resize(nIndex);
-
+    auto sCpy{ str };
+    str = TrimLineEnding(std::move_if_noexcept(sCpy));
```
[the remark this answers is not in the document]
Jamiras: This method was intentionally modifying the input string. Making a copy is undesirable.

src/RA_StringUtils.cpp
```diff
@@ -83,18 +83,24 @@ std::string Narrow(const std::string& str)
...
     return str;
 }
 
+std::string TrimLineEnding(std::string&& str) noexcept
+{
+    auto sRet{ std::move_if_noexcept(str) };
+    
+    // using the .back() and .pop_back() strategy doesn't work
+    if (const auto pos = sRet.rfind('\n', sRet.length()); pos != std::string::npos)
```
Jamiras: This is wrong. We only want to truncate the newline character if it is the _last_ character of the string. `rfind` may return one in the middle of the string. I'm pretty sure the "pos" parameter is where it starts searching backwards from, not where it stops.

src/RA_StringUtils.cpp
```diff
@@ -83,18 +83,24 @@ std::string Narrow(const std::string& str)
...
+    auto sCpy{ str };
+    str = TrimLineEnding(std::move_if_noexcept(sCpy));
     return str;
 }
 
+std::string TrimLineEnding(std::string&& str) noexcept
+{
+    auto sRet{ std::move_if_noexcept(str) };
+    
+    // using the .back() and .pop_back() strategy doesn't work
```
Jamiras: Firstly, why not?

Secondly, this sort of comment is irrelevant once the code is committed. Someone coming across it in six months will have no idea what the "back/pop_back" strategy was. This is one of the things that should be saved for the review comments.

src/RA_StringUtils.h
Jamiras: `vnsprintf` will return the size of buffer required. `vswprintf` does not, so we have to keep trying sizes until we find one that works. Once successful, the string is populated and can be returned. The "if needed>=0" logic below only needs to set the length.

src/api/impl/ConnectedServer.cpp
Jamiras: I agree with the changes, but please remove the comments. They're review explanations, not describing the committed behavior.

src/pch.h
```diff
@@ -77,6 +76,11 @@
 #pragma warning(pop)
 #endif /* RA_UTEST */
 
+/* rcheevos stuff */
+#pragma warning (push)
+#pragma warning (disable:4201) // nameless struct
+#include "rcheevos\include\rcheevos.h"
+#pragma warning (pop)
```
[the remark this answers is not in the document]
Jamiras: nameless struct should have been fixed by this commit: https://github.com/RetroAchievements/rcheevos/pull/8
if there's still another culprit, we can get those fixed.
note that the rcheevos tag is currently `develop:HEAD` as this change has not yet been merged into master.

src/services/SearchResults.cpp
Jamiras: There has to be a better way. There's absolutely no reason to copy the data just to read it. Maybe instead of storing a pointer to the backing array, add a `GetByte(index)` method.

src/RA_StringUtils.cpp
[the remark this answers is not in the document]
SyrianBallaS: Made it so it doesn't need to copy.

src/RA_Dlg_Memory.cpp
SyrianBallaS: I was looking through this again and could've sworn I've done that after you suggested this. It wasn't but now it is.

src/services/SearchResults.cpp
```diff
@@ -230,21 +230,23 @@ void SearchResults::ProcessBlocksNibbles(const SearchResults& srSource, unsigned
...
     {
         if (block.GetSize() > vMemory.capacity())
-            vMemory.reserve(block.GetSize());
+            vMemory.resize(block.GetSize());
 
-        unsigned char* pMemory = vMemory.data();
-        const unsigned char* pPrev = block.GetBytes();
+        
+        for (auto i = 0U; i < sizeof(MemBlock); i++)
+            pPrev.at(i) = block.GetByte(i);
```
Jamiras: This is wrong. `pPrev` is a 16-byte buffer because that's what `sizeof(MemBlock)` is, but `block.GetSize()` may return much more than 16 bytes.
```
class MemBlock
{
...
        union // 8 bytes
        {
            unsigned char m_vBytes[8]{};
            unsigned char* m_pBytes;
        };

        unsigned int nAddress; // 4 bytes
        unsigned int nSize;    // 4 bytes
};
```
You're effectively copying 16-bytes from `m_pBytes`, regardless of how many bytes are available, or if `m_pBytes` is even used.

`MemBlock` is designed like short string optimization. If the block is 8 bytes or less, the internal buffer is used, otherwise memory is allocated on the heap. `pPrev` is a pointer to access the entire buffer of the previous block for comparing against the current state. As I mentioned a previous review - if you want to avoid the direct pointer logic, you need to call a method to get the nth byte instead of `pPrev[i]`. It looks like you added a `GetByte` method, but that only works if the buffer is less than 8 bytes. if `nSize` is greater than 8 bytes, the heap memory needs to be references. See the implementation of `GetBytes()`.

This should be dirt simple for you to test. Simply open the Memory Inspector for any game and hit "New 4-bit test" (`ProcessBlocksNibbles` is only called in 4-bit test mode). This will allocate a single block large enough to hold the entire RAM (i.e. 65536 bytes on NES) - much larger than the 16-bytes you're providing.

Additionally, you didn't make changes to the lambda functions passed to `ProcessBlocks`. I would expect the same problem/solution to be necessary there. There are unit tests for `ProcessBlocks` and large memory, but it doesn't look like there's any for `ProcessBlocksNibbles` with large memory.

src/services/SearchResults.h
```diff
@@ -118,14 +118,15 @@ class SearchResults
 
         unsigned char* GetBytes() { return (nSize > sizeof(m_vBytes)) ? m_pBytes : &m_vBytes[0]; }
         const unsigned char* GetBytes() const { return (nSize > sizeof(m_vBytes)) ? m_pBytes : &m_vBytes[0]; }
-
+        auto GetByte(std::size_t nIndex) noexcept { return m_vBytes[nIndex]; }
+        auto GetByte(std::size_t nIndex) const noexcept { return m_vBytes[nIndex]; }
```
Jamiras: This only works in `m_nSize` is 8 or less. See `GetBytes` implementation on the previous line. This should be:
```
unsigned char GetByte(size_t nIndex) { return GetBytes()[nIndex]; }
```
Which is more-or-less exactly what the code was doing before. It was just storing the result of `GetBytes()` in a local variable.

src/RA_StringUtils.cpp
```diff
@@ -83,16 +83,22 @@ std::string Narrow(const std::string& str)
...
-    if (nIndex > 0 && str[nIndex - 1] == '\n')
-        nIndex--;
-    if (nIndex > 0 && str[nIndex - 1] == '\r')
-        nIndex--;
+    return (str = TrimLineEnding(std::move_if_noexcept(str)));
+}
+
+std::string TrimLineEnding(std::string&& str) noexcept
+{
+    auto sRet{ std::move_if_noexcept(str) };
```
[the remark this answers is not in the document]
SyrianBallaS: OK, I'll do this.

src/RA_Dlg_Memory.cpp
```diff
@@ -616,22 +615,22 @@ void MemoryViewerControl::RenderMemViewer(HWND hTarget)
                     }
                 }
 
-                g_MemManager.ActiveBankRAMRead(data, addr, 16);
+                g_MemManager.ActiveBankRAMRead(data.data(), addr, 16);
 
-                TCHAR* ptr = bufferNative + wsprintf(bufferNative, TEXT("0x%06x  "), addr);
+                TCHAR* ptr = bufferNative + _stprintf_s(bufferNative, 9, TEXT("0x%06x  "), addr);
```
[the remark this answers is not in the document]
SyrianBallaS: sure

src/services/SearchResults.cpp
SyrianBallaS: I'll see if it'll work with a polymorphic container so we don't have to specify size.

src/services/SearchResults.h
SyrianBallaS: sure.

src/services/SearchResults.cpp
```diff
@@ -230,21 +230,22 @@ void SearchResults::ProcessBlocksNibbles(const SearchResults& srSource, unsigned
...
     for (auto& block : srSource.m_vBlocks)
     {
         if (block.GetSize() > vMemory.capacity())
-            vMemory.reserve(block.GetSize());
+            vMemory.resize(block.GetSize());
 
-        unsigned char* pMemory = vMemory.data();
-        const unsigned char* pPrev = block.GetBytes();
+        for (auto i = 0U; i < block.GetSize(); i++)
+            pPrev.emplace_back(block.GetByte(i));
```
Jamiras: You're still creating an unnecessary copy of the data. pPrev will be 4MB on an N64 game (8MB with the expansion pak).

Recomment replacing the "pPrev.at(i)" below with "block.GetByte(i)".

SyrianBallaS: done

PR #162, opened 2018-11-03
Enum to enum class ComparisonType

SyrianBallaS:
Tries to be as minimal as possible. Pretty much just makes it a scoped enum.


src/RA_Condition.h
```diff
@@ -39,18 +39,24 @@ inline constexpr std::array<const LPCTSTR, 13> MEMSIZE_STR
 };
 
 
-enum ComparisonType
+enum class ComparisonType : std::size_t
```
Jamiras: Why specify the size of this, if not limiting it to something like 8-bits? The compiler will choose an appropriate size.

SyrianBallaS: It's not size, its underlying type. I just need it to be unsigned. Most of the at methods for the array want size_t

Jamiras: I had assumed the compiler would use a smaller type for a smaller enum. It looks like it doesn't.
```
enum ComparisonType
{
    Equals,
    LessThan,
    LessThanOrEqual,
    GreaterThan,
    GreaterThanOrEqual,
    NotEqualTo,

    NumComparisonTypes
};
enum class ComparisonType2 : size_t
{
    Equals,
    LessThan,
    LessThanOrEqual,
    GreaterThan,
    GreaterThanOrEqual,
    NotEqualTo,

    NumComparisonTypes
};
static_assert(sizeof(ComparisonType) == sizeof(ComparisonType2));
```

PR #164, opened 2018-11-03
Enum to enum class ReporterColumns

SyrianBallaS:
All in the title. I have two more that have been ready but haven't PR'd them yet.


src/RA_Dlg_AchievementsReporter.cpp
```diff
@@ -126,40 +133,30 @@ INT_PTR CALLBACK Dlg_AchievementsReporter::AchievementsReporterProc(HWND hDlg, U
...
                     {
                         if (ListView_GetCheckState(hList, i) != 0)
                         {
-                            //	NASTY big assumption here...
-                            char buffer[1024];
-                            sprintf_s(buffer, 1024, "%u,", g_pActiveAchievements->GetAchievement(i).ID());
-                            strcat_s(sBuggedIDs, 1024, buffer);
-
-                            //ListView_GetItem( hList );	
+                            sBuggedIDs.append(ra::StringPrintf("%zu, ", g_pActiveAchievements->GetAchievement(i).ID()));
```
Jamiras: You've added an extra space here. While it might be visually appealing, this also gets sent to the API and may have unintended side effects there. For now, recommend removing the space.

src/RA_Dlg_AchievementsReporter.cpp
```diff
@@ -30,78 +27,86 @@ void Dlg_AchievementsReporter::SetupColumns(HWND hList)
...
+        {
+            col.mask       = ra::to_unsigned(LVCF_TEXT | LVCF_WIDTH | LVCF_SUBITEM | LVCF_FMT),
+            col.fmt        = LVCFMT_LEFT | LVCFMT_FIXED_WIDTH,
+            col.cx         = COL_SIZE.at(i),
+            col.pszText    = str.data(),
+            col.cchTextMax = 255,
+            col.iSubItem   = i
+        };
+
+        if (i == ra::to_signed(COL_TITLE.size() - 1))	//If the last element: fill to the end
```
Jamiras: The intent of `i` is to be unsigned. Recommend changing it back to a size_t, and using `to_unsigned` on line 41.

src/RA_Dlg_AchievementsReporter.cpp
```diff
@@ -30,78 +27,86 @@ void Dlg_AchievementsReporter::SetupColumns(HWND hList)
...
+_Use_decl_annotations_
+int Dlg_AchievementsReporter::AddAchievementToListBox(HWND hList, const Achievement* const pAch)
 {
-    for (size_t i = 0; i < NumReporterColumns; ++i)
+    // We aren't actually using the value so we're using iterators
+    for (auto it = COL_TITLE.cbegin(); it != COL_TITLE.cend(); ++it)
     {
-        switch (i)
+        const auto nPos{ std::distance(COL_TITLE.cbegin(), it) };
+        switch (const auto col{ ra::itoe<Column>(nPos) }; col)
```
Jamiras: Why do you need to define `col` here? Can't you just:
```
switch (ra::itoe<Column>(nPos))
```

src/RA_Dlg_AchievementsReporter.h
```diff
@@ -8,34 +8,38 @@ class Achievement;
...
+    inline static constexpr std::array<int, 5> COL_SIZE{ 19, 105, 205, 75, 62 };
 
 public:
     static void DoModalDialog(HINSTANCE hInst, HWND hParent);
     static INT_PTR CALLBACK AchievementsReporterProc(HWND, UINT, WPARAM, _UNUSED LPARAM);
 
 public:
     static void SetupColumns(HWND hList);
-    static int AddAchievementToListBox(HWND hList, const Achievement* pAch);
+    _NODISCARD static int AddAchievementToListBox(_In_ HWND hList, _In_ const Achievement* const pAch);
```
Jamiras: I find it humorous that you make this `_NODISCARD` and add an `_UNUSED` variable in the one place that calls it.

Recommend just changing the function to return `void`.

src/RA_Dlg_AchievementsReporter.cpp
SyrianBallaS: Yeah that can be done.

src/RA_Dlg_AchievementsReporter.h
SyrianBallaS: alright

PR #169, opened 2018-11-05
abstract Login/Logout APIs

Jamiras:
* Provides `Login` and `Logout` container classes, each with `Request`, `Response`, and `Callback` definitions and `Call` and `CallAsync` methods (mimics `Http` container class). This template can easily be extended to other APIs.
  * `CallAsync` manages the request/response on a background thread and does not require redirecting the response through `RA_HandleHttpResults` (which is called every frame).
  * All references to login were removed from `RA_httpthread`. 
* Provides three "server" implementations
  * `ConnectedServer` - talks to a real host
  * `OfflineServer` - will allow for offline functionality (#21)
  * `DisconnectedServer` - `OfflineServer` that converts to a `ConnectedServer` when login succeeds
* Only one implementation will be active at a time and managed by the `ServiceLocator`. The implementation lookup is hidden inside the `Login::Request::Call` method. The caller simply defines the request parameters, and invokes either Call or CallAsync and gets back the response object.
  * More implementations may be added in the future to support unit tests and the v2 site.
* `ConnectedServer` provides generic error handling for things like network errors and invalid JSON so it doesn't have to be duplicated by every API call. Most current calls don't have any error handling. The caller still has to check for success, but will be provided a generic error message if something went wrong, rather than having to handle every case themselves.


src/RA_Core.cpp
```diff
@@ -1005,14 +1003,25 @@ API void CCONV _RA_InvokeDialog(LPARAM nID)
...
+                _RA_RebuildMenu();
 
-            ra::ui::viewmodels::MessageBoxViewModel::ShowInfoMessage(L"You are now logged out.");
-            _RA_RebuildMenu();
+                ra::ui::viewmodels::MessageBoxViewModel::ShowInfoMessage(L"You are now logged out.");
+            }
+            else
+            {
+                ra::ui::viewmodels::MessageBoxViewModel::ShowErrorMessage(L"Logout failed", ra::Widen(response.ErrorMessage));
+            }
```
Jamiras: This is a pretty straight-forward example. Include the header for the API, define a request object and populate the parameters (Logout doesn't have any), then make the call. Check for success and proceed or report the error. The Login calls below are better examples.

src/RA_Dlg_Login.cpp
```diff
@@ -34,76 +38,46 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
...
-                    rapidjson::Document doc;
-                    if (RAWeb::DoBlockingRequest(RequestLogin, args, doc))
+                    const ra::api::Login::Response response = request.Call();
+
+                    if (response.Succeeded())
                     {
-                        std::string sResponse;
-                        std::string sResponseTitle;
-
-                        if (doc["Success"].GetBool())
```
Jamiras: handling Success is now part of the generic response processing.

src/RA_Dlg_Login.cpp
```diff
@@ -34,76 +38,46 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
...
-                        {
-                            const std::string& sUser = doc["User"].GetString();
-                            const std::string& sToken = doc["Token"].GetString();
-                            const unsigned int nPoints = doc["Score"].GetUint();
-                            const unsigned int nUnreadMessages = doc["Messages"].GetUint();
```
Jamiras: extracting the JSON values is done in the API call, which ensures the fields are handled consistently no matter where the API is called from. In this case, Score and Messages are optional and this code doesn't check for that.

src/RA_Dlg_Login.cpp
```diff
@@ -34,76 +38,46 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
...
                     else
                     {
-                        if (!doc.HasParseError() && doc.HasMember("Error"))
-                        {
-                            MessageBox(hDlg, NativeStr(std::string("Server error: ") + std::string(doc["Error"].GetString())).c_str(), TEXT("Error!"), MB_OK);
-                        }
-                        else
-                        {
-                            MessageBox(hDlg, TEXT("Unknown error connecting... please try again!"), TEXT("Error!"), MB_OK);
-                        }
```
Jamiras: more error handling that will be done generically when processing the response.

src/RA_User.cpp
```diff
@@ -57,27 +61,47 @@ LocalRAUser::LocalRAUser(const std::string& sUser) :
...
+            ra::api::Login::Response response = request.Call();
+            HandleLoginResponse(*this, response);
         }
         else
         {
-            AttemptSilentLogin();
+            request.CallAsync([this](const ra::api::Login::Response& response)
+            {
+                HandleLoginResponse(*this, response);
+            });
```
Jamiras: Here you can see examples of both synchronous and asynchronous logins.

src/RA_User.cpp
```diff
@@ -86,36 +110,17 @@ void LocalRAUser::AttemptLogin(bool bBlocking)
...
+    request.Username = Username();
+    request.ApiToken = Token();
+    request.CallAsync([this](const ra::api::Login::Response& response)
     {
-        MessageBox(nullptr, NativeStr(doc["Error"].GetString()).c_str(), TEXT("Login Failed"), MB_OK);
-    }
-    else
-    {
-        MessageBox(nullptr, TEXT("Login failed, please login again."), TEXT("Login Failed"), MB_OK);
-    }
```
Jamiras: Parsing of response and error handling was largely duplicated in the login dialog. Now that's all in the `ConnectedServer` helper functions.

src/api/ApiCall.cpp
```diff
@@ -0,0 +1,26 @@
...
+namespace api {
+
+static ra::api::IServer& Server() noexcept
+{
+    return ra::services::ServiceLocator::GetMutable<ra::api::IServer>();
+}
+
+Login::Response Login::Request::Call() const noexcept
+{
+    return Server().Login(*this);
```
Jamiras: The `Call` method implementation redirects the request parameters to the current `IServer` implementation. By putting the implementation in this file, the individual API header files only have to know about their singular API. `IServer` knows about all APIs.

src/api/ApiCall.hh
```diff
@@ -0,0 +1,143 @@
...
+
+    template<class TRequest, class TCallback>
+    static void CallAsync(const TRequest& request, TCallback&& callback)
+    {
+        ra::services::ServiceLocator::GetMutable<ra::services::IThreadPool>().RunAsync([request, callback = std::move(callback)]
+            {
+                callback(request.Call());
+            });
+    }
+};
```
Jamiras: `ApiRequestBase` provides a generic `CallAsync` that dispatches the specific `Call` method to a background thread. There are no common fields on `ApiRequestBase`

src/api/ApiCall.hh
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /// <summary>
+    /// Gets whether the API call was considered unsuccessful.
+    /// </summary>
+    /// <returns><c>true</c> if the call was unsuccessful, <c>false</c> if not.</returns>
+    bool Failed() const noexcept
+    {
+        return (Result == ApiResult::Error || Result == ApiResult::Unsupported || Result == ApiResult::Failed);
+    }
```
Jamiras: `ApiResponseBase` provides a `Result` and `ErrorMessage` for generic use, and helper methods for classifying `Result`.

src/api/ApiCall.hh
```diff
@@ -0,0 +1,143 @@
...
+};
+
+// ---- template for generating a subclass file (replace XXXX with API name) ----
+// NOTE: Implementation of Call method will be in ApiCall.cpp to avoid references IServer and all of its dependencies
```
Jamiras: The _following_ template can be used for any new APIs - just replace XXXX with the API name, and flesh out the request and response parameters. Then implement the `Call` redirect in `ApiBase` and the actual implementation in `ConnectedServer` and/or the other `IServer` implementations.

src/api/impl/ConnectedServer.cpp
```diff
@@ -0,0 +1,154 @@
...
+        RA_LOG_INFO("%s Request: %s", Login::Name(), sPostData.c_str());
+        AppendUrlParam(sPostData, "p", request.Password);
+    }
+    httpRequest.SetPostData(sPostData);
+
+    ra::services::Http::Response httpResponse = httpRequest.Call();
+
+    Login::Response response;
+    rapidjson::Document document;
+    if (GetJson(Login::Name(), httpResponse, response, document))
```
Jamiras: This is where all the generic error handling occurs.

src/api/impl/ConnectedServer.cpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+        pResponse.Result = ApiResult::Failed;
+        RA_LOG_ERR("-- %s Error: Success=false", sApiName);
+        return false;
+    }
+
+    return true;
+}
+
+static void GetRequiredJsonField(_Out_ std::string& sValue, _In_ const rapidjson::Document& pDocument, _In_ const char* const sField, _Inout_ ApiResponseBase& response) noexcept
```
SyrianBallaS: I think you have to initialize `sValue`.

src/api/impl/ConnectedServer.cpp
```diff
@@ -0,0 +1,154 @@
...
+        if (response.ErrorMessage.empty())
+            response.ErrorMessage = std::string(sField) + " not found in response";
+    }
+    else
+    {
+        sValue = pDocument[sField].GetString();
+    }
+}
+
+static void GetOptionalJsonField(_Out_ unsigned int& value, _In_ const rapidjson::Document& pDocument, _In_ const char* sField, _In_ unsigned int nDefaultValue = 0) noexcept
```
SyrianBallaS: I think you have to initialize `value`.

src/api/impl/ConnectedServer.hh
```diff
@@ -0,0 +1,25 @@
...
+#include "ServerBase.hh"
+
+namespace ra {
+namespace api {
+namespace impl {
+
+class ConnectedServer : public ServerBase
+{
+public:
+    ConnectedServer(const std::string& sHost) : m_sHost(sHost) {}
```
SyrianBallaS: You should put the `explicit`. `noexcept` probably wouldn't work here since the string is copied.

src/api/impl/DisconnectedServer.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "services\ServiceLocator.hh"
+
+namespace ra {
+namespace api {
+namespace impl {
+
+Login::Response DisconnectedServer::Login(const Login::Request& request) noexcept
+{
+    // use the normal ServerApi to attempt to connect
+    std::unique_ptr<ConnectedServer> serverApi(new ConnectedServer(m_sHost));
```
SyrianBallaS: You could use `std::make_unique`;

src/api/impl/DisconnectedServer.hh
```diff
@@ -0,0 +1,26 @@
...
+#include "OfflineServer.hh"
+
+namespace ra {
+namespace api {
+namespace impl {
+
+class DisconnectedServer : public OfflineServer
+{
+public:
+    DisconnectedServer(const std::string& sHost) : m_sHost(sHost) {}
```
SyrianBallaS: Should be explicit.

src/api/impl/OfflineServer.cpp
```diff
@@ -0,0 +1,25 @@
...
+Login::Response OfflineServer::Login(const Login::Request& request) noexcept
+{
+    Login::Response response;
+    response.Result = ApiResult::Success;
+    response.Username = request.Username;
+    response.ApiToken = "offlineToken";
+    return std::move(response);
+}
+
+Logout::Response OfflineServer::Logout(_UNUSED const Logout::Request& request) noexcept
```
SyrianBallaS: Not a big deal but you should probably comment out the names of unused variables.

src/services/Http.cpp
```diff
@@ -65,7 +65,7 @@ void Http::UrlEncodeAppend(std::string& sOutput, const std::string& sInput)
     if (sOutput.capacity() < nNeeded)
         sOutput.reserve(nNeeded);
 
-    for (auto c : sInput)
+    for (unsigned char c : sInput)
```
SyrianBallaS: `c` is never changed, it should be `const`.

src/services/Http.hh
```diff
@@ -10,10 +10,16 @@ namespace services {
...
+    {
+        OK = 200,
+        NotFound = 404,
+    };
+
     class Response
     {
     public:
-        explicit Response(int nStatusCode, const std::string& sResponse) noexcept
+        explicit Response(Http::StatusCode nStatusCode, const std::string& sResponse) noexcept
```
SyrianBallaS: This might not be able to be `noexcept` because `std::string`'s copy constructor is not `noexcept`.

src/services/Http.hh
```diff
@@ -27,15 +33,15 @@ public:
         /// <summary>
         /// Gets the HTTP status code returned from the server.
         /// </summary>
-        unsigned int StatusCode() const { return m_nStatusCode; }
+        Http::StatusCode StatusCode() const { return m_nStatusCode; }
```
SyrianBallaS: This one can be `noexcept`.

src/services/Http.hh
```diff
@@ -27,15 +33,15 @@ public:
...
 
         /// <summary>
         /// Gets the content returned from the server.
         /// </summary>
         const std::string& Content() const { return m_sResponse; }
```
SyrianBallaS: This one can be noexcept.

src/services/Http.hh
```diff
@@ -27,15 +33,15 @@ public:
...
 
     private:
-        unsigned int m_nStatusCode;
+        Http::StatusCode m_nStatusCode;
```
SyrianBallaS: You should probably initialize this.

tests/RA_UnitTestHelpers.h
```diff
@@ -35,6 +39,11 @@ template<> static std::wstring ToString<Condition::ConditionType>(const Conditio
     return ra::Widen(CONDITIONTYPE_STR[(int)t]);
 }
 
+template<> static std::wstring ToString<ra::services::Http::StatusCode>(const ra::services::Http::StatusCode& t)
```
SyrianBallaS: I think making template specializations `static` is non-standard.

tests/api/ConnectedServer_Tests.cpp
```diff
@@ -0,0 +1,272 @@
...
+        MockHttpRequester mockHttp([](const Http::Request& request)
+        {
+            return Http::Response(Http::StatusCode::OK, "{\"Success\":true}");
+        });
+
+        ConnectedServer server("host.com");
+
+        Login::Request request;
+        request.Username = "User";
+        request.Password = "pa$$w0rd";
```
SyrianBallaS: Couldn't this just be initialized at once?

tests/mocks/MockHttpRequester.hh
```diff
@@ -0,0 +1,39 @@
...
+#include "services\ServiceLocator.hh"
+
+namespace ra {
+namespace services {
+namespace mocks {
+
+class MockHttpRequester : public IHttpRequester
+{
+public:
+    explicit MockHttpRequester(std::function<Http::Response(const Http::Request&)>&& fHandler) noexcept
```
SyrianBallaS: I don't think this can be `noexcept`.

SyrianBallaS: It's not that the compiler wouldn't complain, but std::function's copy constructor might throw an exception. It's not a big deal.

tests/api/ConnectedServer_Tests.cpp
SyrianBallaS: Don't worry about it then.

tests/RA_UnitTestHelpers.h
SyrianBallaS: Well, I've never heard that before. Since Windows APIs before WinRT are basically non-standard it depends on "undefined" behavior. Meaning that it's undefined in C++ but defined on Windows. If we really do need them to be static we can try suppressing the warnings.

src/api/ApiCall.hh
```diff
@@ -0,0 +1,143 @@
...
+
+struct ApiRequestBase
+{
+protected:
+    ApiRequestBase() noexcept = default;
+    ~ApiRequestBase() noexcept = default;
+    ApiRequestBase(const ApiRequestBase&) = default;
+    ApiRequestBase& operator=(const ApiRequestBase&) = default;
+    ApiRequestBase(ApiRequestBase&&) noexcept = default;
+    ApiRequestBase& operator=(ApiRequestBase&&) noexcept = default;
```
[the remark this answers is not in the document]
SyrianBallaS: No, if the forced compiler generated constructor is `noexcept` and/or `constexpr` then the implicitly generated ones are identical.

The `define or delete them all` rule is if you do do `=delete` or make a constructor/assignment user-defined instead of compiler generated (has a body instead of `=default`).

src/RA_Dlg_Login.cpp
```diff
@@ -34,76 +38,46 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
...
                     {
-                        MessageBox(nullptr, TEXT("Username/password not valid! Please check and reenter"), TEXT("Error!"), MB_OK);
+                        ra::ui::viewmodels::MessageBoxViewModel::ShowErrorMessage(L"Password is required.");
                         return TRUE;
                     }
 
-                    PostArgs args;
-                    args['u'] = ra::Narrow(sUserEntry);
-                    args['p'] = ra::Narrow(sPassEntry);		//	Plaintext password(!)
+                    ra::api::Login::Request request;
```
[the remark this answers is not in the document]
SyrianBallaS: alright.

src/RA_Dlg_Login.cpp
```diff
@@ -34,76 +38,46 @@ INT_PTR CALLBACK RA_Dlg_Login::RA_Dlg_LoginProc(HWND hDlg, UINT uMsg, WPARAM wPa
             {
                 case IDOK:
                 {
-                    BOOL bValid = TRUE;
                     TCHAR sUserEntry[64];
```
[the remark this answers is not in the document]
SyrianBallaS: Sweet.

src/api/ApiCall.hh
```diff
@@ -0,0 +1,143 @@
...
+    struct Response : ApiResponseBase
+    {
+        // [RESPONSE PARAMETERS]
+
+        Response() noexcept = default;
+        ~Response() noexcept = default;
+        Response(const Response&) = default;
+        Response& operator=(const Response&) = default;
+        Response(Response&&) noexcept = default;
+        Response& operator=(Response&&) noexcept = default;
```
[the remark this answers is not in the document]
SyrianBallaS: It's fine.

src/api/IServer.hh
```diff
@@ -0,0 +1,25 @@
...
+#include "api/Login.hh"
+#include "api/Logout.hh"
+
+namespace ra {
+namespace api {
+
+class IServer
+{
+public:
+    virtual const char* Name() const noexcept = 0;
```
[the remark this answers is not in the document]
SyrianBallaS: Because this is an abstract interface, which seems to be a base class. Which needs a default virtual destructor. The other ones weren't abstract.

src/api/impl/ConnectedServer.cpp
SyrianBallaS: Sweet

src/api/impl/ConnectedServer.cpp
SyrianBallaS: My mistake, I apologize.

tests/mocks/MockHttpRequester.hh
SyrianBallaS: The purpose is to identify where exception handling will need to be targeted.

One of the rulesets; `Function Rules`, will warn on this but hasn't been applied yet (done but holding off on doing a PR for it).
Example:
Also: `fHandler `was never moved but it's move constructor might throw.
`explicit MockHttpRequester(std::function<Http::Response(const Http::Request&)>&& fHandler) noexcept(std::is_nothrow_move_constructible_v<std::function<Http::Response(const Http::Request&)>>);`

I guarantee that it will not be noexcept (hover over the constructor name and the `noexcept `will not be there) if you do that since it's not. Don't worry about it now, we'll handle it when the time comes.

tests/RA_UnitTestHelpers.h
Jamiras: Here's the template methods we're specializing. These methods are not contained within a class.
```
	// User should provide a specialization of below ToString() template to use equality asserts in test code.
	// std::wstring ToString(const Q& q) is used for converting type Q in  user readable std::wstring value. 
	// Converted value is emitted in Assert failure message. Specialization for basic types has been provided below.
	template <typename Q> static std::wstring ToString(const Q& q) { static_assert(false, "Test writer must define specialization of ToString<const Q& q> for your class " __FUNCSIG__ "."); }
	template <typename Q> static std::wstring ToString(const Q* q) { static_assert(false, "Test writer must define specialization of ToString<const Q* q> for your class " __FUNCSIG__ "."); }
	template <typename Q> static std::wstring ToString(Q* q)       { static_assert(false, "Test writer must define specialization of ToString<Q* q> for your class " __FUNCSIG__ "."); }
```
However, it does compile if I remove the static. I couldn't find any documentation stating which is preferred (or non-standard). What makes you say that it's non-standard?

(I meant to post this a couple days ago, but apparently forgot to hit the button - as this is duplicating pre-existing functions, we can address in a later PR if necessary)

SyrianBallaS: If you compile with /W4 it'll say non-standard extension used.

SyrianBallaS: Don't worry about it right now, I have it done (/W4 and conformance) in #161 but can't figure out how to resolve one of the problems since you didn't like how I handled it last time.

PR #170, opened 2018-11-05
PopupMessageType and small fix

SyrianBallaS:
Basically made `PopupMessageType `into an `enum class `as part of `MessagePopup`. For `ra::StringPrintf `to work correctly with `std::wstring` the capacity had to be larger. By default I put 512 maybe it might need to be bigger but it seems big enough for now.


src/RA_AchievementPopup.h
```diff
@@ -33,50 +31,46 @@ class MessagePopup
     {
     }
 
-    MessagePopup(const std::string& sTitle, const std::string& sSubtitle, PopupMessageType nMsgType = PopupInfo) :
+    explicit MessagePopup(const std::string& sTitle, const std::string& sSubtitle = "", Type nMsgType = Type::Info) :
```
SyrianBallaS: Sometimes the subtitle was empty.

src/RA_StringUtils.h
```diff
@@ -45,6 +45,7 @@ _NODISCARD inline auto StringPrintf(_In_z_ _Printf_format_string_ const CharT* c
     }
     else if constexpr(std::is_same_v<CharT, wchar_t>)
     {
+        sFormatted.reserve(512); // The number is arbitrary but the default capacity is too small (7)
```
SyrianBallaS: Had to explicitly reserve more memory for `std::wstring`, or the return would be `-1` for `std::vswprintf`.

SyrianBallaS:
Need to be careful with this one, closing it.

PR #171, opened 2018-11-07
Custom Deleters, Specialized std::unique_ptrs

SyrianBallaS:
**Summary**
- Function objects used as deleters.
- Type aliases for `std::unique_ptr` for simplicity, implementation details are in a `detail` namespace.
   - Reason is that using the structs directly will make it look complicated
- Bug fix for appending Integration version if not on a branch.
- Bug fix for buffer too small in `ra::StringPrintf` when using wide-character strings.
- Applied the RAII style on most `HANDLE`s (except the ones in `RA_httpthread`), the global `HINSTANCE`s , the global `HWND`, and the `FILE*`s.
- Made a C11 style of `fopen_s` to prevent the need of using `std::fopen`. It is somewhat different.
  - It uses the restricted pointer optimization (meaning the pointer never gets shallow copied/referenced)

So I've been trying to find the cleanest way to do this for awhile, but never came forward with this until now. There's also two bugs that were fixed, they don't always happen but they _could_ happen. Made sure they never happen. All the wrapped handles I think we need. To prevent the line count from going too high (it's currently 409+ and 184- lines),  I only applied some of them. The custom deleters are in a new file `ra_deleters.h`.

I'll leave some review comments to explain some stuff.


src/RA_AchievementOverlay.cpp
```diff
@@ -541,11 +541,11 @@ void AchievementOverlay::DrawAchievementsPage(HDC hDC, int nDX, int nDY, const R
...
-                    &g_pActiveAchievements->GetAchievement(nAchIdx),	//	pAch
-                    nDX,												//	X
-                    (nAchTopEdge + (i*nAchSpacing)),				//	Y
-                    bSelected,											//	Selected
-                    TRUE);
+                                &g_pActiveAchievements->GetAchievement(nAchIdx), // pAch
+                                nDX,                                             // X
+                                (nAchTopEdge + (i*nAchSpacing)),                 // Y
+                                bSelected,                                       // Selected
+                                TRUE);
```
SyrianBallaS: Noticed some tabs so I got rid of them.

src/RA_Core.cpp
```diff
@@ -40,9 +40,9 @@
 std::wstring g_sHomeDir;
 std::string g_sROMDirLocation;
 
-HMODULE g_hThisDLLInst = nullptr;
-HINSTANCE g_hRAKeysDLL = nullptr;
-HWND g_RAMainWnd = nullptr;
+ra::InstanceH g_hRAKeysDLL;
+ra::InstanceH g_hThisDLLInst;
+ra::AppWndH g_RAMainWnd;
```
SyrianBallaS: The default constructor of `std::unique_ptr`, even with a custom deleter is `constexpr` (function pointers/objects are locked in the same location) so this it'll also get rid of warnings in the long run. All this stuff gets deallocated when w/e app is integrated ends.

src/RA_Core.cpp
```diff
@@ -56,10 +56,23 @@ inline static unsigned int g_nProcessTimer{ 0U };
...
-        g_hThisDLLInst = hModule;
+    // exceptions can happen here, they must not be allowed to leave
+    try
+    {
+        if (dwReason == DLL_PROCESS_ATTACH)
+        {
+            // resetting it more than once will cause a deadlock or access violation
+            // Library will be freed when the app ends
+            if (!g_hThisDLLInst)
+                g_hThisDLLInst.reset(hModule);
```
SyrianBallaS: This is really important, it can only be assigned once since it's a unique_ptr. I'd personally prefer not to deal with cycles, `std::shared_ptr` is potentially slow (not always but definitely slower than `std::unique_ptr` if you don't use `std::weak_ptr` to break the cycles). Plus `std::shared_ptr` doesn't support custom deleters.

src/RA_Core.cpp
```diff
@@ -56,10 +56,23 @@ inline static unsigned int g_nProcessTimer{ 0U };
...
+        }
 
-    ra::services::Initialization::RegisterCoreServices();
+        ra::services::Initialization::RegisterCoreServices();
+    }
+    catch (const std::exception& e)
+    {
+        RA_LOG_ERR("Exception: %s", e.what());
```
SyrianBallaS: I put this in here as a safe-guard as the entry-point of a DLL should not exit via exception.

src/RA_Core.cpp
```diff
@@ -734,52 +748,52 @@ API int CCONV _RA_HandleHTTPResults()
 //	Following this function, an app should call AppendMenu to associate this submenu.
 API HMENU CCONV _RA_CreatePopupMenu()
 {
-    HMENU hRA = CreatePopupMenu();
-    HMENU hRA_LB = CreatePopupMenu();
+    // Callers need to claim ownership of hRA, hRA_LB is automatically deleted
+    ra::MenuH hRA{ ::CreatePopupMenu() };
+    ra::MenuH hRA_LB{ ::CreatePopupMenu() };
```
SyrianBallaS: The Initlizer syntax is required. `std::unique_ptr'`s constructors are explicit. Notice how we didn't need to specify the delete function.
`explicit unique_ptr( pointer p ) noexcept;`

src/RA_Core.cpp
```diff
@@ -734,52 +748,52 @@ API int CCONV _RA_HandleHTTPResults()
...
+        AppendMenu(hRA.get(), MF_STRING, IDM_RA_FILES_LOGIN, TEXT("&Login to RA"));
     }
 
-    AppendMenu(hRA, MF_SEPARATOR, 0U, nullptr);
-    AppendMenu(hRA, MF_STRING, IDM_RA_FILES_CHECKFORUPDATE, TEXT("&Check for Emulator Update"));
+    AppendMenu(hRA.get(), MF_SEPARATOR, 0U, nullptr);
+    AppendMenu(hRA.get(), MF_STRING, IDM_RA_FILES_CHECKFORUPDATE, TEXT("&Check for Emulator Update"));
 
-    return hRA;
+    return hRA.release(); // must be released
```
SyrianBallaS: If we don't release it, the menu will not appear. Behavior-wise, this is exactly how it was originally except the other handle is deallocated. But the leaderboard menu still appears since `hRA `owns `hRA_LB`. This just makes it clear that `hRA` is not owned anymore. If you think this causes a leak, then what was here before also causes a leak.

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -47,8 +48,8 @@ bool ListFiles(std::string path, std::string mask, std::deque<std::string>& rFil
         directories.pop();
 
         WIN32_FIND_DATA ffd;
-        HANDLE hFind = FindFirstFile(NativeStr(spec).c_str(), &ffd);
-        if (hFind == INVALID_HANDLE_VALUE)
+        ra::FindFileH hFind{ ::FindFirstFile(NativeStr(spec).c_str(), &ffd) };
+        if (hFind.get() == INVALID_HANDLE_VALUE)
```
SyrianBallaS: As a test, I put the `.nes` and `.pas` (headerless ROM) extensions for RANes and it went pass this condition. I put it back and as expected the handle was invalid.

src/RA_httpthread.cpp
```diff
@@ -118,10 +118,12 @@ static void AppendIntegrationVersion(_Inout_ std::string& sUserAgent)
...
                                        RA_INTEGRATION_VERSION_MODIFIED));
-
-    _CONSTANT_LOC posFound{ std::string_view{ RA_INTEGRATION_VERSION_PRODUCT }.find('-') };
-    constexpr std::string_view sAppend{ RA_INTEGRATION_VERSION_PRODUCT };
-    sUserAgent.append(sAppend, posFound);
+    
+    if constexpr (_CONSTANT_LOC pos{ std::string_view{ RA_INTEGRATION_VERSION_PRODUCT }.find('-') }; pos != std::string_view::npos)
+    {
+        constexpr std::string_view sAppend{ RA_INTEGRATION_VERSION_PRODUCT };
+        sUserAgent.append(sAppend, pos);
```
SyrianBallaS: If not on a branch, this line would throw an exception. So it's in an if statement.

src/ra_deleters.h
```diff
@@ -0,0 +1,211 @@
...
+// Handles have to types the way they are or restrict won't work
+// We can remove the const in front if setting to nullptr is desired
+
+struct CFileDeleter { void operator()(_In_ FILE* const __restrict fp) const noexcept { std::fclose(fp); } };
+struct MenuDeleter { void operator()(_In_ HMENU__* const __restrict hMenu) const noexcept { ::DestroyMenu(hMenu); } };
+
+// can also be used for HMODULE
+struct InstanceDeleter
+{
+    void operator()(_In_ HINSTANCE__* const  __restrict hInst) const noexcept
```
SyrianBallaS: To make a true "in" param, the const has to be in front. The object is deleted but not the pointer. If we want to null them out we can get rid of the const.

If I didn't use their struct names, it won't compile. 
`HINSTANCE __restrict hInst; -> resolves to: restrict HINSTANCE. `which is wrong
The way it is resolves correctly: `HINSTANCE__* const  restrict hInst`

src/ra_deleters.h
```diff
@@ -0,0 +1,211 @@
...
+};
+
+// Can be used with HLOCAL as well since there is no local heap
+struct GlobalDeleter { void operator()(_In_ void* const __restrict hMem) const noexcept { ::GlobalFree(hMem); } };
+
+enum class IconHandle
+{
+    Icon,
+    Cursor
+};
```
SyrianBallaS: Because some handles have the same struct name, these enums were necessary.

src/ra_deleters.h
```diff
@@ -0,0 +1,211 @@
...
+using BitmapH  = std::unique_ptr<HBITMAP__, detail::GdiDeleter<detail::GdiType::Bitmap>>;
+using BrushH   = std::unique_ptr<HBRUSH__, detail::GdiDeleter<detail::GdiType::Brush>>;
+using FontH    = std::unique_ptr<HFONT__, detail::GdiDeleter<detail::GdiType::Font>>;
+using PaletteH = std::unique_ptr<HPALETTE__, detail::GdiDeleter<detail::GdiType::Palette>>;
+using RgnH     = std::unique_ptr<HRGN__, detail::GdiDeleter<detail::GdiType::Rgn>>;
+using GdiObjH  = std::unique_ptr<void, detail::GdiDeleter<detail::GdiType::Obj>>;
+
+using FileH     = std::unique_ptr<void, detail::NTKernelDeleter<detail::NTKernelObjType::File>>;
+using FindFileH = std::unique_ptr<void, detail::NTKernelDeleter<detail::NTKernelObjType::FindFile>>;
+using HandleH   = std::unique_ptr<void, detail::NTKernelDeleter<detail::NTKernelObjType::Object>>;
```
SyrianBallaS: Typing out the right side each time and specifying the deleter each time on construction would be tedious. These type aliases attempt to make it simpler.

src/ra_utility.h
```diff
@@ -157,6 +157,32 @@ template<typename Enum, class = std::enable_if_t<std::is_enum_v<Enum>>> _CONSTAN
 operator&=(_Inout_ Enum& a, _In_ Enum b) noexcept { return (a = a & b); }
 
 } // namespace bitwise_ops
+
+// So we don't have to #define _CRT_SECURE_NO_WARNINGS, C99/11 style (no-alias optimization)
+// To use instead of std::fopen for unique_ptr
+template<typename CharT, typename = std::enable_if_t<is_char_v<CharT>>>
+_NODISCARD FILE* __cdecl fopen_s(_In_z_ const CharT* __restrict filename,
+                                 _In_z_ const CharT* __restrict mode,
+                                 _In_opt_     FILE*  __restrict streamptr = nullptr)
```
SyrianBallaS: Somewhat like C11's fopen_s. This is how it looks like in C11...
```
errno_t fopen_s(FILE *restrict *restrict streamptr, 
                const char *restrict filename, 
                const char *restrict mode);
```

I changed it a bit to suit our purposes better. The MSVC versions are still called.

src/RA_Core.cpp
Jamiras: Library should be freed by the client, not the library. We are simply caching our own base pointer (see https://docs.microsoft.com/en-us/windows/desktop/dlls/dllmain)
> A handle to the DLL module. The value is the base address of the DLL. 

Deleting ourself, even if it is when we're being unloaded, can lead to undefined behavior.

src/RA_Core.cpp
```diff
@@ -68,7 +81,8 @@ static void InitCommon(HWND hMainHWND, /*enum EmulatorID*/int nEmulatorID, const
 {
     // initialize global state
     g_EmulatorID = static_cast<EmulatorID>(nEmulatorID);
-    g_RAMainWnd = hMainHWND;
+    if (!g_RAMainWnd)
+        g_RAMainWnd.reset(hMainHWND);
```
Jamiras: So, if for whatever reason, the client calls Init twice, we'll destroy the window associated to the first handle they passed us? This is just a reference to a window owned by the client. We shouldn't be calling `DestroyWindow` on something we didn't create.

SyrianBallaS:
So which ones should I keep? Or should we just not do this?

src/RA_Core.cpp
SyrianBallaS: makes sense.

src/RA_Core.cpp
```diff
@@ -734,52 +748,52 @@ API int CCONV _RA_HandleHTTPResults()
 //	Following this function, an app should call AppendMenu to associate this submenu.
 API HMENU CCONV _RA_CreatePopupMenu()
 {
-    HMENU hRA = CreatePopupMenu();
-    HMENU hRA_LB = CreatePopupMenu();
+    // Callers need to claim ownership of hRA, hRA_LB is automatically deleted
```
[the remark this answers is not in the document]
SyrianBallaS: I thought I did enough research but I guess I missed that part.

src/RA_Core.h
```diff
@@ -6,13 +6,14 @@
...
 #include "Exports.hh"
+#include "ra_handles.h"
 
 //	Non-exposed:
 extern std::wstring g_sHomeDir;
 
-extern HINSTANCE g_hRAKeysDLL;
-extern HMODULE g_hThisDLLInst;
-extern HWND g_RAMainWnd;
+extern ra::InstanceH g_hRAKeysDLL;
```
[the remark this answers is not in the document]
SyrianBallaS: sure

Jamiras:
Honestly, the only change I'd like to keep is this:https://github.com/RetroAchievements/RAIntegration/pull/171/files#diff-5f269daa1d43eae5f0106751651f3a31R122

which isn't even related to the intent of the PR.

SyrianBallaS:
It should be included in other prs I did that originally in my master branch. I'll close this for now.

PR #172, opened 2018-11-08
ClangFormat (based off of CONTRIBUTING)

SyrianBallaS:
So I've recently learned about this thing called clang-format which is kinda like EditorConfig but lets you do a lot more customization (closer to the .NET EditorConfig). The .clang-format file is used automatically in all formatting scenarios (by default such as typing ";" ) or by doing "format document/selection" (no plugin needed, VS2017 is supposedly a requirement).

There was one thing I could not figure out how to do as it doesn't seem to have an option for it (had to make it fully custom for the brace-break part as most are K&R-derived) was to have an empty line after the namespace brace and before the end. Hopefully it's not a big deal.

As a "demo", I just applied it to all the files in the "Helpers" filter, had to manually realign some stuff in ra_fwd.h though.


SyrianBallaS:
I'm not sure about some of the settings in the .clang-format file. They can change if need be.
Explanations for them are here: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

src/RA_StringUtils.h
```diff
@@ -45,6 +45,9 @@ _NODISCARD inline auto StringPrintf(_In_z_ _Printf_format_string_ const CharT* c
     }
     else if constexpr(std::is_same_v<CharT, wchar_t>)
     {
+        // needs to be more than default (7 bytes, "3.5" chars?) (4096 bytes, 2048 chars)
+        // must be a bug in MSVC
+        sFormatted.reserve(4096U);
```
Jamiras: This seems excessive. Most strings won't be more than 40 characters. The point of the first swprintf is to determine how much space to allocate for the second call. If this is really necessary, recommend starting at 32 characters.

Also note that reserve takes the number of elements, not number of bytes, so you're actually allocating 8K here.

SyrianBallaS: alright I'll see if 32 will work instead.

Jamiras: That's probably because `vswprintf` doesn't behave like `vsnprintf`:
http://www.cplusplus.com/reference/cwchar/vswprintf/

> A negative number is returned on failure, **including when the resulting string to be written to ws would be longer than n characters.**

http://www.cplusplus.com/reference/cstdio/vsnprintf/

> The number of characters that would have been written if n had been sufficiently large, not counting the terminating null character.

Jamiras: so, for vswprintf, we'll just have to start at some value (I'm fine with 64) and keep doubling it until we find something big enough.

Jamiras: This still needs to be addressed before I'll merge. It looks like you tried in #161, and I left feedback on that implementation there.

PR #190, opened 2018-11-13
Enum to Enum Class (AlignType)

SyrianBallaS:
All in the title, put in some minor optimizations as well. Will explain them after this PR is submitted.

References:
- https://en.cppreference.com/w/c/language/restrict
- https://docs.microsoft.com/en-us/cpp/cpp/extension-restrict?view=vs-2017


src/RA_Defs.h
```diff
@@ -79,39 +79,42 @@ class RARect : public RECT
 class ResizeContent
 {
 public:
-    enum AlignType
+    enum class AlignType
     {
-        NO_ALIGN,
```
SyrianBallaS: `NO_ALIGN` was never used, not even in a `[[fallthrough]]` statement.

src/RA_Defs.h
```diff
@@ -79,39 +79,42 @@ class RARect : public RECT
...
 public:
-    HWND hwnd;
-    POINT pLT;
-    POINT pRB;
-    AlignType nAlignType;
-    int nDistanceX;
-    int nDistanceY;
-    bool bResize;
-
-    ResizeContent(HWND parentHwnd, HWND contentHwnd, AlignType newAlignType, bool isResize)
```
SyrianBallaS: We can get `parentHwnd` via `::GetParent`

src/RA_Defs.h
```diff
@@ -79,39 +79,42 @@ class RARect : public RECT
...
+        assert(check != 0);
 
-        pLT.x = rect.left;	pLT.y = rect.top;
-        pRB.x = rect.right; pRB.y = rect.bottom;
+        pLT ={ rect.left, rect.top };
+        pRB ={ rect.right, rect.bottom };
 
-        ScreenToClient(parentHwnd, &pLT);
-        ScreenToClient(parentHwnd, &pRB);
+        HWND__* const restrict parentHwnd = ::GetParent(contentHwnd);
```
SyrianBallaS: The restrict qualifier asserts disjoint storage (no-alias). Meaning no other pointer points the same address in the same scope. It optimizes the generated assembly code. The purpose is to get it to have the same performance as FORTRAN.

The` HWND` parameter above could not be restrict qualified because it gets aliased by `hwnd `(same address). If this were C, it would cause undefined behavior if it were restrict qualified but MSVC just ignores it if it was wrong.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -2252,37 +2252,23 @@ void GenerateResizes(HWND hDlg)
...
-        GetDlgItem(hDlg, IDC_RA_PASTECOND), ResizeContent::ALIGN_BOTTOM, FALSE));
-    vDlgAchEditorResize.push_back(ResizeContent(hDlg,
-        GetDlgItem(hDlg, IDC_RA_MOVECONDUP), ResizeContent::ALIGN_BOTTOM, FALSE));
-    vDlgAchEditorResize.push_back(ResizeContent(hDlg,
-        GetDlgItem(hDlg, IDC_RA_MOVECONDDOWN), ResizeContent::ALIGN_BOTTOM, FALSE));
-
-    vDlgAchEditorResize.push_back(ResizeContent(hDlg,
-        GetDlgItem(hDlg, IDCLOSE), ResizeContent::ALIGN_BOTTOM_RIGHT, FALSE));
+    using AlignType = ResizeContent::AlignType;
+    vDlgAchEditorResize.emplace_back(::GetDlgItem(hDlg, IDC_RA_LBX_CONDITIONS), AlignType::BottomRight, true);
```
SyrianBallaS: Skips the copy/move step and just perfectly forwards parameters via parameter pack.

src/RA_Defs.h
SyrianBallaS: Just to note the C++ versions of GCC and Clang use `__restrict__` instead of `__restrict`.

PR #194, opened 2018-11-16
Ui bitmap

Jamiras:
* Add offscreen surface and offscreen surface with alpha channel (GDI doesn't implicitly support alpha channel)
* Add factory for generating offscreen surfaces
* Render popup to offscreen surface (with alpha)
* Draw pre-rendered popup to screen during display pass.

Opacity is currently set to 85%:
image

### DIFF
--- a/src/RA_AchievementPopup.h
+++ b/src/RA_AchievementPopup.h
@@ -2,12 +2,9 @@
 #define RA_ACHIEVEMENTPOPUP_H
 #pragma once
 
-#include "RA_Defs.h"
 #include "RA_Interface.h"
 
-#include "ui\ImageReference.hh"
-
-#include "ui\drawing\gdi\ResourceRepository.hh"
+#include "ui\drawing\ISurface.hh"
 
 //	Graphic to display an obtained achievement
 enum PopupMessageType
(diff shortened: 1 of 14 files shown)


src/RA_AchievementPopup.h
```diff
@@ -43,42 +42,42 @@ class MessagePopup
     {
     }
 
-public:
+    ~MessagePopup() = default;
```
SyrianBallaS: This one needs to be noexcept.

src/RA_AchievementPopup.h
```diff
@@ -43,42 +42,42 @@ class MessagePopup
...
+    MessagePopup(const MessagePopup&) = delete;
+    MessagePopup(MessagePopup&&) = default;
```
SyrianBallaS: This one should be `noexcept`. I think you should `delete` the copy assignment since copying is deleting and the move assignment operator since all members are `const`.

src/ui/drawing/ISurface.hh
```diff
@@ -72,10 +73,47 @@ public:
...
     ISurface() noexcept = default;
 };
 
+class ISurfaceFactory
+{
+public:
+    virtual ~ISurfaceFactory() = default;
+    ISurfaceFactory(const ISurfaceFactory&) noexcept = delete;
+    ISurfaceFactory& operator=(const ISurfaceFactory&) noexcept = delete;
+    ISurfaceFactory(ISurfaceFactory&&) noexcept = delete;
```
SyrianBallaS: I'd need to double check this, but I think this move constructor needs to be protected because it's used in a `unique_ptr`. If it still works with this deleted, don't worry about it.

src/ui/drawing/gdi/GDIBitmapSurface.cpp
```diff
@@ -0,0 +1,182 @@
...
+        *nTarget = nBlend;
+        return;
+    }
+
+    // blend each of the RGB values based on the blend pixel's alpha value
+    // do not modify the target pixel's alpha value.
+    UINT8* pTarget = reinterpret_cast<UINT8*>(nTarget);
+    UINT8* pBlend = reinterpret_cast<UINT8*>(&nBlend);
+
+    *pTarget++ = (UINT8)(((UINT32)*pBlend++ * alpha + (UINT32)*pTarget * (256 - alpha)) / 256);
```
SyrianBallaS: Not a big deal, this can be done in another PR. We really shouldn't be using C-Style casts though.

src/ui/drawing/gdi/GDIBitmapSurface.hh
```diff
@@ -0,0 +1,122 @@
...
+        : GDISurface(CreateBitmapHDC(nWidth, nHeight), RECT { 0, 0, nWidth, nHeight })
+    {
+    }
+
+    GDIBitmapSurface(const GDIBitmapSurface&) noexcept = delete;
+    GDIBitmapSurface& operator=(const GDIBitmapSurface&) noexcept = delete;
+    GDIBitmapSurface(GDIBitmapSurface&&) noexcept = delete;
+    GDIBitmapSurface& operator=(GDIBitmapSurface&&) noexcept = delete;
+
+    ~GDIBitmapSurface()
```
SyrianBallaS: Destructors need to be `noexcept` no matter what.

src/ui/drawing/ISurface.hh
Jamiras: The move constructor is being deleted from all of the existing "interface" classes.

src/ui/drawing/gdi/GDIBitmapSurface.cpp
Jamiras: fixed.

src/ui/drawing/gdi/GDIBitmapSurface.hh
Jamiras: ok

PR #196, opened 2018-11-17
Add ToString, AppendString, and BuildString methods

Jamiras:
This provides a variadic template version of `StingPrintf`, as well as more straight-forward `AppendString` and `BuildString` methods. All are built on top of a templated `ToString` method. `AppendString`, `BuildString` and `ToString` have unicode variants: `ToWString`, `AppendWString`, and `BuildWString` - primarily because `ToString` and `BuildString` don't have any sort of identifiable input. `StringPrintf` is still overloaded to support either UTF-8 or Unicode via the `format` parameter.

`ToString` converts pretty much anything to a string, including unicode text.
```
        Assert::AreEqual(std::string("99"), ToString(99));
        Assert::AreEqual(std::string("-3"), ToString(-3));
        Assert::AreEqual(std::string("0"), ToString(0U));
        Assert::AreEqual(std::string("99"), ToString(99U));
        Assert::AreEqual(std::string("4.200000"), ToString(4.2));
        Assert::AreEqual(std::string("Apple"), ToString("Apple"));
        Assert::AreEqual(std::string("Apple"), ToString(std::string("Apple")));
        Assert::AreEqual(std::string("Apple"), ToString(L"Apple"));
        Assert::AreEqual(std::string("Apple"), ToString(std::wstring(L"Apple")));
```

`BuildString` combines a series of items using `ToString`:
```
        std::string sTest = BuildString("Happy ", 40, std::string("th Birthday!"));
        Assert::AreEqual(std::string("Happy 40th Birthday!"), sTest);
```

`AppendString` works just like `BuildString`, but modifies an input string:
```
        std::string sTest;
        AppendString(sTest, "Happy ", 40, std::string("th Birthday!"));
        Assert::AreEqual(std::string("Happy 40th Birthday!"), sTest);

        AppendString(sTest, L"Happy ", 50U, std::wstring(L"th Birthday!"));
        Assert::AreEqual(std::string("Happy 40th Birthday!Happy 50th Birthday!"), sTest);
```

`StringPrintf` works just like before, but leverages `AppendString`:
```
        Assert::AreEqual(std::string("This is a test."), StringPrintf("This is a %s.", "test"));
        Assert::AreEqual(std::string("1, 2, 3, 4"), StringPrintf("%d, %u, %zu, %li", 1, 2U, 3U, 4L));
        Assert::AreEqual(std::string("53.45%"), StringPrintf("%.2f%%", 53.45f));
        Assert::AreEqual(std::string("01ae"), StringPrintf("%04x", 0x1AEU));
        Assert::AreEqual(std::string("0080"), StringPrintf("%04d", 80));
        Assert::AreEqual(std::string("  80"), StringPrintf("%4d", 80));
```

Internally, `AppendString` builds a list of string references, constructing strings where needed. Then it totals the required size for all of the references and does one `reserve`. Finally, it iterates over the references one last time and copies the data into the output string.

In a release build, running the `TestStringPrintf` in a loop resulted in similar performance. (37.9us per iteration vs. 38.4us per iteration with the previous code).

A comparison of the performance of various string building methods (release build) for concatenating three strings and an integer:
```
1.5us direct append
1.7us StringPrintf
1.8us AppendString
1.8us BuildString
3.1us ostringstream
```

### DIFF
--- a/src/RA_StringUtils.h
+++ b/src/RA_StringUtils.h
@@ -25,65 +25,486 @@ _NODISCARD std::string Narrow(_In_ const std::string& wstr);
 /// <returns>Reference to <paramref name="str" /> for chaining.</returns>
 std::string& TrimLineEnding(_Inout_ std::string& str) noexcept;
 
+// ----- ToString -----
+
+template<typename T>
+_NODISCARD inline const std::string ToString(_In_ const T value)
+{
+    if constexpr (std::is_arithmetic_v<T>)
+    {
+        return std::to_string(value);
+    }
+    else if constexpr (std::is_enum_v<T>)
+    {
+        return std::to_string(static_cast<std::underlying_type_t<T>>(value));
+    }
+    else
+    {
+        static_assert(false, "No ToString implementation for type");
+        return std::string();
+    }
+}
+
+template<> _NODISCARD inline const std::string ToString(_In_ const std::string value)
+{
+    return value;
+}
+
+_NODISCARD inline const std::string& ToString(_In_ const std::string& value)
+{
+    return value;
+}
+
+template<> _NODISCARD inline const std::string ToString(_In_ const std::wstring value)
+{
+    return ra::Narrow(value);
+}
+
+_NODISCARD inline const std::string ToString(_In_ const std::wstring& value)
+{
+    return ra::Narrow(value);
+}
+
+template<> _NODISCARD inline const std::string ToString(_In_ const char* value)
+{
+    return std::string(value);
+}
+
+_NODISCARD inline const std::string ToString(_In_ char* value)
+{
+    return std::string(value);
+}
+
+template<> _NODISCARD inline const std::string ToString(_In_ const char value)
+{
+    return std::string(1, value);
+}
+
+template<> _NODISCARD inline const std::string ToString(_In_ const wchar_t* value)
+{
+    return ra::Narrow(value);
+}
+
+// ----- ToWString -----
+
+template<typename T>
+_NODISCARD inline const std::wstring ToWString(_In_ const T value)
+{
+    if constexpr (std::is_arithmetic_v<T>)
+    {
+        return std::to_wstring(value);
+    }
+    else if constexpr (std::is_enum_v<T>)
+    {
+        return std::to_wstring(static_cast<std::underlying_type_t<T>>(value));
+    }
+    else
+    {
+        static_assert(false, "No ToWString implementation for type");
+        return std::string();
+    }
+}
+
+template<> _NODISCARD inline const std::wstring ToWString(_In_ const std::wstring value)
+{
+    return value;
+}
+
+_NODISCARD inline const std::wstring& ToWString(_In_ const std::wstring& value)
+{
+    return value;
+}
+
+template<> _NODISCARD inline const std::wstring ToWString(_In_ const std::string value)
+{
+    return ra::Widen(value);
+}
+
+_NODISCARD inline const std::wstring ToWString(_In_ const std::string& value)
+{
+    return ra::Widen(value);
+}
+
+template<> _NODISCARD inline const std::wstring ToWString(_In_ const wchar_t* value)
+{
+    return std::wstring(value);
+}
+
+template<> _NODISCARD inline const std::wstring ToWString(_In_ const char* value)
+{
+    return ra::Widen(value);
+}
+
+template<> _NODISCARD inline const std::wstring ToWString(_In_ char* value)
+{
+    return ra::Widen(value);
+}
+
+template<> _NODISCARD inline const std::wstring ToWString(_In_ const char value)
+{
+    return std::wstring(1, value);
+}
+
+// ----- string building -----
+
+class StringBuilder
+{
+public:
+    StringBuilder(bool bPrepareWide = false)
+        : m_bPrepareWide(bPrepareWide)
+    {
+    }
+
+    template<typename T>
+    void Append(T arg)
+    {
+        PendingString pPending;
+        if (m_bPrepareWide)
+        {
+            pPending.WString = ra::ToWString(arg);
+            pPending.DataType = PendingString::Type::WString;
+        }
+        else
+        {
+            pPending.String = ra::ToString(arg);
+            pPending.DataType = PendingString::Type::String;
+        }
+        m_vPending.emplace_back(std::move(pPending));
+    }
+
+    template<>
+    void Append(const std::string& arg)
+    {
+        PendingString pPending;
+        pPending.Ref.String = &arg;
+        pPending.DataType = PendingString::Type::StringRef;
+        m_vPending.emplace_back(std::move(pPending));
+    }
+
+    template<>
+    void Append(const std::wstring& arg)
+    {
+        PendingString pPending;
+        pPending.Ref.WString = &arg;
+        pPending.DataType = PendingString::Type::WStringRef;
+        m_vPending.emplace_back(std::move(pPending));
+    }
+
+    template<>
+    void Append(const char* arg)
+    {
+        AppendSubString(arg, strlen(arg));
+    }
+
+    template<>
+    void Append(const wchar_t* arg)
+    {
+        AppendSubString(arg, wcslen(arg));
+    }
+
+    void AppendSubString(const char* pStart, size_t nLength)
+    {
+        if (nLength == 0)
+            return;
+
+        PendingString pPending;
+        pPending.Ref.Char.Pointer = pStart;
+        pPending.Ref.Char.Length = nLength;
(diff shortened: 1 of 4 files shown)


SyrianBallaS:
I'm going to look at this later more in-depth but seems pretty good. I'm gonna reverse that one "problem" commit so that part that you didn't like doesn't show up.

src/RA_StringUtils.h
```diff
@@ -25,65 +25,486 @@ _NODISCARD std::string Narrow(_In_ const std::string& wstr);
...
+        PendingString pPending;
+        pPending.Ref.WChar.Pointer = pStart;
+        pPending.Ref.WChar.Length = nLength;
+        pPending.DataType = PendingString::Type::WCharRef;
+        m_vPending.emplace_back(std::move(pPending));
     }
-    else if constexpr(std::is_same_v<CharT, wchar_t>)
+
+    template<typename T, typename... Ts>
+    void Append(T arg, Ts... args)
```
SyrianBallaS: `arg` _should_ probably be passed by `const reference`, `args` definitely needs be passed by `rvalue reference`.

src/RA_StringUtils.h
```diff
@@ -25,65 +25,486 @@ _NODISCARD std::string Narrow(_In_ const std::string& wstr);
...
+            {
+                AppendSubString(pFormat, pScan - pFormat);
+            }
+
+            pFormat = pScan;
+        }
+    }
+
+    template<typename CharT, typename = std::enable_if_t<is_char_v<CharT>>, typename T, typename... Ts>
+    void AppendPrintf(const CharT* pFormat, T value, Ts... args)
```
SyrianBallaS: If T is guaranteed to be a literal type then leave it. args should be passed by `rvalue reference` since parameter packs are forwarded. Each arg will be forwarded either as an lvalue or rvalue reference depending on the what current `Ts` is.

This is `std::forward` definitions
```
template< class T >
constexpr T&& forward( typename std::remove_reference<T>::type& t ) noexcept;

template< class T >
constexpr T&& forward( typename std::remove_reference<T>::type&& t ) noexcept;
```

`std::move` is basically the same except that's when the type is already known, `std::forward` is for templates.

src/RA_StringUtils.h
```diff
@@ -25,65 +25,486 @@ _NODISCARD std::string Narrow(_In_ const std::string& wstr);
...
+    StringBuilder builder(true);
+    builder.Append(std::forward<Ts>(args)...);
+    return builder.ToWString();
+}
+
+/// <summary>
+/// Constructs a <see cref="std::basic_string" /> from a format string and parameters
+/// </summary>
+template<typename CharT, typename = std::enable_if_t<is_char_v<CharT>>, typename... Ts>
+_NODISCARD inline auto StringPrintf(_In_z_ _Printf_format_string_ const CharT* const __restrict sFormat, Ts... args)
```
SyrianBallaS: `args` should be passed by `rvalue reference`.

src/RA_StringUtils.h
```diff
@@ -25,65 +25,486 @@ _NODISCARD std::string Narrow(_In_ const std::string& wstr);
...
+        Append(oss.str());
+    }
+
+    template<>
+    void AppendFormat(int arg, const std::string& sFormat)
+    {
+        std::ostringstream oss;
+        if (sFormat.front() == '0')
+            oss << std::setfill('0');
+        int nDigits = atoi(sFormat.c_str());
```
SyrianBallaS: I'm going to suggest using `std::stoi`, mainly because `atoi` fails silently. They basically do the same thing, but stoi is more secure.

src/RA_StringUtils.h
Jamiras: done

src/RA_StringUtils.h
Jamiras: done

src/RA_StringUtils.h
Jamiras: changed `T` to const reference, added && to Ts

src/RA_StringUtils.h
Jamiras: Doesn't work. This is the entry point function and most of the parameter are passed ad hoc.
```
cannot convert from 'const size_t' to 'const size_t &&'
cannot convert from 'const int' to 'const int &&'
cannot convert from 'const char*' to 'const char* const &&'
...
```

SyrianBallaS: Makes sense. Rvalue refs cant be const. In the case of c strings, it cant be a const pointer (const after *)

src/RA_StringUtils.h
```diff
@@ -190,14 +187,30 @@ class StringBuilder
         m_vPending.emplace_back(std::move(pPending));
     }
 
+    void Append(const std::string&& arg)
```
SyrianBallaS: I think you may have misunderstood something, but that's OK.
rvalue references that are const are useless (as well as undefined).
The purpose of std::move is to eliminate copying, the right side has to be in a default state. The left side steals the contents of the right side. 

Example:
```
arg is const&&: auto str = std::move(arg); // arg is unchanged, str is empty
arg is &&: auto str = std::move(arg); // str = arg, arg is empty.
```
If you wanted to use a reference wrapper, this wouldn't even compile:
```
template <class T>
void ref(const T&&) = delete;

template <class T>
void cref(const T&&) = delete;
```

src/RA_StringUtils.h
```diff
@@ -227,21 +240,21 @@ class StringBuilder
     }
 
     template<typename T, typename... Ts>
-    void Append(T arg, Ts... args)
+    void Append(const T&& arg, const Ts&&... args)
     {
         Append(arg);
-        Append(std::forward<Ts>(args)...);
+        Append(std::forward<const Ts>(args)...);
```
SyrianBallaS: These forwards can't use `const`. You can't forward a `const`.

src/RA_StringUtils.h
```diff
@@ -441,49 +453,49 @@ class StringBuilder
...
+        Type DataType{Type::String};
     };
 
     mutable std::vector<PendingString> m_vPending;
-};
+}; // namespace ra
 
 template<typename... Ts>
-void AppendString(std::string& sString, Ts... args)
+void AppendString(std::string& sString, const Ts... args)
```
SyrianBallaS: if it's not possible to put `Ts&&... args`, at least get rid of the `const`.

src/RA_StringUtils.h
```diff
@@ -441,49 +453,49 @@ class StringBuilder
...
+    builder.Append(std::forward<const Ts>(args)...);
     return builder.ToWString();
 }
 
 /// <summary>
 /// Constructs a <see cref="std::basic_string" /> from a format string and parameters
 /// </summary>
 template<typename CharT, typename = std::enable_if_t<is_char_v<CharT>>, typename... Ts>
-_NODISCARD inline auto StringPrintf(_In_z_ _Printf_format_string_ const CharT* const __restrict sFormat, Ts... args)
+_NODISCARD inline auto StringPrintf(_In_z_ _Printf_format_string_ const CharT* const __restrict sFormat, Ts&&... args)
```
SyrianBallaS: I'm not 100% sure. But I think last you said it didn't work because it couldn't forward the const references. 
I believe it's better to get rid of the && here and got rid of the consts for the other parameter packs.

I'd have to test it myself to be sure.

SyrianBallaS:
Elaborated more in the OP

src/RA_StringUtils.h
Jamiras: I think this is ok. Because the `const` is not part of `Ts`, removing it causes the following error:
```
2>e:\source\raintegration\src\ra_stringutils.h(246): error C2665: 'std::forward': none of the 2 overloads could convert all the argument types
```

See also https://stackoverflow.com/questions/8283616/why-doesnt-const-int-ci-2-stdforwardintci-work-and-how-to-fix-wor

> You can forward less cv-qualified expressions to more cv-qualified expressions.

src/RA_StringUtils.h
Jamiras: `const Ts&&` doesn't like string literals:
```
cannot convert argument 2 from 'const char [7]' to 'const char (&&)[7]'
```
But I was able to make it work using `Ts&&` as the input and forwarding them as `const`.

src/RA_StringUtils.h
Jamiras: I believe that's related to the string literals. The input `Ts` must be non-const, but through forwarding magic, they can be converted to `const`.

Jamiras:
In the end, I ended up removing the `const` from all of the forwards. While the parameters are meant to be read-only, the explicit `const` might be affecting the efficiency of the `&&`.

PR #199, opened 2018-11-19
Compile-time check for literal types

SyrianBallaS:
To attempt to optimize code further, I made this `type_trait` to adjust for situations if a type `T` is a primitive or not. I'll explain some things after this PR is submitted.

The only thing I couldn't figure out how to do was test for unions, but aggregate types pass/fail as expected.

Example:
```
static_assert(ra::is_literal_type<std::array<int, 5>>); // passes
static_assert(ra::is_literal_type<std::array<std::string, 5>>); // fails, std::string is not a literal type.
```

Reference:
https://en.cppreference.com/w/cpp/named_req/LiteralType

### DIFF
--- a/src/ra_type_traits.h
+++ b/src/ra_type_traits.h
@@ -33,14 +33,16 @@ struct _NODISCARD is_char : std::bool_constant<(std::is_same_v<CharT, char> || s
 /// <summary>
 ///   This should only be used to compare the sizes of data types and not objects.
 /// </summary>
-template<typename T, typename U> struct _NODISCARD is_same_size : std::bool_constant<sizeof(T) == sizeof(U)>
+template<typename T, typename U>
+struct _NODISCARD is_same_size : std::bool_constant<sizeof(T) == sizeof(U)>
 {
 };
 
 /// <summary>
 ///   This should only be used to compare the sizes of data types and not objects.
 /// </summary>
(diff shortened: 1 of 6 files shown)


src/ra_type_traits.h
```diff
@@ -125,9 +132,28 @@ _CONSTANT_VAR is_lessthan_comparable_v{detail::is_lessthan_comparable<LessThanCo
...
+_CONSTANT_VAR is_nothrow_comparable_v{detail::is_nothrow_comparable<Comparable>::value};
+
+namespace detail {
+
+template<typename T>
+struct _NODISCARD is_literal_type
+    : std::bool_constant<(std::is_scalar_v<T> || std::is_reference_v<T>) ||
+                         (std::is_aggregate_v<T> && std::is_scalar_v<typename T::value_type>) ||
+                         (std::is_class_v<T> && std::is_trivially_destructible_v<T> &&
+                          std::is_nothrow_constructible_v<T>)>
```
SyrianBallaS: According the standard a LiteralType is (Summarized):
- scalar type or
- reference type or
- an array of literal type or (array is an aggregate type)
- a class of literal type

I couldn't find a way to test for a constexpr constructor, union volatility, or closure type (lambda). 

The last condition should mostly be enough for our purposes. If the ones for `is_class_v` pass, it has a very high chance of having a constexpr constructor. We don't deal with device drivers or other type of hardware code so volatile shouldn't matter. Lambdas are _usually_ (not always) reference types, there is a `std::is_function_v`, but not all function types are literal types.

PR #201, opened 2018-11-24
Tabs to spaces

SyrianBallaS:
### Summary
- Tabs to spaces
- Applied the ClangFormat only to some files to see if the .clang-format file needs adjustments
- Spell checked everything
- Suppressed a VC++ tool-chain deprecation warning.
- Re-enabled assignment alignment

Review comments will be posted for the most significant areas.

Edit: for most of them I just replaced` "\t"` with `" "` by document (using regular expressions for future reference). I didn't do it by project because lua, rcheevos, and md5 are currently in there.

The line count can be reduced significantly be getting rid of commented out code.


src/RA_Achievement.cpp
```diff
@@ -574,225 +606,225 @@ void Achievement::Set(const Achievement& rRHS)
...
-// 
-//		nIterations++;
-//	}
-// 
-//	if( nIterations == 20 )
-//	{
-//		ASSERT(!"Bugger... can't parse this 'progress' thing. Too many iterations!");
-//	}
-// 
-//	return fProgressValue;
```
SyrianBallaS: Should we just get rid of all this commented out code? It'll reduce the lines "added" significantly.

SyrianBallaS:
I'll fix the conflicts asap but not sure when. Closing.

PR #202, opened 2018-11-24
Enum to enum class (CondSubItems)

SyrianBallaS:
### Summary
- Made CondSubItems an `enum class`
- Made related arrays as bounded `std::array `(except for that 3D one)
- Put a subtraction operator for the `enum class` (it was needed here)
- Applied ClangFormat
- Cleaned up the header file a bit

There will be review comments for some things

### DIFF
--- a/src/RA_Dlg_AchEditor.cpp
+++ b/src/RA_Dlg_AchEditor.cpp
@@ -1,14 +1,14 @@
 #include "RA_Dlg_AchEditor.h"
 
 #include "RA_AchievementSet.h"
-#include "RA_Resource.h"
 #include "RA_Core.h"
 #include "RA_Dlg_Achievement.h"
 #include "RA_Dlg_Memory.h"
-#include "RA_httpthread.h"
 #include "RA_ImageFactory.h"
 #include "RA_MemManager.h"
+#include "RA_Resource.h"
 #include "RA_User.h"
+#include "RA_httpthread.h"
 
 #include "services\IConfiguration.hh"
 #include "services\ServiceLocator.hh"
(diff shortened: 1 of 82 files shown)


src/RA_Dlg_AchEditor.cpp
```diff
@@ -2147,9 +2043,14 @@ void Dlg_AchievementEditor::LoadAchievement(Achievement* pCheevo, _UNUSED BOOL)
...
 
-    //InvalidateRect( hList, nullptr, TRUE );
-    //UpdateWindow( hList );
-    //RedrawWindow( hList, nullptr, nullptr, RDW_INVALIDATE );
+    // InvalidateRect( hList, nullptr, TRUE );
+    // UpdateWindow( hList );
+    // RedrawWindow( hList, nullptr, nullptr, RDW_INVALIDATE );
+}
+
+inline char* Dlg_AchievementEditor::LbxDataAt(unsigned int nRow, CondSubItems nCol)
```
SyrianBallaS: To not have to put the enum in the header file, this is defined in the source file.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -631,22 +507,23 @@ BOOL CreateIPE(int nItem, int nSubItem)
...
-            //	Size: dropdown
+            // Size: dropdown
             ASSERT(g_hIPEEdit == nullptr);
             if (g_hIPEEdit)
                 break;
 
-            //	Note: this relies on column order :S
+            // Note: this relies on column order :S
+            using namespace ra::arith_ops;
             if (strcmp(g_AchievementEditorDialog.LbxDataAt(nItem, nSubItem - 1), "Value") == 0)
```
SyrianBallaS: This is were the subtraction operator was needed.

src/RA_Dlg_AchEditor.h
```diff
@@ -2,33 +2,25 @@
...
     void FetchNewBadgeNamesThreaded();
     void AddNewBadgeName(const char* pStr, bool bAndSelect);
     void OnNewBadgeNames(const rapidjson::Document& data);
 
 private:
-    HWND m_hDestComboBox;
+    HWND m_hDestComboBox = nullptr;
 };
 
+enum class CondSubItems : std::size_t;
```
SyrianBallaS: This needs to be forward declared since it's not defined here.
If putting the enum as a private member is desired, CreateIPE needs to be a friend function.

src/RA_Dlg_AchEditor.h
```diff
@@ -77,22 +71,24 @@ class Dlg_AchievementEditor
     void UpdateCondition(HWND hList, LV_ITEM& item, const Condition& Cond, unsigned int nCurrentHits);
 
 private:
-    static const int m_nNumCols = 10;//;sizeof( g_sColTitles ) / sizeof( g_sColTitles[0] );
+    static constexpr std::size_t m_nNumCols = 10U;
+    static constexpr std::size_t MAX_CONDITIONS = 200U;
+    static constexpr std::size_t MEM_STRING_TEXT_LEN = 80U;
```
SyrianBallaS: Just put these all here. I left the `m_nNumCols` name alone, but static class variables aren't really data members since they aren't associated with a class instance.

src/ra_utility.h
```diff
@@ -214,6 +216,21 @@ _CONSTANT_FN& operator&=(_Inout_ Enum& a, _In_ Enum b) noexcept
...
+{
+    return ra::itoe<Enum>(ra::etoi(a) - b);
+}
+
+template<typename Enum, typename = std::enable_if_t<std::is_enum_v<Enum>>>
+_CONSTANT_FN& operator-=(_Inout_ Enum& a, _In_ std::underlying_type_t<Enum> b) noexcept
+{
+    return (a = a - b);
+}
+} // namespace arith_ops
```
SyrianBallaS: Pretty simple, just cast around with valid input. wasn't sure if we needed more though. On that note, I wasn't sure we needed more bitwise operator overloads at the time I did them.

PR #204, opened 2018-11-27
Enum to Enum Class (From RA_httpthread.h)

SyrianBallaS:
- Made the enums in RA_httpthread.h as `enum class`.
- Made the unbounded C arrays into bounded compile-time constants of  `std::array` .
  - Did not rename the contents of the arrays.
- Minor optimization.
  - Mainly where string building was taking place
- This should be the last of the "Enum to Enum Class" PRs.


SyrianBallaS:
I'll fix the conflicts asap but not sure when. Closing.

PR #205, opened 2018-11-30
Support for AndNext condition flag

Jamiras:
runtime support here: https://github.com/RetroAchievements/rcheevos/pull/12

The AndNext flag tells the runtime that the flagged condition and the next condition must both be true at the same time. Multiple AndNext conditions may be chained to create a single complex requirement. The final condition may have a HitCount that will only be incremented if final condition and all of the preceding AndNext conditions are true. Similarly, the final condition may be marked with PauseIf or ResetIf and will only pause or reset if the final condition and all of the preceding AndNext conditions are true.


Jamiras:
Requested binary for testing: [RA_Integration-AndNext.zip](https://github.com/RetroAchievements/RAIntegration/files/2631810/RA_Integration-AndNext.zip)

Keltron3030:
Thank you for this.  I can confirm it works as intended for my use case, providing a hit only when the two prescribed conditions were met.  I also tested with a resetif flag following the addnext and this also worked as intended, resetting all hits only when the two prescribed conditions were met.

rzumer:
Is there something blocking this?

Jamiras:
There's two reasons that I haven't merged this yet:
1) No RetroArch support - I think it's wise not to have people writing achievements that we know will fail for roughly half of our users.
2) The rcheevos portion of the code has not made it's way to master. Technically, this isn't a big issue because we had to point to rcheevos/develop for the 0.75 release to pick up a bug fix that also hasn't made it to rcheevos/master yet. Although, I'd prefer we get back to pointing at master ASAP.

rzumer:
Fair enough, I assume https://github.com/libretro/RetroArch/issues/7341 is the issue tracking remaining issues with it in RetroArch.

Jamiras:
Implements #6 and can be used to satisfy #3

Issue #176, opened 2018-11-11
New Achievement "Type" (Mem/Delta/Value) type: Previous, similar to delta


kdecks:
As inspired/suggested by **Eldexter** in the devs channel in Discord:

A new "Type" type called **Previous** (name unimportant), aka ``prev.``, which would compare the previous value, for any duration of time, until the value has changed.

- Currently there are three "Type"s: ``Mem``/``Delta``/``Value``
- You can easily see the ``prev.`` value of an address using the Memory Bookmarker.

As a developer I'd like to have flexible tools which can give me a wide variety of solutions. ``Prev.`` would give us that. Its use would would be similar to ``delta``, but provide unique solutions which would otherwise not be possible. Also, ``prev.`` would also be a simpler concept for beginner developers to get their head around.

@Jamiras said on the topic: 
> deltas are already stored in a buffer. the only thing that would change is the logic that updates the buffer would only occur if the actual value changes.

Meaning implementation is realistic, and possibly not difficult.

### In context
In a game I'm on stage 1 and the go to stage 2. I'm watching the stage ID address which are 1 & 2.

For ``mem`` the value will be which every stage I'm on. For ``delta``, it would be which ever stage I've been on, one frame ago. For ``prev.``, for the duration of stage 2 the value would be 1. For stage 3 the value would be 2, and so on.

### Some use cases
(while working on Banjo-Tooie)
> You can use it for beat a boss, if you weren't in the previous room at the boss room pause.
> 
> You can do that also with cutscenes, text or audios if the game work with IDs of that cases if the player didn't listen/view/read x element pause
> 
> That's useful at games like only change the value when is used like Banjo Tooie, the text ID is rewritten after the text box is gone

— Eldexter

PR #206, opened 2018-11-30
Support for Prior memory type (implements #176)

Jamiras:
runtime support here: https://github.com/RetroAchievements/rcheevos/pull/11

The Prior memory type works like Delta, but reports the last differing value, instead of the value from the previous frame. The first frame a value changes, Delta and Prior will be the old value. Every frame after that until the value changes again, Delta will be the value and Prior will be the old value.

For example:

| Frame | Value | Delta | Prior |
| ----- | ----- | ----- | ----- |
|    1 |   25 |    0 |    0 |
|    2 |   25 |   25 |    0 |
|    3 |   30 |   25 |   25 |
|    4 |   35 |   30 |   30 |
|    5 |   35 |   35 |   30 |   
|    6 |   35 |   35 |   30 |
|    7 |   40 |   35 |   35 |


Jamiras:
Requested binary for testing: [RA_Integration-Prior.zip](https://github.com/RetroAchievements/RAIntegration/files/2631865/RA_Integration-Prior.zip)

PR #208, opened 2018-12-02
Cppcorecheck function rules

SyrianBallaS:
This only addresses one rule but was triggered many times
`F.6: If your function may not throw, declare it noexcept`

Basically this does three things (from order of most to least changes):
- All functions guaranteed not to throw (nothrow exception guarantee) are marked as `noexcept`
- Functions that have a high chance of throwing have `noexcept` removed
  - Some virtual methods had noexcept removed or had a warning where it could be noexcept suppressed because overrides could throw and subclasses can't have a different exception spec than the parent class.
  - I could be wrong, if I am, please say so where so I can suppress the analysis warning
- Functions that have a low chance of throwing have the rule suppressed.
  - Had to analyze these very carefully
  - Made some changes to functions to minimize the chance of exceptions being thrown

If there's any suggestions for improvement, please say so.

### DIFF
--- a/src/RA_Achievement.h
+++ b/src/RA_Achievement.h
@@ -17,14 +17,14 @@ public:
     enum class DirtyFlags
     {
         Clean,
-        Title = 1 << 0,
-        Desc = 1 << 1,
-        Points = 1 << 2,
-        Author = 1 << 3,
-        ID = 1 << 4,
-        Badge = 1 << 5,
-        Conditions = 1 << 6,
-        Votes = 1 << 7,
+        Title       = 1 << 0,
+        Desc        = 1 << 1,
+        Points      = 1 << 2,
+        Author      = 1 << 3,
+        ID          = 1 << 4,
+        Badge       = 1 << 5,
+        Conditions  = 1 << 6,
+        Votes       = 1 << 7,
         Description = 1 << 8,
 
         All = std::numeric_limits<std::underlying_type_t<DirtyFlags>>::max()
@@ -107,7 +107,7 @@ public:
 
     // Returns the new char* offset after parsing.
     [[gsl::suppress(f.6)]] const char*
-        ParseLine(const char* sBuffer); /*Doesn't throw in tests but does in Integration */
+        ParseLine(const char* restrict sBuffer); /*Doesn't throw in tests but might in Integration */
     const char* ParseStateString(const char* sBuffer, const std::string& sSalt);
 
 #ifndef RA_UTEST
--- a/src/RA_Achievement.h
+++ b/src/RA_Achievement.h
@@ -17,14 +17,14 @@ public:
     enum class DirtyFlags
     {
         Clean,
-        Title       = 1 << 0,
-        Desc        = 1 << 1,
-        Points      = 1 << 2,
-        Author      = 1 << 3,
-        ID          = 1 << 4,
-        Badge       = 1 << 5,
-        Conditions  = 1 << 6,
-        Votes       = 1 << 7,
+        Title = 1 << 0,
+        Desc = 1 << 1,
+        Points = 1 << 2,
+        Author = 1 << 3,
+        ID = 1 << 4,
+        Badge = 1 << 5,
+        Conditions = 1 << 6,
+        Votes = 1 << 7,
         Description = 1 << 8,
 
         All = std::numeric_limits<std::underlying_type_t<DirtyFlags>>::max()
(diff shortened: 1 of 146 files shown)


src/RA_Interface.h
```diff
@@ -40,81 +40,85 @@ enum EmulatorID
...
 extern void RA_CauseUnpause();
 extern void RA_CausePause();
 extern void RA_RebuildMenu();
 extern void RA_GetEstimatedGameTitle(char* sNameOut);
 extern void RA_ResetEmulation();
 extern void RA_LoadROM(const char* sFullPath);
-
+#ifdef __cplusplus
+}
+#endif // __cplusplus
```
SyrianBallaS: I put this around an `extern "C"` block because the analyzer was telling me to mark them `noexcept`, but wasn't sure if an emulator supports at least C++11.

SyrianBallaS:
I was looking at this again and noticed some unexpected formatting changes. I'll reopen this tomorrow or something.

src/RA_Dlg_Memory.cpp
```diff
@@ -57,33 +57,24 @@ _NODISCARD _CONSTANT_FN GetMaxNibble(_In_ MemSize size) noexcept
...
-
-INT_PTR CALLBACK MemoryViewerControl::s_MemoryDrawProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
+LRESULT CALLBACK MemoryViewerControl::s_MemoryDrawProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
 {
     switch (uMsg)
     {
         case WM_NCCREATE:
-        case WM_NCDESTROY:
-            return TRUE;
+        case WM_NCDESTROY: return TRUE;
```
Jamiras: I'm not sure I like this formatting change. It's fine if all the cases collapse to a single line,  like above in `GetMaxNibble`, but interspersed with multi-line blocks, it feels unnatural. I think I'd prefer to side on making the simple switches more verbose rather than trying to collapse parts of a more complex switch.

Recommend changing `AllowShortCaseLabelsOnASingleLine` to false, since it seems to ignore the "Allow" and behaves like "EnforceShortCaseLabelsOnASingleLine".

src/RA_Dlg_Memory.cpp
```diff
@@ -998,11 +977,11 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
...
         case WM_GETMINMAXINFO:
         {
-            LPMINMAXINFO lpmmi = (LPMINMAXINFO)lParam;
+            LPMINMAXINFO lpmmi      = (LPMINMAXINFO)lParam;
             lpmmi->ptMaxTrackSize.x = pDlgMemoryMin.x;
-            lpmmi->ptMinTrackSize = pDlgMemoryMin;
+            lpmmi->ptMinTrackSize   = pDlgMemoryMin;
         }
-        return TRUE;
+            return TRUE;
```
Jamiras: this looks weird, recommend moving the return statement inside the brackets.

src/RA_Dlg_Memory.cpp
```diff
@@ -1014,11 +993,9 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
 
             RememberWindowSize(hDlg, "Memory Inspector");
         }
-        return TRUE;
+            return TRUE;
```
Jamiras: this looks weird, recommend moving the return statement inside the brackets.

src/RA_Dlg_Memory.cpp
```diff
@@ -1104,7 +1083,7 @@ INT_PTR Dlg_Memory::MemoryProc(HWND hDlg, UINT nMsg, WPARAM wParam, LPARAM lPara
 
                     EnableWindow(GetDlgItem(hDlg, IDC_RA_DOTEST), nMatches > 0);
                 }
-                return TRUE;
+                    return TRUE;
```
Jamiras: this looks weird, recommend moving the return statement inside the brackets.

src/RA_Dlg_Memory.cpp
```diff
@@ -1440,7 +1421,7 @@ void Dlg_Memory::RepopulateMemNotesFromFile()
     size_t nSize = 0;
 
     const auto& pGameContext = ra::services::ServiceLocator::Get<ra::data::GameContext>();
-    const auto nGameID = pGameContext.GameId();
+    const auto nGameID       = pGameContext.GameId();
```
Jamiras: I think I'd like to disable `AlignConsecutiveAssignments` as well. It's nice when assigning constant values, but is less pleasing when aligning function results.

src/RA_Integration.vcxproj
```diff
@@ -230,7 +230,7 @@
     <ProjectName>RA_Integration</ProjectName>
     <ProjectGuid>{825CB292-F069-4B27-82CF-3417746ACDF3}</ProjectGuid>
     <RootNamespace>RA_Integration</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
```
Jamiras: Why did this change?

src/RA_Interface.h
```diff
@@ -15,22 +15,22 @@ struct ControllerInput
...
-    RA_Nester = 4,
-    RA_FCEUX = 5,
-    RA_PCE = 6,
-    RA_Libretro = 7,
-    RA_Meka = 8,
+    RA_Nester           = 4,
+    RA_FCEUX            = 5,
+    RA_PCE              = 6,
+    RA_Libretro         = 7,
+    RA_Meka             = 8,
```
Jamiras: This is a good use of aligning equals, but if it means causing the other alignings to happen, I'd prefer to leave it the way it was.

src/RA_Interface.h
```diff
@@ -40,81 +40,85 @@ enum EmulatorID
...
+    Events           = 23, // not an actual console
+    XboxOne          = 24,
+    Atari2600        = 25,
+    MSDOS            = 26,
+    Arcade           = 27,
+    VirtualBoy       = 28,
+    MSX              = 29,
+    C64              = 30,
+    ZX81             = 31,
+    // unused32 = 32,
```
Jamiras: this comment is confusing the aligner. another vote for disabling it.

src/RA_Leaderboard.cpp
```diff
@@ -97,40 +89,16 @@ void RA_Leaderboard::Submit(unsigned int nScore)
...
-    {
-        if ((*iter).m_nRank == nRank)
-        {
-            (*iter) = newEntry;
-            return;
-        }
-        iter++;
-    }
+    if (EntryExists(nRank))
+        return;
```
Jamiras: This assumes the list is always sorted, which implies that it needs to be resorted after the emplace_back, or the item needs to be inserted in the correct place.

src/RA_Leaderboard.h
```diff
@@ -8,41 +8,55 @@ class RA_Leaderboard
 {
 public:
     explicit RA_Leaderboard(_In_ const ra::LeaderboardID nLBID) noexcept;
-    virtual ~RA_Leaderboard() noexcept = default;
+    virtual ~RA_Leaderboard() noexcept             = default;
     RA_Leaderboard(const RA_Leaderboard&) noexcept = delete;
     RA_Leaderboard& operator=(const RA_Leaderboard&) noexcept = delete;
-    RA_Leaderboard(RA_Leaderboard&&) noexcept = default;
+    RA_Leaderboard(RA_Leaderboard&&) noexcept                 = default;
```
Jamiras: Is this because of the "AlignConsecutiveAssignments" flag? Why are the first two not aligned with the last two, or the next one? Another vote to disable it.

src/RA_Leaderboard.h
```diff
@@ -51,18 +65,41 @@ class RA_Leaderboard
...
 private:
-    const ra::LeaderboardID          m_nID;                 //  DB ID for this LB
+    const ra::LeaderboardID m_nID = ra::LeaderboardID(); //  DB ID for this LB
 
-    void*                            m_pLeaderboard;        //  rc_lboard_t
-    std::shared_ptr<unsigned char[]> m_pLeaderboardBuffer;  //  buffer for rc_lboard_t
-    unsigned int                     m_nCurrentValue = 0;
+    void* m_pLeaderboard = nullptr;                        //  rc_lboard_t
+    std::shared_ptr<unsigned char[]> m_pLeaderboardBuffer; //  buffer for rc_lboard_t
+    unsigned int m_nCurrentValue = 0U;
```
Jamiras: Here, the formatter actually undid some manual alignments. I'm okay with that, but the comments should be realigned.

src/api/IServer.hh
```diff
@@ -16,11 +16,12 @@ public:
...
-    virtual Login::Response Login(const Login::Request& request) noexcept = 0;
-    virtual Logout::Response Logout(const Logout::Request& request) noexcept = 0;
-    virtual StartSession::Response StartSession(const StartSession::Request& request) noexcept = 0;
-    virtual Ping::Response Ping(const Ping::Request& request) noexcept = 0;
+    // TBD: Overrides can throw
+    [[gsl::suppress(f.6)]] virtual Login::Response Login(const Login::Request& request)    = 0;
+    [[gsl::suppress(f.6)]] virtual Logout::Response Logout(const Logout::Request& request) = 0;
+    [[gsl::suppress(f.6)]] 
+    virtual StartSession::Response StartSession(const StartSession::Request& request) = 0;
+    [[gsl::suppress(f.6)]] virtual Ping::Response Ping(const Ping::Request& request) = 0;
```
Jamiras: None of the toolkit code is designed to handle exceptions very well, so I'd rather have everything written assuming `noexcept` and patch the areas that can throw exceptions with try/catch blocks.

src/api/impl/ConnectedServer.cpp
```diff
@@ -118,9 +121,9 @@ static void AppendUrlParam(_Inout_ std::string& sParams, _In_ const char* const
     ra::services::Http::UrlEncodeAppend(sParams, sValue);
 }
 
-Login::Response ConnectedServer::Login(const Login::Request& request) noexcept
+Login::Response ConnectedServer::Login(const Login::Request& request)
 {
-    ra::services::Http::Request httpRequest(m_sHost + "/login_app.php");
+    ra::services::Http::Request httpRequest(ra::StringPrintf("%s/login_app.php", m_sHost.c_str()));
```
Jamiras: `ra::StringPrintf` implicitly supports std::string via the variadic templating. You don't need to downcast via `c_str()`.

src/RA_Achievement.h
```diff
@@ -33,8 +33,8 @@ class Achievement
     Achievement() noexcept;
 
 public:
-    void Clear();
-    bool Test();
+    void Clear() noexcept;
+    [[gsl::suppress(f.6)]] bool Test() noexcept;
```
Jamiras: Consider using the `GSL_SUPPRESS()` macro. It adds some clang compatibility, and makes it easier to disable if we ever support a platform that doesn't understand it.

src/RA_Dlg_Memory.cpp
SyrianBallaS: I did that to prevent changes to areas that already have them on one line. It can go to false. 

"Short" depends on the line length currently set, which is 120. So it might look weird sometimes. Putting it to false is what I prefer as well.

src/RA_Dlg_Memory.cpp
SyrianBallaS: For all of these, I'll do that.

src/RA_Dlg_Memory.cpp
SyrianBallaS: That should be fine. It causes more problems than it solves with that option on.

src/RA_Integration.vcxproj
SyrianBallaS: I need yo reinstall the fall creators SDK again

src/RA_Interface.h
SyrianBallaS: Sure

src/RA_Leaderboard.cpp
SyrianBallaS: I probably shouldn't have moved it to another function, but it finds it via binary search

src/api/IServer.hh
SyrianBallaS: Alright, I'll suppress where it can throw but mark as noexcept

src/api/impl/ConnectedServer.cpp
SyrianBallaS: Sweet, that's convenient

src/RA_Leaderboard.cpp
Jamiras: Yes, I see that, but BinarySearch only works if the list is sorted. This method only appends to the list and Sort is called later. If the items aren't added in order, the BinarySearch may fail.

SyrianBallaS:
All right, I think I addressed everything. I did as Microsoft's version of gsl as a submodule. It will help immensely.

PR #213, opened 2018-12-12
Cppcorecheck smart ptr rules

SyrianBallaS:
**Summary**
- Added shared and unique pointer rules
- Addressed possible memory leaks
- Unit tests pass

This adds two rulesets which are basically the same, there are minor differences but when one was addressed the other did not produce warnings. They are for `shared_ptr `and `unique_ptr`. I tried finding some rulesets that might be included in another one because the main ones get triggered to much times (sometimes over 800).

`R.5: Prefer scoped objects, don’t heap-allocate unnecessarily`
Where there was heap allocation, it was necessary. 

It warned from this: `Warn if a local Unique_ptr or Shared_ptr is not moved, copied, reassigned or reset before its lifetime ends.`

There was something that the analyzer did not catch, that was with the shared pointers with dynamic arrays. std::shared_ptr doesn't have an array specialization yet so it'll call delete instead of delete[], causing a leak. So there is a vector inside the shared pointer and that vector is the "array". Stuff is correctly allocated using `make_shared `and no leaks were observed from the heap profiler.


src/RA_Core.cpp
```diff
@@ -1455,6 +1455,7 @@ std::string GetFolderFromDialog()
         ret = ra::Narrow(lpbi->pszDisplayName);
     }
     ::OleUninitialize();
+    pDisplayName.reset();
```
Jamiras: The point of the rule complaining here is to avoid using a unique pointer when one isn't necessary. 

> Smart pointers are convenient for dynamic resource management, but they are not always necessary. For example, creating of a local dynamic buffer can be easily (end sometimes more efficiently) managed by standard containers. 

I believe the recommended solution is to make `pDisplayName` a stack variable (i.e. either  `TCHAR[RA_MAX_PATH]` or `std::array<TCHAR, RA_MAX_PATH>`) instead of heap variable. I'm surprised it's not also complaining about `lpbi`. I'd recommend giving it similar treatment.

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -286,11 +286,11 @@ void Dlg_GameLibrary::ThreadedScanProc()
             const DWORD nSize = ftell(pf);
             rewind(pf);
 
-            // May have caused a buffer overrun, this is way to big to be on the stack
-            auto pBuf{ std::make_unique<unsigned char[]>(6 * 1024 * 1024) };
+            auto pBuf = std::make_unique<unsigned char[]>(6 * 1024 * 1024);
 
             fread(static_cast<void*>(pBuf.get()), sizeof(unsigned char), nSize, pf); //Check
             Results.insert_or_assign(FilesToScan.front(), RAGenerateMD5(pBuf.get(), nSize));
+            pBuf.reset();
```
Jamiras: In this case, it's unreasonable to put 6MB on the stack, so the change seems appropriate, but unnecessary. When pBuf is destructed it will also release the memory. But releasing early is always good, so I'd rather not fight the tool over this.

PR #214, opened 2018-12-20
CppCoreCheck Owner Pointer Rules

SyrianBallaS:
**Summary**
- Added the owner pointer rules
- Needed to overhaul `Dlg_MemBookmark` for stuff to work
   - Makes stuff cleaner and expressive, none of that messy manual stuff (at least with free-store)
- More details will be in the review comments after this PR is submitted

If this needs any conflicts fixed before it's accepted I'll fix them when the time comes.

Many rulesets have child rulesets within them. This is a child ruleset. Main rulesets have too many warnings to work with at once. The goal is to get bounds, lifetime, and type profiles added to the code-base but doing them all at once seems too much at once. This didn't take that long but a lot of time for testing to make sure undefined stuff doesn't happen in release mode (tested in all configurations).

There is one weird thing I noticed and wasn't sure how to address it. If you load a bookmark file from another game, the emulator doesn't crash but the memory values keep spazzing out. I think there needs to be a naming convention for bookmark files unless the default is the convention.

**Rules Addressed**
`F.23: Use a not_null<T> to indicate that “null” is not a valid value`
-  Whenever a pointer was never tested for nullness is was made a `not_null`
- It helped find areas were a `nullptr` was not expected yet exists. 

`R.3: A raw pointer (a T*) is non-owning`
- Basically, don't use `new` and `delete` directly. It sometimes wasn't triggered but ignored the ones it didn't catch for now.
- It's non-owning because a raw pointer doesn't manage the resource at all

`R.4: A raw reference (a T&) is non-owning`
`R.11: Avoid calling new and delete explicitly`
- Perfect example is in Dlg_MemBookmark, when you new, you're supposed to delete but it causes an access violation.


src/RA_Dlg_MemBookmark.cpp
```diff
@@ -20,437 +22,434 @@ namespace ra {
...
+        HANDLE_MSG(hDlg, WM_GETMINMAXINFO, OnGetMinMaxInfo);
+        HANDLE_MSG(hDlg, WM_SIZE, OnSize);
+        HANDLE_MSG(hDlg, WM_MEASUREITEM, OnMeasureItem);
+        HANDLE_MSG(hDlg, WM_DRAWITEM, OnDrawItem);
+        HANDLE_MSG(hDlg, WM_NOTIFY, OnNotify);
+        HANDLE_MSG(hDlg, WM_COMMAND, OnCommand);
+
+        case WM_MOVE:
+            RememberWindowPosition(hDlg, "Memory Bookmarks");
+            break;
```
SyrianBallaS: The stuff in `WM_MOVE` seemed like it didn't need a callback for the message cracker.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -514,43 +513,39 @@ void Dlg_MemBookmark::PopulateList()
...
 
-        ASSERT(item.iItem == m_nNumOccupiedRows);
-
+        Ensures(item.iItem == m_nNumOccupiedRows);
         m_nNumOccupiedRows++;
     }
 
     ListView_EnsureVisible(hList, m_vBookmarks.size() - 1, FALSE); // Scroll to bottom.
-    //ListView_EnsureVisible( hList, topIndex, TRUE ); // return to last position.
+    // Don't sort here
```
SyrianBallaS: What I meant is don't sort the vector. We could sort the ListView here though, but don't know how to do it. Would need to research and learn it.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -750,42 +724,43 @@ void Dlg_MemBookmark::ExportJSON()
...
+    const auto file_name{std::move(sFilename)};
+    std::ifstream ifile{file_name};
     if (!ifile.is_open())
     {
-        char buf[2048U]{};
-        strerror_s(buf, errno);
-        RA_LOG("%s: %s", ra::Narrow(sFilename).c_str(), buf);
+        std::array<char, 2048> buf{};
+        strerror_s(buf.data(), sizeof(buf), errno);
+        RA_LOG("%s: %s", ra::Narrow(sFilename).c_str(), buf.data());
```
SyrianBallaS: I could have used `std::string`, but the internal pointer is dynamically allocated on the free-store.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -799,27 +774,25 @@ void Dlg_MemBookmark::ImportFromFile(std::wstring sFilename)
...
         ClearAllBookmarks();
 
-        const auto& BookmarksData{ doc["Bookmarks"] };
+        const auto& BookmarksData{doc["Bookmarks"]};
+
         for (const auto& bmData : BookmarksData.GetArray())
         {
-            auto NewBookmark{ new MemBookmark };
-            NewBookmark->SetDescription(ra::Widen(bmData["Description"].GetString()));
+            MemBookmark NewBookmark;
```
SyrianBallaS: The darn new was the source of all our problems. But there were many problems with this class. Like casting that could have been avoided, algorithm complexity, raw pointers, etc.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -888,15 +848,8 @@ void Dlg_MemBookmark::OnLoad_NewRom()
...
-        std::wstring file;
-        {
-            const auto& pGameContext = ra::services::ServiceLocator::Get<ra::data::GameContext>();
-            std::wostringstream oss;
-            oss << g_sHomeDir << RA_DIR_BOOKMARKS << pGameContext.GameId() << L"-Bookmarks.txt";
-            file = oss.str();
-        }
-        ImportFromFile(file);
+        const auto& pGameContext = ra::services::ServiceLocator::Get<ra::data::GameContext>();
+        ImportFromFile(ra::StringPrintf(L"%s%s%u-Bookmarks.txt", g_sHomeDir, RA_DIR_BOOKMARKS, pGameContext.GameId()));
```
SyrianBallaS: A function return is a rvalue-reference. Originally ImportFromFile used copying which is slow. const ref caused problems, moving is always better.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -957,12 +905,12 @@ BOOL Dlg_MemBookmark::EditLabel(int nItem, int nSubItem)
         return FALSE;
     };
 
-    SendMessage(g_hIPEEditBM, WM_SETFONT, (WPARAM)GetStockObject(DEFAULT_GUI_FONT), TRUE);
-    SetWindowText(g_hIPEEditBM, NativeStr(m_vBookmarks[nItem]->Description()).c_str());
+    SetWindowFont(g_hIPEEditBM, ::GetStockObject(DEFAULT_GUI_FONT), TRUE);
```
SyrianBallaS: This message cracker casts everything so it doesn't really matter if you use `GetStockObject `or `GetStockFont`. It does expect `HFONT`, but the value is "unboxed".

src/RA_Dlg_MemBookmark.h
```diff
@@ -3,42 +3,92 @@
...
-    unsigned int Previous() const noexcept { return m_sPrevious; }
-    unsigned int Count() const noexcept { return m_nCount; }
-
-    bool Frozen() const noexcept { return m_bFrozen; }
-    bool Decimal() const noexcept { return m_bDecimal; }
+public: // Too much trouble making this a literal type
+    MemBookmark() noexcept = default;
+    ~MemBookmark() noexcept = default;
+    // Copying deleted to enforce uniqueness and speed
+    MemBookmark(const MemBookmark&) = delete;
```
SyrianBallaS: This doesn't mean that there can't be two bookmarks with the same address, it just means that it can't be copied in the same execution (uniqueness).
Think if it like `unordered_multimap `where stuff is sorted into buckets if you want them sorted but doesn't sort automatically.

src/RA_Dlg_MemBookmark.h
```diff
@@ -3,42 +3,92 @@
...
+    _NODISCARD _CONSTANT_FN Type() const noexcept { return m_nType; }
+    _NODISCARD _CONSTANT_FN Value() const noexcept { return m_sValue; }
+    _NODISCARD _CONSTANT_FN Previous() const noexcept { return m_sPrevious; }
+    _NODISCARD _CONSTANT_FN Count() const noexcept { return m_nCount; }
+
+    _NODISCARD _CONSTANT_FN Frozen() const noexcept { return m_bFrozen; }
+    _NODISCARD _CONSTANT_FN Decimal() const noexcept { return m_bDecimal; }
 
 private:
     std::wstring m_sDescription;
```
SyrianBallaS: We could make this a literal type of we made this `const wchar_t*`, but it seems to cause problems.

src/RA_Dlg_MemBookmark.h
```diff
@@ -3,42 +3,92 @@
...
+    bool m_bFrozen = false;
+    bool m_bDecimal = false;
+
+    // Nonmember helper functions to not need a map
+    friend _CONSTANT_FN operator==(const MemBookmark& a, const MemBookmark& b) noexcept;
+    friend _CONSTANT_FN operator==(const MemBookmark& value, unsigned int key) noexcept;
+    friend _CONSTANT_FN operator==(unsigned int key, const MemBookmark& value) noexcept;
+    friend _CONSTANT_FN operator<(const MemBookmark& a, const MemBookmark& b) noexcept;
+    friend _CONSTANT_FN operator<(unsigned int key, const MemBookmark& value) noexcept;
+    friend _CONSTANT_FN operator<(const MemBookmark& value, unsigned int key) noexcept;
```
SyrianBallaS: These operators give the vector in the dialog similar capability as a map.

src/RA_Dlg_MemBookmark.h
```diff
@@ -51,56 +101,62 @@ class Dlg_MemBookmark
...
-    const MemBookmark* FindBookmark(const ra::ByteAddress& nAddr) const
+    auto& FindBookmark(const ra::ByteAddress& nAddr) const
     {
-        std::map<ra::ByteAddress, std::vector<const MemBookmark*>>::const_iterator iter = m_BookmarkMap.find(nAddr);
-        return(iter != m_BookmarkMap.end() && iter->second.size() > 0) ? iter->second.back() : nullptr;
+        static const MemBookmark dummyBookmark;
+        if (BookmarkExists(nAddr)) // Should be sorted when it was added (if it exists)
+            // TBD: There could be a better algorithm but this should be OK
+            return *std::find(m_vBookmarks.cbegin(), m_vBookmarks.cend(), nAddr);
+        return dummyBookmark; // throw?
```
SyrianBallaS: This basically does what was here before but uses a vector instead. A bookmark with an address of 0 is considered invalid but wasn't sure if we should've thrown instead.

src/RA_Dlg_MemBookmark.h
```diff
@@ -51,56 +101,62 @@ class Dlg_MemBookmark
...
 
+    void Sort() { std::sort(m_vBookmarks.begin(), m_vBookmarks.end()); }
+
 private:
-    HWND m_hMemBookmarkDialog;
-    std::map<ra::ByteAddress, std::vector<const MemBookmark*>> m_BookmarkMap;
-    std::vector<MemBookmark*> m_vBookmarks;
+    int m_nNumOccupiedRows{};
+    HWND m_hMemBookmarkDialog{};
+    std::vector<MemBookmark> m_vBookmarks;
```
SyrianBallaS: The map is gone, the class no throw `DefaultConstructible `now.

Also, the value_type should never be a pointer unless you have no choice, we have a choice here.

src/RA_Dlg_Memory.cpp
```diff
@@ -1777,10 +1781,17 @@ bool Dlg_Memory::GetGameMemoryRange(ra::ByteAddress& start, ra::ByteAddress& end
 
 static TCHAR* ParseAddress(TCHAR* ptr, ra::ByteAddress& address) noexcept
 {
+    Expects(ptr != nullptr);
```
SyrianBallaS: Can't use `gsl::not_null` with pointer arithmetic so we gotta check for nullness on all paths.

src/RA_StringUtils.h
```diff
@@ -329,54 +329,77 @@ class StringBuilder
...
             }
-
+            Ensures(pScan != nullptr);
             pFormat = pScan;
         }
     }
 
     template<typename CharT, typename = std::enable_if_t<is_char_v<CharT>>, typename T, typename... Ts>
-    void AppendPrintf(const CharT* const pFormat, const T& value, Ts&&... args)
+    void AppendPrintf(const CharT* const restrict pFormat, const T& restrict value, Ts&&... args)
```
SyrianBallaS: `args` isn't `restrict` qualified because `value `is the first `args`.

src/RA_httpthread.cpp
```diff
@@ -380,15 +380,14 @@ BOOL RAWeb::DoBlockingImageUpload(UploadType nType, const std::string& sFilename
...
-    ra::services::ServiceLocator::GetMutable<ra::services::IThreadPool>().RunAsync([pObj]()
+    ra::services::ServiceLocator::GetMutable<ra::services::IThreadPool>().RunAsync([nType, PostData, sData]() mutable
     {
+        auto pObj = std::make_unique<RequestObject>(nType, PostData, sData);
         std::string sResponse;
         DoBlockingRequest(pObj->GetRequestType(), pObj->GetPostArgs(), sResponse);
 
         pObj->SetResponse(sResponse);
-        ms_LastHttpResults.PushItem(pObj);
+        ms_LastHttpResults.PushItem(std::move(pObj)); // pointer gets corrupted if released here
```
SyrianBallaS: Made a `unique_ptr `overload. Nothing leaks. We can't `delete` here since it'll cause heap corruption and it's confusing, this is more simple.

src/RA_httpthread.cpp
```diff
@@ -431,6 +430,15 @@ void HttpResults::PushItem(RequestObject* pObj)
     ReleaseMutex(RAWeb::Mutex());
 }
 
+void HttpResults::PushItem(std::unique_ptr<RequestObject> pObj)
```
SyrianBallaS: `unique_ptr `can't be copied it's passed by value on purpose, a move will be forced to transfer ownership.

src/api/impl/ConnectedServer.cpp
```diff
@@ -159,11 +159,11 @@ Login::Response ConnectedServer::Login(const Login::Request& request) noexcept
 Logout::Response ConnectedServer::Logout(_UNUSED const Logout::Request& /*request*/) noexcept
 {
     // update the global API pointer to a disconnected API
-    ra::services::ServiceLocator::Provide<ra::api::IServer>(new (std::nothrow) DisconnectedServer(m_sHost));
+    ra::services::ServiceLocator::Provide<ra::api::IServer>(std::make_unique<DisconnectedServer>(m_sHost).release());
```
SyrianBallaS: We should probably make a `std::unique_ptr` overload to explicitly show we are transferring ownership of the pointer.

src/pch.h
```diff
@@ -64,6 +64,9 @@
 #include <rapidjson\istreamwrapper.h>
 #include <rapidjson\ostreamwrapper.h>
 #include <rapidjson\writer.h> // has stringbuffer.h
+
+/* gsl stuff */
+#include <gsl\gsl>
```
SyrianBallaS: Maybe they fixed it but one CppCoreGuideline warning is triggered from their `make_not_null `function.

src/ra_type_traits.h
```diff
@@ -143,7 +143,7 @@ namespace detail {
 template<typename T>
 struct _NODISCARD is_literal_type
     : std::bool_constant<(std::is_scalar_v<T> || std::is_reference_v<T>) ||
-                         (std::is_aggregate_v<T> && std::is_scalar_v<typename T::value_type>) ||
+                         (std::is_aggregate_v<T>) ||
```
SyrianBallaS: I was originally trying to make MemBookmark a literal type but doesn't have value_type type member like all STL style  classes.

src/ra_utility.h
```diff
@@ -10,23 +10,17 @@
...
-#pragma warning(push)
-// TODO: Finish narrow_cast in another PR
-// we don't use gsl::narrow_cast, ra::narrow_cast won't trigger this when its
-// done
-#pragma warning(disable : 26472)
 template<typename SignedType, typename = std::enable_if_t<std::is_signed_v<SignedType>>>
 _NODISCARD _CONSTANT_FN to_unsigned(_In_ SignedType st) noexcept
 {
-    return static_cast<std::make_unsigned_t<SignedType>>(st);
+    return gsl::narrow_cast<std::make_unsigned_t<SignedType>>(st);
```
SyrianBallaS: we gotta use gsl::narrow_cast instead of gsl::narrow because gsl::narrow throws (meaning not a constant expression).

src/RA_StringUtils.h
```diff
@@ -329,54 +329,77 @@ class StringBuilder
         while (*pFormat)
         {
             auto* pScan = pFormat;
+            Expects(pScan != nullptr);
```
SyrianBallaS: `T` couldn't be deduced, even explicitly, so we had to check for nullness on all paths.

src/RA_User.cpp
```diff
@@ -83,7 +83,7 @@ void LocalRAUser::RequestFriendList()
 
 RAUser* LocalRAUser::AddFriend(const std::string& sUser, unsigned int nScore)
 {
-    RAUser* pUser = RAUsers::GetUser(sUser);
+    const auto pUser = gsl::make_not_null(RAUsers::GetUser(sUser));
```
SyrianBallaS: The wrapped pointer is not a pointer to const.
Fully it would be `const gsl::not_null<RAUser*>`

src/services/SearchResults.cpp
```diff
@@ -163,7 +168,7 @@ bool SearchResults::ContainsAddress(unsigned int nAddress) const
     return false;
 }
 
-void SearchResults::AddMatches(unsigned int nAddressBase, const unsigned char pMemory[], const std::vector<unsigned int>& vMatches)
+void SearchResults::AddMatches(unsigned int nAddressBase, const unsigned char* restrict pMemory, const std::vector<unsigned int>& vMatches)
```
SyrianBallaS: pMemory will always decay into a pointer, there are other places where stuff like this was happend. Plus you can't restrict qualify a C Array.

src/services/SearchResults.cpp
```diff
@@ -173,7 +178,7 @@ void SearchResults::AddMatches(unsigned int nAddressBase, const unsigned char pM
         m_vMatchingAddresses.push_back(nAddressBase + nMatch);
 }
 
-void SearchResults::ProcessBlocks(const SearchResults& srSource, std::function<bool(unsigned int nIndex, const unsigned char pMemory[], const unsigned char pPrev[])> testIndexFunction)
+void SearchResults::ProcessBlocks(const SearchResults& srSource, std::function<bool(unsigned int nIndex, const unsigned char* restrict pMemory, const unsigned char* restrict pPrev)> testIndexFunction)
```
SyrianBallaS: Gives an optimization asserting `pMemory `and `pPrev `are not the same.

tests/mocks/MockServer.hh
```diff
@@ -12,26 +12,25 @@ namespace mocks {
...
-    /// Callback should return <c>true</c> if it populated the response, <c>false</c> to return the default response (unsupported)
+    /// Callback should return <c>true</c> if it populated the response, <c>false</c> to return the default response
+    /// (unsupported)
     /// </remarks>
     template<typename TApi>
     void HandleRequest(std::function<bool(const typename TApi::Request&, typename TApi::Response&)>&& fHandler)
     {
-        m_mHandlers.insert_or_assign(std::string(TApi::Name()), [fHandler=std::move(fHandler)](void* pRequest, void* pResponse)
+        m_mHandlers.insert_or_assign(std::string(TApi::Name()), [fHandler = std::move(fHandler)](
+                                                                    void* restrict pRequest, void* restrict pResponse)
```
SyrianBallaS: Asserts `pRequest` and` pResponse` are not the same.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -20,437 +22,434 @@ namespace ra {
...
+    {
+        if (vk == VK_RETURN || vk == VK_ESCAPE)
+            DestroyWindow(hwnd); // Causing a killfocus :)
+    };
+
+    switch (nMsg)
+    {
+        HANDLE_MSG(hwnd, WM_DESTROY, OnDestroy);
+        HANDLE_MSG(hwnd, WM_KILLFOCUS, OnKillFocus);
+        HANDLE_MSG(hwnd, WM_KEYDOWN, OnKeyDown);
```
Jamiras: These changes contradict the changes made in DialogBase and it's children and will be replaced by that logic when this dialog is converted. Also these changes are pretty drastic - I hope you tested them thoroughly.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -20,437 +22,434 @@ namespace ra {
...
+    };
+
+    // id->LOWORD(wParam), hwndCtl->lParam, codeNotify->HIWORD(wParam)
+    const auto OnCommand = [this](HWND hDlg, int id, [[maybe_unused]] HWND /*hwndCtl*/,
+                            [[maybe_unused]] UINT /*codeNotify*/) {
+        switch (id)
         {
-            switch (LOWORD(wParam))
+            case IDOK:
+                [[fallthrough]];
```
Jamiras: explicit fallthrough comment is not necessary if no other code is associated with the case statement. it's clear that multiple cases will execute the same code.

src/RA_LeaderboardPopup.h
```diff
@@ -18,6 +18,7 @@ class LeaderboardPopup
 
 public:
     void Update(_In_ float fDelta);
+    GSL_SUPPRESS(con.4) // line 249
```
Jamiras: putting line numbers in comments is bad practice as code shifts around frequently. in this case, it's not even in the same file.

src/RA_MemManager.cpp
```diff
@@ -103,23 +103,27 @@ unsigned char MemManager::ActiveBankRAMByteRead(ra::ByteAddress nOffs) const
...
-    const int numBanks = m_Banks.size();
+    auto bankID = 0U;
+    const auto numBanks = m_Banks.size();
     while (bankID < numBanks)
     {
         bank = &m_Banks.at(bankID);
+        Expects(bank != nullptr);
         if (nOffs < bank->BankSize)
+        {
+            Ensures(bank != nullptr);
```
Jamiras: Clearly I need a tutorial on Expects and Ensures. Expects clearly behaves like an assert, but I'm assuming it also throws an execption on a release build. What does Ensures do? We just validated it wasn't null two lines ago, why would it be null here?

src/RA_StringUtils.h
```diff
@@ -11,8 +11,8 @@ _NODISCARD std::string Narrow(_In_z_ const wchar_t* wstr);
 _NODISCARD std::wstring Widen(_In_ const std::string& str);
 _NODISCARD std::wstring Widen(_In_z_ const char* str);
 
-[[gsl::suppress(f.6), nodiscard]] std::string Narrow(std::wstring&& wstr) noexcept;
-[[gsl::suppress(f.6), nodiscard]] std::wstring Widen(std::string&& str) noexcept;
+[[ gsl::suppress(f .6), nodiscard ]] std::string Narrow(std::wstring&& wstr) noexcept;
+[[ gsl::suppress(f .6), nodiscard ]] std::wstring Widen(std::string&& str) noexcept;
```
Jamiras: I'm assuming the auto-formatter changed these. Is the space between "f" and ".6" going to cause problems?

src/api/impl/ConnectedServer.cpp
Jamiras: The old structure was most similar to constructing a `unique_ptr` instance:
```
unique_ptr<ra::api::IServer> foo(new DisconnectedServer(m_sHost));
```
This seems excessively verbose without any benefit.

I can see that having `Provide` wrap a `unique_ptr` may be confusing, would changing the method name help? `ProvideUnique<T>`?

src/api/impl/ServerBase.hh
```diff
@@ -14,37 +14,38 @@ class ServerBase : public IServer
...
     }
 
-    GSL_SUPPRESS(f.6) Logout::Response Logout(_UNUSED const Logout::Request& /*request*/) noexcept override
+    GSL_SUPPRESS(f .6) Logout::Response Logout(_UNUSED const Logout::Request& /*request*/) noexcept override
     {
         return UnsupportedApi<Logout::Response>(Logout::Name());
     }
 
-    GSL_SUPPRESS(f.6) StartSession::Response StartSession(_UNUSED const StartSession::Request& /*request*/) noexcept override
+    GSL_SUPPRESS(f .6)
```
Jamiras: what is doing this, and won't it cause problems?

src/services/Initialization.cpp
```diff
@@ -56,19 +56,17 @@ void Initialization::RegisterCoreServices()
...
-
-    auto* pFileSystem = new ra::services::impl::WindowsFileSystem();
-    ra::services::ServiceLocator::Provide<ra::services::IFileSystem>(pFileSystem);
-
-    auto* pLogger = new ra::services::impl::WindowsDebuggerFileLogger(*pFileSystem);
-    ra::services::ServiceLocator::Provide<ra::services::ILogger>(pLogger);
-
+    auto pClock = std::make_unique<ra::services::impl::Clock>();
+    auto pFileSystem = std::make_unique<ra::services::impl::WindowsFileSystem>();
+    auto pLogger = std::make_unique<ra::services::impl::WindowsDebuggerFileLogger>(*pFileSystem);
```
Jamiras: why change these but not everything in `RegisterServices` below?

src/RA_Dlg_MemBookmark.cpp
SyrianBallaS: Sure it's fine, the compiler doesn't warn on fallthrough so I guess it doesn't really matter.

src/RA_LeaderboardPopup.h
SyrianBallaS: I'll change it

src/RA_MemManager.cpp
SyrianBallaS: It's for pre/post-conditions. If it fails the program terminates. They are basically the same. The rule says you have to test it for nullness on all paths. I'll see if wrapping the whole thing in an if statement does anything to suppress it.

src/RA_StringUtils.h
SyrianBallaS: I don't think it does, I ran the analyzer a few times with the existing rules and they didn't pop up.

src/api/impl/ConnectedServer.cpp
SyrianBallaS: It just warned on using `new` and not deleting it.

I'd think that the `ProvideUnique<T>` should have a unique pointer as a param to show an explicit transfer of ownership, then it releases it's owned pointer to the service container.

Kind of like this:
```
auto pDisconnectedServer = std::make_unique<DisconnectedServer>(m_sHost);
// The param would be by value since it can't be copied
 ra::services::ServiceLocator::Provide<ra::api::IServer>(std::move(pDisconnectedServer ));
```
Inside the `Provide<T>`, the owned pointer would be released.
If you want to call it `ProvideUnique `I guess that's fine.

src/api/impl/ServerBase.hh
SyrianBallaS: I'll need to see what's doing it. I don't think it's causing problems though. Better to be safe than sorry though.

src/services/Initialization.cpp
SyrianBallaS: It didn't catch them, I can change them.

SyrianBallaS:
This might take me awhile to address so I'm closing it.

SyrianBallaS:
I believe this is ready, gonna leave some review comments.

src/services/ServiceLocator.hh
```diff
@@ -58,9 +58,9 @@ public:
     /// pointer and delete it at the end of execution or when a different implementation is provided.
     /// </param>
     template <class TClass>
-    static void Provide(TClass* pInstance) noexcept
+    static void Provide(std::unique_ptr<TClass> pInstance) noexcept
     {
-        Service<TClass>::s_pInstance.reset(pInstance);
+        Service<TClass>::s_pInstance = std::move(pInstance);
```
SyrianBallaS: I just made it like this by default. If `s_pInstance `is a `unique_ptr`, it makes sense that `pInstance` is also a `unique_ptr`. As far as this method is concerned it's leak proof with no "`new`"s anywhere. No copying was done at all.

src/RA_Achievement.cpp
```diff
@@ -16,13 +16,19 @@
 #include "RA_Core.h"
 #else
 // duplicate of code in RA_Core, but RA_Core needs to be cleaned up before it can be pulled into the unit test build
-void _ReadStringTil(std::string& value, char nChar, const char*& pSource)
+/* clang-format off */
+GSL_SUPPRESS(f.23)
+/* clang-format on */ // tested for nullness but duplication is confusing the analyzer
```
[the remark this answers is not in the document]
Jamiras: I don't like this solution - particularly having to do it everywhere. Can we just create a macro for the macro and use that instead?
```
/* clang-format off */
#define GSL_SUPPRESS_F6 GSL_SUPPRESS(f.6)
#define GSL_SUPPRESS_F23 GSL_SUPPRESS(f.23)
/* clang-format on */
```

src/RA_AchievementSet.h
```diff
@@ -35,19 +31,28 @@ class AchievementSet
...
-    _Success_(return)
-    bool LoadFromFile(_Inout_ unsigned int nGameID);
+    _Success_(return ) bool LoadFromFile(_Inout_ unsigned int nGameID);
     bool SaveToFile() const;
 
     //	Get Achievement at offset
-    Achievement& GetAchievement(size_t nIter) { return m_Achievements[nIter]; }
+    Achievement& GetAchievement(size_t nIter)
+    {
+        Ensures((nIter >= 0) && (nIter < m_Achievements.size()));
```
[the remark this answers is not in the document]
Jamiras: Shouldn't this be Expects? nothing has happened yet, so it's not "post-" anything.

src/RA_Core.cpp
```diff
@@ -495,18 +495,24 @@ API void CCONV _RA_ClearMemoryBanks()
...
 {
-    char* pPart;
+    char* pPart{};
     const auto major = strtoull(sVersion, &pPart, 10);
+    Expects(pPart != nullptr);
     if (*pPart == '.')
+    {
         ++pPart;
-
+    }
```
Jamiras: these braces are no longer necessary.

src/RA_Core.cpp
```diff
@@ -1310,26 +1324,37 @@ BOOL _ReadTil(const char nChar, char* buffer, unsigned int nSize, DWORD* pCharsR
...
 char* _ReadStringTil(char nChar, char* restrict& pOffsetInOut, BOOL bTerminate) noexcept
 {
+    Expects(pOffsetInOut != nullptr);
     char* pStartString = pOffsetInOut;
 
     while ((*pOffsetInOut) != '\0' && (*pOffsetInOut) != nChar)
+    {
         pOffsetInOut++;
-
+        Ensures(pOffsetInOut != nullptr);
```
Jamiras: This will only be true if `pOffsetInOut` is `0xFFFFFFFF` before it's incremented. Is the tool telling you to add all of these things? I don't like it. I'm fine with null assertions on input parameters, or function returns, but throwing these around (especially when it's clear they'll never be false) is just polluting the code.

Jamiras: Additionally, `assert` is compiled out of a release build. Terminating the app is arguably less useful than throwing an uncaught exception. At least with an exception you have some idea why the app terminated.

src/RA_Dlg_MemBookmark.h
```diff
@@ -51,56 +101,69 @@ class Dlg_MemBookmark
...
+    inline auto AddBookmark(MemBookmark&& newBookmark)
+    {
+        if (BookmarkExists(newBookmark.Address()))
+        {
+            auto duplicate = std::move(newBookmark); // move it so it can be destroyed
+            ::MessageBox(::GetActiveWindow(), _T("A bookmark with the same address has already been added"),
+                         _T("Error!"), MB_OK | MB_ICONERROR);
+            return;
+        }
+        m_vBookmarks.push_back(std::move(newBookmark));
```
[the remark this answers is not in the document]
Jamiras: I wouldn't worry about manual sorting now. We can leave that for the ViewModel implementation - it's requested in other places as well.

src/RA_Dlg_MemBookmark.h
Jamiras: You could theorhetically bookmark address 0, it shows up in the inspector. Prefer using MAX_INT instead.

src/RA_Dlg_Memory.cpp
```diff
@@ -324,9 +320,9 @@ bool MemoryViewerControl::OnEditInput(UINT c)
 
         if (g_MemBookmarkDialog.GetHWND() != nullptr)
         {
-            const MemBookmark* Bookmark = g_MemBookmarkDialog.FindBookmark(nByteAddress);
-            if (Bookmark != nullptr)
-                g_MemBookmarkDialog.WriteFrozenValue(*Bookmark);
+            const auto& Bookmark{g_MemBookmarkDialog.FindBookmark(nByteAddress)};
+            if (Bookmark.Address() != 0) // 0 is the value for dummyBookmark
```
Jamiras: magic numbers should have constants

src/RA_Dlg_Memory.cpp
```diff
@@ -1775,12 +1775,19 @@ bool Dlg_Memory::GetGameMemoryRange(ra::ByteAddress& start, ra::ByteAddress& end
...
+        return ptr;
+
     if (*ptr == '$')
+    {
         ++ptr;
+    }
     else if (ptr[0] == '0' && ptr[1] == 'x')
+    {
         ptr += 2;
+    }
```
Jamiras: more unnecessarily added braces

src/RA_MemManager.cpp
```diff
@@ -101,25 +96,26 @@ unsigned int MemManager::ActiveBankRAMRead(ra::ByteAddress nOffs, MemSize size)
...
-    int bankID = 0;
-    const int numBanks = m_Banks.size();
+    auto bankID = 0U;
+    const auto numBanks = m_Banks.size();
     while (bankID < numBanks)
     {
-        bank = &m_Banks.at(bankID);
+        const auto bank = &m_Banks.at(bankID);
+        if (bank == nullptr)
+            continue;
```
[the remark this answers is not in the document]
Jamiras: The API allows the caller to provide the bank key, so we can't assume they're contiguous. I think this is leftover from a design where the player could select active banks individually (method name is _ActiveBank_ RAMByteRead.)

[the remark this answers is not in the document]
Jamiras: `bank` should never be null. if `at()` can't find a key, it'll through an `out_of_range` exception.

src/RA_StringUtils.h
```diff
@@ -329,41 +333,59 @@ class StringBuilder
...
+        if (pScan == nullptr)
+        {
+            ::MessageBox(::GetActiveWindow(), _T("Invalid format string!"), _T("Error!"), MB_OK | MB_ICONERROR);
+            return;
+        }
+
         while (*pScan && *pScan != '%')
+        {
             ++pScan;
+        }
```
Jamiras: more unnecessarily added braces

src/RA_StringUtils.h
```diff
@@ -329,41 +333,59 @@ class StringBuilder
...
     }
 
     template<typename CharT, typename = std::enable_if_t<is_char_v<CharT>>, typename T, typename... Ts>
-    void AppendPrintf(const CharT* const pFormat, const T& value, Ts&&... args)
+    void AppendPrintf(const CharT* const restrict pFormat, const T& restrict value, Ts&&... args)
     {
         auto* pScan = pFormat;
+        if (pScan == nullptr)
+        {
+            ::MessageBox(::GetActiveWindow(), _T("Invalid format string!"), _T("Error!"), MB_OK | MB_ICONERROR);
```
Jamiras: `MessageBox` should not be called directly. Use `MessageBoxViewModel` instead. However, in this particular case, I don't think it should be called at all. The user can't do anything with this, so why tell them? Recommend guarding with an `assert` or `Expects`.

src/RA_User.cpp
Jamiras: `GetUser` will allocate a user if one is not found. It's guaranteed to not return null. Would changing the signature to return "`RAUser&`" satisfy the tool?

src/RA_httpthread.cpp
Jamiras: Should be able to eliminate the other `PushItem` implementation. I think this is the only place it's called.

tests/mocks/MockLocalStorage.hh
```diff
@@ -48,7 +48,8 @@ public:
 
     std::unique_ptr<TextReader> ReadText(StorageItemType nType, const std::wstring& sKey) override
     {
-        std::string* pText = GetText(nType, sKey, false);
+		// std::string* const
```
Jamiras: extraneous comment

tests/services/AchievementRuntime_Tests.cpp
```diff
@@ -180,8 +182,8 @@ TEST_CLASS(AchievementRuntime_Tests)
         runtime.mockGameContext.AddAchivement(3U, "1=1.10.");
         runtime.mockGameContext.AddAchivement(5U, "1=1.2.");
 
-        auto* pAchievement3 = runtime.mockGameContext.FindAchievement(3U);
-        auto* pAchievement5 = runtime.mockGameContext.FindAchievement(5U);
+        const auto pAchievement3 = gsl::make_not_null(runtime.mockGameContext.FindAchievement(3U));
+        const auto pAchievement5 = gsl::make_not_null(runtime.mockGameContext.FindAchievement(5U));
```
Jamiras: Why remove the * from auto? It makes it more clear what the variable type is. It also prevents possible problems if the implicit type changes to not be a pointer anymore. The compiler will automatically call the copy constructor, and as long as the -> operator is still valid, you won't even know anything has changed.

src/RA_Core.cpp
SyrianBallaS: got rid of them

src/RA_Core.cpp
SyrianBallaS: exceptions are enabled

src/RA_Dlg_Memory.cpp
SyrianBallaS: done

src/RA_StringUtils.h
SyrianBallaS: I used the ViewModel in other places where MessageBox was used

tests/services/AchievementRuntime_Tests.cpp
SyrianBallaS: auto is relpaced with the type name to show the actual class is not a pointer, it just wraps a pointer to enforce it can't be null.

PR #222, opened 2018-12-26
Microsoft Recommended Native Rules

SyrianBallaS:
This one didn't need that much but will leave some review comments.

**Rules Addressed:**
- Function specification not adhered
  - This happens when a variable used does not satisfy the SAL annotation for a parameter
- inconsistent annotations
  - Mostly when `_Use_decl_annotations_ `wasn't used on the definition even though there were SAL annotations in the declaration.
- `ES.84: Don't (try to) declare a local variable with no name`
- `io.5/ES.106: Don't try to avoid negative values by using unsigned`
   - Mostly suppressed 
   - Visual Studio calls it "io.5" but it's actually es.106
- `Type.6: Always initialize a member variable: always initialize, possibly using default constructors or default member initializers.`
   - Sometimes suppressed


src/RA_Dlg_AchEditor.cpp
```diff
@@ -300,7 +304,8 @@ LRESULT CALLBACK DropDownProc(HWND hwnd, UINT nMsg, WPARAM wParam, LPARAM lParam
             ZeroMemory(&lvDispinfo, sizeof(LV_DISPINFO));
             lvDispinfo.hdr.hwndFrom = hwnd;
             lvDispinfo.hdr.idFrom = GetDlgCtrlID(hwnd);
-            lvDispinfo.hdr.code = LVN_ENDLABELEDIT;
+#pragma warning(suppress: 26454)
+            GSL_SUPPRESS(io.5) lvDispinfo.hdr.code = LVN_ENDLABELEDIT;
```
SyrianBallaS: `LVN_ENDLABELEDIT` is a negative `UINT`, I have no idea what the design behind it means but just suppressed it. It does show up as` MAX_UINT` but almost everything in the file it came from is as well. The API probably does some weird stuff with it or uses the name instead of the value.

src/RA_Dlg_MemBookmark.cpp
```diff
@@ -310,6 +311,7 @@ INT_PTR Dlg_MemBookmark::MemBookmarkDialogProc(HWND hDlg, UINT uMsg, WPARAM wPar
                         if (nSelect == -1)
                             break;
                     }
+#pragma warning(suppress: 26454) // io.5; arithmetic overflow
                     else if (((LPNMHDR)lParam)->code == NM_DBLCLK)
```
SyrianBallaS: You can't put C++ attributes (at least currently) before an `else if`, so it uses the warning code instead.

src/ra_rules.ruleset
```diff
@@ -1,11 +1,21 @@
...
-  <Include Path="cppcorecheckconstrules.ruleset" Action="Default" />
-  <Include Path="cppcorecheckfuncrules.ruleset" Action="Default" />
-  <Include Path="cppcorechecksharedpointerrules.ruleset" Action="Default" />
-  <Include Path="cppcorecheckuniquepointerrules.ruleset" Action="Default" />
+  <Include Path="cppcorecheckclassrules.ruleset" Action="Warning" />
+  <Include Path="cppcorecheckconstrules.ruleset" Action="Warning" />
+  <Include Path="cppcorecheckfuncrules.ruleset" Action="Warning" />
+  <Include Path="cppcorechecksharedpointerrules.ruleset" Action="Warning" />
+  <Include Path="cppcorecheckuniquepointerrules.ruleset" Action="Warning" />
+  <Include Path="nativerecommendedrules.ruleset" Action="Warning" />
```
SyrianBallaS: I was originally trying to see if I could get these to show up without having a huge list at the bottom but it doesn't work. It shows up as checked in the UI but I guess the compiler doesn't understand it.

src/ui/drawing/gdi/GDIBitmapSurface.cpp
```diff
@@ -147,6 +147,7 @@ void GDIAlphaBitmapSurface::Blend(HDC hTargetDC, int nX, int nY) const noexcept
 
     UINT32* pBits;
     HBITMAP hBitmap = CreateDIBSection(hMemDC, &bmi, DIB_RGB_COLORS, reinterpret_cast<LPVOID*>(&pBits), nullptr, 0);
+    Expects(hBitmap != nullptr);
     SelectBitmap(hMemDC, hBitmap);
```
SyrianBallaS: Example of not adhering to specification. In `SelectObject`, the `HDIOBJ `is an "in" param but `hBitmap `could be a `nullptr`, so we have to make it a requirement that it's not a `nullptr`.

src/ui/drawing/gdi/GDIBitmapSurface.hh
```diff
@@ -15,10 +15,12 @@ namespace gdi {
 class GDIBitmapSurface : public GDISurface
 {
 public:
+    GSL_SUPPRESS(type.6) // m_hBitmap not initizalied, adhered to note below
     explicit GDIBitmapSurface(const int nWidth, const int nHeight, ResourceRepository& pResourceRepository) noexcept :
         GDISurface(CreateBitmapHDC(nWidth, nHeight), RECT{0, 0, nWidth, nHeight}, pResourceRepository)
     {}
```
SyrianBallaS: I saw the comment near the data members so I suppressed the rules for these constructors.

tests/mocks/MockClock.hh
```diff
@@ -15,7 +15,7 @@ public:
     MockClock() noexcept : m_Override(this)
     {
         // force GMT timezone for unit tests
-        _putenv("TZ=GMT");
+        Expects(_putenv("TZ=GMT") == 0);
```
SyrianBallaS: `_putenv `has a `_Check_return_` annotation which is the SAL equivalent to the C++ `[[nodiscard]]` attribute. Except `_Check_return_` is only active during code analysis and `[[nodiscard]]` is always active.

tests/mocks/MockFileSystem.hh
```diff
@@ -130,13 +130,13 @@ public:
             return false;
 
         hNode.key() = sNewPath;
-        m_mFileContents.insert(std::move(hNode));
+        (void)m_mFileContents.insert(std::move(hNode));
```
SyrianBallaS: It'll say "avoid unnamed objects with custom construction and destruction (RAII)".

Even if an RAII object does not have a `[[nodiscard]]` there shouldn't be an unnamed RAII object. The standard sanctioned why to suppress it is to cast to `void`. I'm not sure if a cast to void will trigger "Don't use C-Style casts" when the type profile is run, if it does

src/RA_LeaderboardPopup.cpp
```diff
@@ -273,9 +276,9 @@ _Use_decl_annotations_ void LeaderboardPopup::Render(ra::ui::drawing::ISurface&
             if (m_pScoreboardSurface != nullptr)
             {
                 const auto fOffscreenAmount = (GetOffsetPct() * 600);
-                const auto fFadeOffs = (nWidth - 300 - 10) + fOffscreenAmount;
-                const int nScoreboardX = static_cast<int>(fFadeOffs);
-                const int nScoreboardY = nHeight - 200 - 10;
+                const auto fFadeOffs = (gsl::narrow<std::int64_t>(nWidth - 300 - 10) + fOffscreenAmount);
+                const auto nScoreboardX = ra::ftoi(fFadeOffs);
```
Jamiras: what does the `narrow` do here? it looks like it's trying to convert the float to an int64, which implies the `ftoi` on the next line is unnecessary.

src/RA_RichPresence.cpp
```diff
@@ -193,7 +193,7 @@ static bool GetLine(ra::services::TextReader& pReader, std::string& sLine)
         if (index != std::string::npos)
         {
             // if a comment marker was found, remove it and any trailing whitespace
-            while (index > 0 && isspace(sLine[index - 1]))
+            while (index > 0 && isspace(gsl::narrow<int>(sLine.at(index - 1))))
```
Jamiras: `sLine` is a string, so an element of `sLine` is a character. `narrow` seems like an inaccurate conversion to get to `int`.

src/RA_LeaderboardPopup.cpp
SyrianBallaS: It's for narrowing conversions. Since double is 8 bytes, and nwidth is 4, it needs to be cast yo a wider type to prevent overflow

src/RA_RichPresence.cpp
SyrianBallaS: At returns a char, would it be better to cast it to unsigned char instead?

src/RA_LeaderboardPopup.cpp
SyrianBallaS: The code wont compile if it wasn't necessary as the function  requires a floating point number at compile time

src/RA_RichPresence.cpp
SyrianBallaS: I'll look up a better way to do this. Narrow is basically a static cast that that throws if casting changes the value

SyrianBallaS:
Reopening this later.

src/RA_LeaderboardPopup.cpp
SyrianBallaS: The int needs to be cast to a wider type to prevent overflow.
This warning comes up:
`Warning	C26451 Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2)`

_The rule ids are not the same because "io" is used for `SL.io: Iostream` in the guidelines._
Here's what it says in the C++ Core Guidelines (linked from MS Docs)
```
ES.103: Don't overflow
Reason
Overflow usually makes your numeric algorithm meaningless. Incrementing a value beyond a maximum value can lead to memory corruption and undefined behavior.
```

src/RA_StringUtils.h
```diff
@@ -394,7 +394,7 @@ class StringBuilder
                 {
                     const char c = gsl::narrow<char>(*pScan);
                     sFormat.push_back(c);
-                    if (isalpha(c))
+                    if (isalpha(gsl::narrow<int>(c)))
```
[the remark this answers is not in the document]
SyrianBallaS: Done

tests/mocks/MockClock.hh
SyrianBallaS: Done

src/RA_LeaderboardPopup.cpp
SyrianBallaS: I dont know. Its complained before about multiplying and division (they changed since then do wont show up now). I'm pretty sure its complaint about one operand having a smaller byte width which makes to me. It doesnt seem immediately apparent unless it's with binary format.

I will try your suggestion anyway.

SyrianBallaS: Your suggestion worked, I'm going to keep that in mind for the future. I also added the Arithmetic rules for CppCoreCheck since it's completely including in this one. Mainly so we don't have to run it.

Issue #178, opened 2018-11-11
Remove Scan for Games and Check for Emu Update from RetroAchievments titlebar


kdecks:
![image](https://user-images.githubusercontent.com/32706333/48311915-1fee4780-e564-11e8-9106-8d30d94b178e.png)

As far as I know, **scan for games** has never worked, and I'm not certain what the intention was for it. Perhaps to create a list of supported games in the overlay, or some other menu. Regardless it does not work, and so either it should be fixed or pruned.

**Check for Emulator** update is also unnecessary. The emulator will prompt you automatically at startup if an update is available.

SyrianBallaS:
The main reason "Scan for games" doesn't work is because it only checks for Genesis ROMs (*.bin;*.gen). It should be relatively easy to fix. I was just waiting on the new rehash strategy to go live before doing this since hashes were always mismatched (rehash was different than Server MD5) anyway.

kdecks:
@SyrianBallaS 

What happens when **Scan for games** works as intended?

As you see it what value might **Check for Emulator Update** have?

SyrianBallaS:
I actually dont know is intended for it as there was no documentation on it and was there way before I started contributing. 

I can only guess it scans for roms and populates the list of it matches gamehashlibrary.txt in the racache folder. 
I was able to get yo scan the files but the md5 never matches.

I don't really think there's any value in checking for updates manually if it's already done.

I don't feel comfortable getting rid of all this.

kdecks:
>I can only guess it scans for roms and populates the list of it matches gamehashlibrary.txt in the racache folder.
If one can get it to work then it offers utility.

>I don't really think there's any value in checking for updates manually if it's already done.

In my opinion if these features offers little or no utility then keeping them is a detriment; they are just visual garbage, or vestigial. I can't speak for @ScottFromDerby but I suspect he would agree.

Jamiras:
I've removed the "Check for Emulator Updates" menu item in PR225. I've also commented out the "Scan for Games" menu item. The emulators are currently responsible for determining the proper way to hash their ROMs. Duplicating that logic in the toolkit seems undesirable (it's already duplicated in RALibRetro and RetroArch - but that's a topic for another discussion). I think there could be benefit in the feature if it worked, so I haven't removed all of the code associated to it yet, and as such, will leave this open even after 225 is merged.

kdecks:
Here's what @ScottFromDerby said about it when I asked him about it near the beginning of this month:

[...]

PR #225, opened 2018-12-27
Require latest emulator version for hardcore mode

Jamiras:
* Version is validated when the user logs in, and when switching to hardcore mode.

* On initial startup, if the emulator version is not current, and the user is not in hardcore mode, they are presented with this dialog:
image
  * Pressing Yes opens the download page and closes the dialog.
  * Pressing No just closes the dialog.
  * In both cases, the user is allowed to open games and proceed in non-hardcore mode.
* On initial startup, if the emulator version is not current, and the user is in hardcore mode, they are presented with this dialog:
image
  * Pressing OK opens the download page, logs the user out, and closes the dialog. Since the user is logged out, achievements will not load.
  * Pressing Cancel closes the dialog and disables hardcore mode. The player is allowed to proceed in non-hardcore mode.
* When switching to hardcore mode, if the emulator version is not current, the user is presented with the second dialog.
  * Pressing OK opens the download page, logs the user out, and closes the dialog. Since the user is logged out, achievements will not trigger.
  * Pressing Cancel closes the dialog and leaves the user in non-hardcode mode.
* If the user chooses to log in after initial startup, and the emulator version is not current, and hardcore mode is enabled, the user is presented with the second dialog.
  * Pressing OK opens the download page, cancels the login, and closes the dialog. Since the user is logged out, achievements will not trigger.
  * Pressing Cancel closes the dialog, disables hardcore, and completes the login.
* If the emulator version is current, no dialog is shown for any of the above scenarios.

* `g_EmulatorId`, `g_sClientVersion`, and `g_sClientName` are now properties on `EmulatorContext`.
* `g_sClientDownloadURL` and `g_sClientEXEName` were removed. They were only referenced in commented out code.

* This PR also removes the "Check for Emulator Updates" menu item and hides the "Scan for Games" menu item (addresses #178).

### DIFF
--- a/src/Exports.cpp
+++ b/src/Exports.cpp
@@ -6,6 +6,7 @@
 
 #include "api\Login.hh"
 
+#include "data\EmulatorContext.hh"
 #include "data\SessionTracker.hh"
 #include "data\UserContext.hh"
 
@@ -35,10 +36,13 @@ API int CCONV _RA_HardcoreModeIsActive()
 
 static void HandleLoginResponse(const ra::api::Login::Response& response)
 {
+    auto& pUserContext = ra::services::ServiceLocator::GetMutable<ra::data::UserContext>();
+    if (pUserContext.IsLoginDisabled())
+        return;
+
     if (response.Succeeded())
     {
         // initialize the user context
-        auto& pUserContext = ra::services::ServiceLocator::GetMutable<ra::data::UserContext>();
         pUserContext.Initialize(response.Username, response.ApiToken);
         pUserContext.SetScore(response.Score);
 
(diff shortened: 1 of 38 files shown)


rzumer:
What is the rationale for doing this? It's not like it would stop people from making changes and recompiling with the same version as far as I can tell, and it would be annoying to be barred from an older version if there is a bug in a newer one, and no change is made to the RA implementation.

Jamiras:
This isn't attempting to address users who go out of their way to make a custom build of the emulator (or the toolkit). I don't know that it's even feasible to do that with an open source project. And limiting the functionality to hardcore should give you a pretty good idea that we're just trying to enforce a "common denominator" for hardcore players. There's been a lot of talk recently about cheating, and this is just one step to help address some of the concerns.

For example, TAS was allowed in hardcore in RANes versions prior to 0.12. As it stands, there's nothing currently preventing players from running a 0.11 version (if they can find one) and earning a bunch of hardcore achievements without having to even touch the controller.

This relies on the server telling us what the lowest supported version is. If we do release a bug, we simply need to have the server say the older version is still allowed while we fix it. If the server wants to return a MinimumVersion as well as a LatestVersion, this could be modified to look for that.

rzumer:
>This isn't attempting to address users who go out of their way to make a custom build of the emulator (or the toolkit). I don't know that it's even feasible to do that with an open source project.

Without having done it myself, I can think of two relatively simple solutions: signing with a private key (it doesn't have to be in the repository), or storing lists of valid hashes generated by the binaries (using a scheme more robust to collisions than MD5). Any invalid emulator should still be able to contact the API, at least when logged in as a developer, but unlocks would simply be voided server-side.

>This relies on the server telling us what the lowest supported version is. If we do release a bug, we simply need to have the server say the older version is still allowed while we fix it.

That sounds fine for the time being.

PR #221, opened 2018-12-25
Cppcorecheck gsl rules

SyrianBallaS:
Even though it says GSL Rules it's actually a subset of the bounds profile. It's a bit more focused.

**Rule Addressed:**
`Bounds.4: Don't use standard-library functions and types that are not bounds-checked: Use the standard library in a type-safe manner.`

Probably not gonna run the bounds profile soon until I get a better understanding of spans (non-owning array that does not decay into pointer).

PR #235, opened 2019-01-04
CppCoreCheck Type Profile

SyrianBallaS:
This PR has to do with the type profile of the C++ Core Guidelines, concerning type safety. I don't expect this PR to be merged soon but thought I'd present it to see if it needs any changes. I'll fix merge conflicts when the time comes. There will be review comments after this PR is submitted (they won't show up immediately).

_*auto formatting was off the entire-time_
**Rules Addressed**:
- `ES.47: Use nullptr rather than 0 or NULL`
   - This had to be suppressed sometimes because the warning was coming from an API macro, they use the C-Style of `nullptr` (`(void*)0`), it didn't make sense to redefine it so it's suppressed.
- `Type.1: Avoid casts:` 
    - (a) `Don't use reinterpret_cast; A strict version of Avoid casts and prefer named casts.`
       -  For a lot of these I either made an intermediate message handler or a direct message handler
       -  I put some notes about transitioning them to the ViewModel Framework
       -  It had to be suppressed sometimes since it couldn't be avoided (to my knowledge)
    - (b) `Don't use static_cast for arithmetic types; A strict version of Avoid casts and prefer named casts.` 
       -  `ra::narrow_cast` is gone
       -  `gsl::narrow_cast` is used for constant expressions and for other functions that can't throw (like `ra::to_unsigned`)
      -  `gsl::narrow` is used for non-constant expressions and for other functions that might throw
    - (c) `Don't cast between pointer types where the source type and the target type are the same; A strict version of Avoid casts. `
       -  Examples are `HMODLUE` and `HINSTANCE `or `const unsigned char*` and `const BYTE*`
       -  The are exactly the same
    - (d) `Don't cast between pointer types when the conversion could be implicit; A strict version of Avoid casts.`
       -  This usually happened when casting to `void*`, all pointers can be cast to `void*` (block of memory) implicitly but `void*` cannot be cast to a pointer to object implicitly (would need `static_cast`)
- `Type.3: Don't use const_cast to cast away const (i.e., at all): Don't cast away const.`
- `Type.4: Don't use C-style (T)expression or functional T(expression) casts: Prefer construction or named casts.`
   - Mostly approached it the same way as in `type.1`
- `Type.5: Don't use a variable before it has been initialized: always initialize.`
- `Type.7: Avoid naked union: Use variant instead.`
   - The `union` in the `StringBuilder` is a variant now
   - It didn't warn about the `union` in `MemBlock` but decided not to touch it because the size requirement of `MemBlock` would be violated
   - `std::variant` is the size of all the types it holds combined.


src/RA_Achievement.cpp
```diff
@@ -213,7 +213,7 @@ void Achievement::ParseTrigger(const char* sTrigger)
     {
         // allocate space and parse again
         m_pTriggerBuffer = std::make_shared<std::vector<unsigned char>>(nSize);
-        auto* pTrigger = rc_parse_trigger(static_cast<void*>(m_pTriggerBuffer.get()->data()), sTrigger, nullptr, 0);
+        auto* pTrigger = rc_parse_trigger(m_pTriggerBuffer->data(), sTrigger, nullptr, 0);
```
SyrianBallaS: casting to `void*` can be done implicitly

src/RA_Achievement.cpp
```diff
@@ -403,9 +403,9 @@ void Achievement::SetActive(BOOL bActive) noexcept
             if (!m_bActive)
                 pRuntime.DeactivateAchievement(ID());
             else if (m_bPauseOnReset)
-                pRuntime.MonitorAchievementReset(ID(), reinterpret_cast<rc_trigger_t*>(m_pTrigger));
+                pRuntime.MonitorAchievementReset(ID(), static_cast<rc_trigger_t*>(m_pTrigger));
```
SyrianBallaS: casting to pointer to object must be done explicitly from `void*`.

src/RA_AchievementOverlay.cpp
```diff
@@ -636,41 +636,39 @@ void AchievementOverlay::DrawFriendsPage(HDC hDC, int nDX, _UNUSED int, const RE
...
+    const auto tModified = pAch->ModifiedDate();
 
     DrawAchievement(hDC, pAch, nDX + nAchievementStartX, nAchievementStartY, TRUE, FALSE);
 
-    if (m_nAchievementsSelectedItem >= (int)nNumAchievements)
+    if (m_nAchievementsSelectedItem >= ra::to_signed(nNumAchievements))
         return;
 
+    char bufTime[256]{};
+    char buffer[256]{};
```
SyrianBallaS: #221 Has stuff I was trying to do here so instead of doing that, I just initialized the array (ensures null-termination)

src/RA_AchievementOverlay.cpp
```diff
@@ -751,7 +749,7 @@ void AchievementOverlay::DrawNewsPage(HDC hDC, int nDX, _UNUSED int, const RECT&
 
     HGDIOBJ hOldObject = SelectObject(hDC, g_hFontDesc2);
 
-    for (int i = m_nNewsSelectedItem; i < static_cast<int>(m_LatestNews.size()); ++i)
+    for (int i = m_nNewsSelectedItem; i < gsl::narrow<int>(m_LatestNews.size()); ++i)
```
SyrianBallaS: #221 has a different (more efficient) way of doing this but just wanted to get rid of the warning.

src/RA_CodeNotes.cpp
```diff
@@ -35,9 +35,9 @@ size_t CodeNotes::Load(unsigned int nID)
         if (sNote.length() < 2U)
             continue;
 
-        const std::string& sAddr { note["Address"].GetString() };
-        const auto nAddr { static_cast<ra::ByteAddress>(std::stoul(sAddr, nullptr, 16)) };
-        const std::string& sAuthor { note["User"].GetString() }; // Author?
+        const std::string& sAddr{note["Address"].GetString()};
+        const ra::ByteAddress nAddr{std::stoul(sAddr, nullptr, 16)};
```
SyrianBallaS: If the left-sides type can hold the right sides value, brace-initialization is the way to go. The `string&`s from a C-String (possible invalidation, dangling) will have to change later but tried to keep the PR focused.

src/RA_Core.cpp
```diff
@@ -268,7 +268,7 @@ static unsigned int IdentifyRom(const BYTE* pROM, unsigned int nROMSize, std::st
     }
     else
     {
-        sCurrentROMMD5 = RAGenerateMD5(pROM, nROMSize);
+        sCurrentROMMD5 = RAGenerateMD5(gsl::make_span(pROM, nROMSize));
```
SyrianBallaS: I was trying to learn about spans.
It's like an array but doesn't decay. There will be std::span in the next C++, this is just a placeholder.

This lower params in functions that take pointer and size, instead we can just use span since size is always retained.

src/RA_Core.cpp
```diff
@@ -397,8 +397,9 @@ API void CCONV _RA_OnReset()
 
 API void CCONV _RA_InstallMemoryBank(int nBankID, void* pReader, void* pWriter, int nBankSize)
 {
-    g_MemManager.AddMemoryBank(static_cast<size_t>(nBankID), (_RAMByteReadFn*)pReader, (_RAMByteWriteFn*)pWriter, static_cast<size_t>(nBankSize));
-    g_MemoryDialog.AddBank(nBankID);
+    g_MemManager.AddMemoryBank(ra::to_unsigned(nBankID), (_RAMByteReadFn*)pReader, (_RAMByteWriteFn*)pWriter,
```
SyrianBallaS: Oddly enough, it didn't care about C-Style casts here but it did care about the `static_cast`s.

src/RA_Core.cpp
```diff
@@ -572,7 +573,7 @@ API HMENU CCONV _RA_CreatePopupMenu()
         AppendMenu(hRA, pConfiguration.IsFeatureEnabled(ra::services::Feature::Hardcore) ? MF_CHECKED : MF_UNCHECKED, IDM_RA_HARDCORE_MODE, TEXT("&Hardcore Mode"));
         AppendMenu(hRA, MF_SEPARATOR, 0U, nullptr);
 
-        AppendMenu(hRA, MF_POPUP, reinterpret_cast<UINT_PTR>(hRA_LB), TEXT("Leaderboards"));
+        GSL_SUPPRESS_TYPE1 AppendMenu(hRA, MF_POPUP, reinterpret_cast<UINT_PTR>(hRA_LB), TEXT("Leaderboards"));
```
SyrianBallaS: As far as I'm aware, this reinterpret_cast can't be avoided. Maybe in the future it can be but if it's low-level it can't be helped sometimes. So it's suppressed, there are many cases like this.

src/RA_Defs.h
```diff
@@ -23,20 +23,24 @@
...
-_CONSTANT_VAR MAX_BUFSIZ{ BUFSIZ*128U }; // Try not to use this one
-#endif	// RA_EXPORTS
+_CONSTANT_VAR BUFSIZ = 512U; // ANSI buffer size
+_CONSTANT_VAR STRING_BUFSIZ = BUFSIZ*4U; // For any string, the number is more if not a byte-string
+_CONSTANT_VAR MAX_BUFSIZ = BUFSIZ*128U; // Try not to use this one
+
+_CONSTANT_VAR WINHTTP_NO_PROXY_NAME = nullptr;
+_CONSTANT_VAR WINHTTP_NO_PROXY_BYPASS = nullptr;
+_CONSTANT_VAR WINHTTP_HEADER_NAME_BY_INDEX = nullptr;
+_CONSTANT_VAR WINHTTP_NO_HEADER_INDEX = nullptr;
```
SyrianBallaS: Redefined these as `nullptr` as compile-time constants.
Got rid of `WIDESTRING_BUFSIZ` since the one above it will multiply if not a byte string.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -398,8 +396,8 @@ BOOL CreateIPE(int nItem, CondSubItems nSubItem)
 
     HWND hList = GetDlgItem(g_AchievementEditorDialog.GetHWND(), IDC_RA_LBX_CONDITIONS);
 
-    RECT rcSubItem;
-    ListView_GetSubItemRect(hList, nItem, ra::etoi(nSubItem), LVIR_BOUNDS, &rcSubItem);
+    RECT rcSubItem{};
+    GSL_SUPPRESS_ES47 ListView_GetSubItemRect(hList, nItem, ra::etoi(nSubItem), LVIR_BOUNDS, &rcSubItem);
```
SyrianBallaS: This was the macro that needed the use `nullptr` rule suppressed. Our input is fine.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -714,481 +712,362 @@ INT_PTR Dlg_AchievementEditor::AchievementEditorProc(HWND hDlg, UINT uMsg, WPARA
...
-            // Ignore if we are currently editing a box!
-            if (g_hIPEEdit != nullptr)
-                break;
+        lpMinMaxInfo->ptMinTrackSize = pDlgAchEditorMin;
+    };
 
-            Achievement* pActiveAch = ActiveAchievement();
-            if (pActiveAch == nullptr)
+    // hwndCtl depends on id
+    const auto OnCommand = [this, &bHandled](HWND hwnd, int id, HWND hwndCtl, UINT codeNotify)
```
SyrianBallaS: I was able to use all the values supplied by the Windows API here and a few other places.

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -260,36 +260,30 @@ void Dlg_GameLibrary::ThreadedScanProc()
...
+            const auto nSize = ra::filesize(FilesToScan.front());
 
             auto pBuf = std::make_unique<unsigned char[]>(6 * 1024 * 1024);
 
-            fread(static_cast<void*>(pBuf.get()), sizeof(unsigned char), nSize, pf); // Check
-            Results.insert_or_assign(FilesToScan.front(), RAGenerateMD5(pBuf.get(), nSize));
-            pBuf.reset();
+            ifile.read(pBuf.get(), nSize); // Check
+            Results.insert_or_assign(FilesToScan.front(),
+                                     RAGenerateMD5(std::move(pBuf), gsl::narrow<std::size_t>(nSize)));
```
SyrianBallaS: The analyzer didn't like `unique_ptr` because it didn't use ownership semantics, I overloaded `RAGenerateMD5 `to give it ownership semantics.

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -532,6 +521,110 @@ INT_PTR CALLBACK Dlg_GameLibrary::s_GameLibraryProc(HWND hDlg, UINT uMsg, WPARAM
...
+                auto& pConfiguration = ra::services::ServiceLocator::GetMutable<ra::services::IConfiguration>();
+                pConfiguration.SetRomDirectory(sROMDirLocation);
+                RA_LOG("Selected Folder: %s\n", sROMDirLocation.c_str());
+                SetDlgItemText(hwnd, IDC_RA_ROMDIR, NativeStr(sROMDirLocation).c_str());
+                return FALSE;
+            }
+
+            case IDC_RA_LBX_GAMELIST:
+            {
+                const int nSel = ListView_GetSelectionMark(hwndCtl);
```
SyrianBallaS: `hwndCtl` is unboxed from an `LPARAM`. But it is the handle to the last control interacted with (usually the same `id` as the case label). In `fallthrough`  cases we couldn't assume that, so left how the handle was retrieved in those cases.

src/RA_Dlg_Memory.cpp
```diff
@@ -782,661 +783,662 @@ void Dlg_Memory::ClearLogOutput() noexcept
...
+                        if (nMatches > MIN_RESULTS_TO_DUMP)
+                            _stprintf_s(buffer, sizeof(buffer),
+                                _T("Found %u matches! (Displaying first %u results)"), nMatches,
+                                MIN_RESULTS_TO_DUMP);
+                        else if (nMatches == 0)
+                            _stprintf_s(buffer, sizeof(buffer), _T("Found *ZERO* matches!"));
+                        else
+                            _stprintf_s(buffer, sizeof(buffer), _T("Found %u matches!"), nMatches);
+                    }
+                    auto myRect{lpDrawItem->rcItem};
```
SyrianBallaS: The Windows API passes` lpDrawItem` as a pointer to const, so we needed to copy it. `RECT` is a literal type so it shouldn't be expensive.

src/RA_Dlg_Memory.cpp
```diff
@@ -782,661 +783,662 @@ void Dlg_Memory::ClearLogOutput() noexcept
...
-        case WM_GETMINMAXINFO:
-        {
-            LPMINMAXINFO lpmmi = (LPMINMAXINFO)lParam;
-            lpmmi->ptMaxTrackSize.x = pDlgMemoryMin.x;
-            lpmmi->ptMinTrackSize = pDlgMemoryMin;
-        }
-            return TRUE;
+    const auto OnGetMinMaxInfo = []([[maybe_unused]] HWND, MINMAXINFO* restrict lpMinMaxInfo)
+    {
+        Expects(lpMinMaxInfo != nullptr);
```
SyrianBallaS: I put all these preconditions to make sure it was working in all configurations.

src/RA_StringUtils.cpp
```diff
@@ -145,21 +145,21 @@ void StringBuilder::AppendToString(_Inout_ std::string& sResult) const
                 break;
 
             case PendingString::Type::StringRef:
-                nNeeded += pPending.Ref.String->length();
+                nNeeded += std::get<const std::string*>(pPending.Ref)->length();
```
SyrianBallaS: With `std::variant` you can get the desired type either by index or type via `std::get`. I decided to do it by type since it's clearer that way.

src/RA_StringUtils.h
```diff
@@ -174,7 +174,7 @@ class StringBuilder
     void Append(const std::string& arg)
     {
         PendingString pPending;
-        pPending.Ref.String = &arg;
+        pPending.Ref = &arg;
```
SyrianBallaS: As long as it's in the `std::variant` type list everything is fine, otherwise you'll get intellisense/compiler error.

src/RA_StringUtils.h
```diff
@@ -174,7 +174,7 @@ class StringBuilder
...
         pPending.DataType = PendingString::Type::StringRef;
         m_vPending.emplace_back(std::move(pPending));
```
SyrianBallaS: These `emplace_back`s should actually be `push_back `since the purpose of `emplace` is to construct in-place via perfect forwarding (no copying or moving). But ignored them for later.

src/RA_StringUtils.h
```diff
@@ -466,22 +464,7 @@ class StringBuilder
...
-                size_t Length;
-            } Char;
-
-            struct
-            {
-                const wchar_t* Pointer;
-                size_t Length;
-            } WChar;
-        } Ref{};
+        std::variant<const std::string*, const std::wstring*, gsl::cstring_span<>, gsl::cwstring_span<>> Ref;
```
SyrianBallaS: `cstring_span `and `cwstring_span` seemed like good candiates since they have pointer and size information.

src/RA_md5factory.cpp
```diff
@@ -2,65 +2,54 @@
...
+    md5_finish(&pms, digest.data());
 
-    memset(buffer, 0, MD5_STRING_LEN + 1);
-    sprintf_s(buffer, MD5_STRING_LEN + 1,
-        "%02x%02x%02x%02x%02x%02x%02x%02x"
-        "%02x%02x%02x%02x%02x%02x%02x%02x",
-        digest[0], digest[1], digest[2], digest[3], digest[4], digest[5], digest[6], digest[7],
-        digest[8], digest[9], digest[10], digest[11], digest[12], digest[13], digest[14], digest[15]);
+    std::array<char, MD5_STRING_LEN + 1> buffer{};
+    sprintf_s(buffer.data(), MD5_STRING_LEN + 1,
```
SyrianBallaS: Tried to make these safer/efficient. It seems the the md5 returned has to be on the stack.

The mem* functions shouldn't be used at all, unless we have no choice which I don't know why we wouldn't since C++ is based around RAII.

src/data/SessionTracker.cpp
```diff
@@ -62,7 +62,10 @@ void SessionTracker::LoadSessions()
 
             const auto nSessionLength = strtoul(&sLine[nIndex2], nullptr, 10);
 
-            auto md5 = RAGenerateMD5(reinterpret_cast<const unsigned char*>(sLine.c_str()), nIndex + 1);
+            
+#pragma warning(suppress: 26490)
+            GSL_SUPPRESS_TYPE1 auto md5 =
+                RAGenerateMD5(gsl::make_span(reinterpret_cast<const unsigned char*>(sLine.c_str()), nIndex + 1));
```
SyrianBallaS: There is a function in gsl called `as_bytes`, but that's for spans of `const std::byte*`. Plus it had to do `reinterpret_cast` anyway so it didn't seem worth the hassle.

src/data/SessionTracker.hh
```diff
@@ -70,7 +70,7 @@ private:
 
     std::vector<GameStats> m_vGameStats;
 
-    long m_nFileWritePosition{};
+    std::streampos m_nFileWritePosition{};
```
SyrianBallaS: Made this `std::streampos` to reduce the need of casting, it's implicitly convertible `std::streamoff` (`long long`)

src/services/impl/LeaderboardManager.cpp
```diff
@@ -23,15 +23,23 @@ LeaderboardManager::LeaderboardManager(const ra::services::IConfiguration& pConf
 {
 }
 
-RA_Leaderboard* LeaderboardManager::FindLB(ra::LeaderboardID nID)
+RA_Leaderboard* LeaderboardManager::FindLB(ra::LeaderboardID nID) noexcept
```
SyrianBallaS: These are guaranteed not to throw anymore because of utilizing ranged-for.

src/ui/win32/DialogBase.cpp
```diff
@@ -93,71 +94,78 @@ _Use_decl_annotations_ void DialogBase::CreateModalWindow(const TCHAR* restrict
...
 INT_PTR CALLBACK DialogBase::DialogProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
 {
+    // Message handlers made to reduce explicit casting only
+    const auto _OnCommand = [this]([[maybe_unused]] HWND /*hwnd*/, int id, HWND hwndCtl, UINT codeNotify)
+    {
+        switch (codeNotify)
+        {
+            case EN_KILLFOCUS:
+            {
+                auto* pControlBinding = FindControlBinding(hwndCtl);
```
SyrianBallaS: Gave it an intermediate handler to not have to cast ourselves, the member function is unchanged.

src/ui/win32/DialogBase.cpp
```diff
@@ -93,71 +94,78 @@ _Use_decl_annotations_ void DialogBase::CreateModalWindow(const TCHAR* restrict
...
+                if (pControlBinding)
+                    pControlBinding->LostFocus();
+
+                pControlBinding = FindControlBinding(hwndCtl);
+                if (pControlBinding)
+                    pControlBinding->OnCommand();
+                break;
+            }
+        }
+        return this->OnCommand(gsl::narrow<WORD>(id));
```
SyrianBallaS: It doesn't need to be `LOWORD`'d because they are the same.

tests/mocks/MockServer.hh
```diff
@@ -28,11 +29,13 @@ public:
     template<typename TApi>
     void HandleRequest(std::function<bool(const typename TApi::Request&, typename TApi::Response&)>&& fHandler)
     {
-        m_mHandlers.insert_or_assign(std::string(TApi::Name()), [fHandler = std::move(fHandler)](
-                                                                    void* restrict pRequest, void* restrict pResponse) {
-            const gsl::not_null<const typename TApi::Request* const> pTRequest{
+        m_mHandlers.insert_or_assign(
+            std::string(TApi::Name()),
+            [fHandler = std::move(fHandler)](const void* restrict pRequest, void* restrict pResponse)
```
SyrianBallaS: pRequest needs to be const void* to not use const_cast as well as in the map. It makes sense since all requests are passed by const reference and responses by reference.
References and pointers have the same exact semantics just different syntax.

SyrianBallaS:
I need to test this rigorously again after merging the new stuff, closing but will reopen later.

src/RA_Core.cpp
Jamiras: Will `gsl::make_span` work on unix systems? I'm concerned all this gsl stuff you're adding is actually making it harder to ever have a linux version of the dll. That was one of the primary motivations for switching to an MVVM model.

SyrianBallaS:
I'm gonna break it up as one warning code at a time but not add the rule set until the last one. Gsl works on linux. The wndproc will stay mostly the same, but will have the reinterpret cast rule disabled. Closing this.

src/RA_Core.cpp
SyrianBallaS: This is taken direclty from Microsoft's GSL repo readme, so yes it works on linux. They've been trying to be cross-platform ever since .NET Core was a thing in 2015.
image

Issue #236, opened 2019-01-06
Checksum sometimes remains uninitialized while a valid title is loaded


rzumer:
It seems like the current (master) version of `RAIntegration` will leave the checksum uninitialized when closing via without a reset following `RA_OnLoadNewRom(NULL, 0)` and subsequently loading a valid title without resetting.

I reproduced this via `RAQUASI88`, by loading an unlinked file and swapping to a linked one via the `Disk` menu. The relevant functions are called in the following sequence:
1. Load an unlinked file
  * `RA_IdentifyRom()` returns 0 and `RA_ActivateGame(0)` is called
2. Load a linked file
  * `RA_OnLoadNewRom(NULL, 0)` is called as the unlinked file is unloaded
  * `RA_IdentifyRom()` returns a value `n` above 0 and `RA_ActivateGame(n)` is called

Without a call to `OnReset()`, the checksum remains unpopulated and the Report function is unusable, while the rest of the menu works as expected.

Jamiras:
The problem is that the actual order of calls is:
* `RA_IdentifyRom()` returns n
* `RA_OnLoadNewRom(NULL, 0)`
* `RA_ActivateGame(n)`

The problem is the hash is stored when it's calculated in `RA_IdentifyROM`, and cleared out by `RA_OnLoadNewRom(NULL, 0)`. `RA_ActivateGame` has no knowledge of the hash and cannot regenerate it.

rzumer:
Good catch, I can confirm that removing the `RA_OnLoadNewRom()` call while loading allows the checksum to be filled correctly. However, the Report function is still disabled.

PR #237, opened 2019-01-07
don't set hash until game activated

Jamiras:
Improvement to #228 - stores the hash calculated by `RA_IdentifyRom`, but doesn't update the context until `RA_ActivateGame` is called. Also addresses situation described in #236.


rzumer:
Doesn't that actually break disk swapping?

Jamiras:
`RA_IdentifyRom` still sets the hash if the game ID doesn't change: https://github.com/RetroAchievements/RAIntegration/pull/237/files#diff-135d78adae4e06a4737c3a4090a9273dR374

rzumer:
OK, seems good.

Issue #238, opened 2019-01-08
trigger achievements only when the condition set becomes true after being false


meleu:
This feature request is a result of the discussion that happened here: RetroAchievements/RALibretro#52

The main goal here is to increase compatibility between all RA client emulators.

To avoid false triggering due to non-standardized initialization of the RAM (different behavior among various emulators), I suggest adding a logic responsible to trigger the achievements only when they become true after being false.

As pointed by @leiradel this is the RetroArch's behavior.
Reference: https://github.com/libretro/RetroArch/blob/c2047c8c04818715cf606249db630a3836619aa2/cheevos/cheevos.c#L1630

And according to him, `rc_reset_trigger` can be used to accomplish this.

rzumer:
This would not solve every potential issue caused by making wrong assumptions on the contents of uninitialized RAM, but it seems like a good change to reduce the likelihood of such errors triggering unlocks on load. Either way, the implementations should match.

Jamiras:
It's actually the `last` flag (line 1634) that provides this behavior in RetroArch. 

I believe this is a valid alternative to the "100 frames" solution and will be implementing it as such. If an achievement is active on game load, it will be ignored until it becomes inactive, at which point it will be allowed to trigger in the future. If any achievements would trigger in the first frame, they will be ignored, and all other achievements will become activate for the second frame. This effectively sets the "100 frames" delay to "1 frame" for most achievements and "as many frames as necessary" for any achievements that are triggered by the uninitialized memory.

Similarly, on reset, all active achievements will have their HitCounts reset and be temporarily deactivated for "1 frame" or "as many frames as necessary" to prevent triggering during the reset process.

It also provides consistency between the implementations.

PR #240, opened 2019-01-10
replace 100 frame delay with delayed activation logic (addresses #238)

Jamiras:


kdecks:
this is great

rzumer:
I'm not sure I understood your description of the behavior in #238, so would it not be a problem to remove the delay if memory is "initialized" twice? I have seen cases where the memory layout changes completely in multiple phases.

Jamiras:
Memory can be initialized as many times as the emulator feels is necessary and whenever it feels like doing it. We can't control that. This code implements a solution that matches what RetroArch is doing on load to prevent triggers due to uninitialized memory. Upon load, all achievements are ignored until they evaluate false, at which point, they become eligible to trigger on the next frame. If re-initializing the memory causes an achievement to evaluate false, it will become eligible. If it has previously evaluated false and re-initialization causes it to evaluate true, it will trigger. 

By requiring the trigger to be false, we can prevent "on load" triggers from firing for as long as necessary. We don't have to guess if 10 frames is too little or 100 frames is too much.

There is similar behavior in the leaderboard code. The leaderboard cannot be reactivated after a cancel until the submit condition is false for at least one frame. This prevents reactivation in the middle of a challenge if the player fails it.

rzumer:
Seems good to me, it is just worth keeping in mind that this might expose issues already present in RetroArch that were not visible in the standalone emulators and RALibretro.

PR #245, opened 2019-01-12
CppCoreCheck Type Profile (es.47)

SyrianBallaS:
Basically this just replaces `0` with `nullptr`, the rule was suppressed where it couldn't be helped.


src/RA_httpthread.cpp
```diff
@@ -245,8 +245,8 @@ BOOL DoBlockingImageUpload(UploadType nType, const std::string& sFilename, std::
     // Use WinHttpOpen to obtain a session handle.
     HINTERNET hSession = WinHttpOpen(RAWeb::GetUserAgent().c_str(),
         WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
-        WINHTTP_NO_PROXY_NAME,
-        WINHTTP_NO_PROXY_BYPASS, 0);
+        nullptr,
+        nullptr, 0);
```
Jamiras: Why suppress for the LVIR constants, but not for the WINHTTP constants? The [documentation](https://docs.microsoft.com/en-us/windows/desktop/api/winhttp/nf-winhttp-winhttpopen) specifically says the constants must be used:
> If dwAccessType is not set to WINHTTP_ACCESS_TYPE_NAMED_PROXY, this parameter must be set to WINHTTP_NO_PROXY_NAME.

src/services/impl/WindowsHttpRequester.cpp
```diff
@@ -17,7 +17,8 @@ static bool ReadIntoString(const HINTERNET hRequest, std::string& sBuffer, DWORD
 {
     DWORD dwSize = sizeof(DWORD);
     DWORD nContentLength = 0;
-    if (!WinHttpQueryHeaders(hRequest, WINHTTP_QUERY_CONTENT_LENGTH | WINHTTP_QUERY_FLAG_NUMBER, WINHTTP_HEADER_NAME_BY_INDEX, &nContentLength, &dwSize, WINHTTP_NO_HEADER_INDEX))
+    if (!WinHttpQueryHeaders(hRequest, WINHTTP_QUERY_CONTENT_LENGTH | WINHTTP_QUERY_FLAG_NUMBER, nullptr,
+                             &nContentLength, &dwSize, nullptr))
```
Jamiras: > If the flag in dwInfoLevel is not WINHTTP_QUERY_CUSTOM, set this parameter to WINHTTP_HEADER_NAME_BY_INDEX.

src/RA_httpthread.cpp
Jamiras: Same comment applies several times below.

SyrianBallaS: `LVIR_BOUNDS` parameter is an `int` but the macro is defined like this:
```
#define ListView_GetSubItemRect(hwnd, iItem, iSubItem, code, prc) \
        (BOOL)SNDMSG((hwnd), LVM_GETSUBITEMRECT, (WPARAM)(int)(iItem), \
                ((prc) ? ((((LPRECT)(prc))->top = (iSubItem)), (((LPRECT)(prc))->left = (code)), (LPARAM)(prc)) : (LPARAM)(LPRECT)NULL))
```

I changed the `WINHTTP` constants (even though they are pointers) as requested anyway.

SyrianBallaS: If it were C, `NULL` would be `(void*)0` which is the same as `nullptr` in C++. You can't implicitly cast `void*` to pointer to object so it's `0` in C++, that's why there's `nullptr`.

src/services/impl/WindowsHttpRequester.cpp
SyrianBallaS: I put all the macros back, but had to suppress by code sometimes.

PR #249, opened 2019-01-13
CppCoreCheck Type Profile (type.4)

SyrianBallaS:
Replace c-style casts either with an API that casts for us, static_cast (via function) or reinterpret_cast. `WNDPROC`s have not been changed in structure. Doing `type.1a` depends on this one since we need all the `c-casts` to be `named casts` to suppress `type.1a` since using `Message Crackers` aren't allowed (the callback functions for each message). These message crackers can be found in `WindowsX.h`.

Message Crackers are extremely useful but w/e.


src/RA_Dlg_AchEditor.cpp
```diff
@@ -1079,7 +1078,7 @@ INT_PTR Dlg_AchievementEditor::AchievementEditorProc(HWND hDlg, UINT uMsg, WPARA
                                 m_ConditionClipboard.Add(NewCondition);
                             }
 
-                            (void)m_ConditionClipboard.Count();
+                            static_cast<void>(m_ConditionClipboard.Count());
```
Jamiras: The `_NODISCARD` is doing it's job here. `Count()` returns a member variable, there's no reason it needs to be called. Suggest just removing it.

Jamiras:
I'm not totally against the message crackers, I just ask that you do that as a separate PR. When small changes like those present in this PR are hidden in massive refactoring, it's hard to judge them. 

Additionally, knowing that the long-term plan is to convert to an MVVM structure, the value of converting now is diminished.

src/RA_Dlg_AchEditor.cpp
SyrianBallaS: done.

PR #251, opened 2019-01-13
CppCoreCheck Type Profile (type.7)

SyrianBallaS:
Replace the `union` in `PendingString` with a `std::variant`. The speed is unaffected (going by the tests and running RANes). Pointers cannot be moved as they are invalidated immediately. I believe everything's fine but if any of you need any changes from it, I'm all ears. 

To make this easier, should I just disable the warning code that complains `reinterpret_cast` and be done with it? With all things considered, it's impossible to address it _**at all**_ with the options available to us at this time. We could do it later though. It is however, still impossible to address it completely as we are using relatively low level APIs.


src/RA_StringUtils.h
```diff
@@ -496,24 +461,31 @@ class StringBuilder
...
-            } WChar;
-        } Ref{};
+        // These are needed by m_vPending's allocator
+        PendingString(const PendingString&) = default;
+        PendingString(PendingString&&) noexcept = default;
+        PendingString() noexcept = delete;
+
+    private:
+        using RefType = std::variant<const std::string, const std::wstring, std::string_view, std::wstring_view>;
+        RefType Ref;
```
Jamiras: I believe this creates copies of `std::string` and `std::wstring`. `RefType` should probably just be a variant of `string_view` and `wstring_view`, where they can be populated with the `c_str` and `length` of the source strings. Or the `std::string` and `std::wstring` variant parts should be references.

src/RA_StringUtils.h
```diff
@@ -528,6 +500,8 @@ class StringBuilder
             WCharRef,
         };
         Type DataType{Type::String};
+
+        friend class StringBuilder; // No other class needs access to PendingString's data members
```
Jamiras: `PendingString` is a private subclass, so it doesn't really matter.

src/RA_StringUtils.h
SyrianBallaS: I made it so string_view would do it for only pointers.
Originally I made the views for `StringRef `but it kept on being called incorrectly when they were supposed to be for `String` (`string_views `are non-owning references of strings so when called incorrectly it printed out a dangled string). After corrected, the `string_view` constructors weren't called at all, so I just removed them.

src/RA_StringUtils.h
```diff
@@ -177,52 +177,33 @@ class StringBuilder
     template<typename T>
     void Append(const T& arg)
     {
-        PendingString pPending;
         if (m_bPrepareWide)
-        {
-            pPending.WString = ra::ToWString(arg);
-            pPending.DataType = PendingString::Type::WString;
-        }
+            m_vPending.emplace_back(std::wstring{ra::ToWString(arg)});
```
SyrianBallaS: This might seem unnecessary but because the `ToWString` and `ToString` methods are const, it has to be converted to `rvalue` reference (a non-owning temporary).

src/RA_StringUtils.h
```diff
@@ -496,24 +461,33 @@ class StringBuilder
...
+            Ref{std::wstring_view{ptr, len}},
+            DataType{Type::WCharRef}
+        {}
+        PendingString() noexcept = delete;
+        ~PendingString() noexcept = default;
+
+        // To prevent copying moving
+        PendingString(const PendingString&) = delete;
+        PendingString& operator=(const PendingString&) = delete;
+        PendingString& operator=(PendingString&&) noexcept = delete;
```
SyrianBallaS: No copying is ensured now. Move construction is still needed for perfect forwarding.

Issue #115, opened 2018-09-21
Freely deletable notes? Seriously?


ubergeek77:
I wanted to get started on developing achievements for a game, and there was some clutter I wanted to get rid of in the Memory Inspector notes. What _**wasn't made perfectly clear**_ is that, when a note is deleted, _**it deletes that note for ALL developers**_ as it syncs this change to the server immediately. I created a new RetroAchievements account to test this, and sure enough, those notes are gone. It is eluded to in the documentation, but it is glazed over, and certainly isn't made as urgently clear as it should be.

One, single button, makes this change. All I have to do it click the "delete" button, and BOOM, note gone.

There are _**NO**_ warnings. _**Zero**_ indication that I may be screwing over other developers by doing this. Yes, I could save my own bookmarks, but the current UX doesn't make this immediately clear or easy to do. Using notes as though they are bookmarks is much more convenient. I can still see a JSON with usernames and addresses in XYZ-Notes2.txt, but the text content of the "notes" field is empty in all of them.

I realize this is an unpaid, open-source hobby project, but quite frankly, this is unacceptable.

At the absolute minimum, users should be shown a warning that makes it very clear that deleting and saving notes will do the same for other developers. But this won't stop trolls from going through a bunch of games and deleting notes, or even **creating fake notes to waste people's time.** This is the internet. Stuff like this will absolutely happen.

As for long term fixes, the default server behavior shouldn't be "overwrite" as it currently is. There are many ways to go about this, but one of them is to do the following:

- _Always_ keep a local copy of locally-created notes. These will override any notes downloaded from the server. This also prevents notes from being lost in the event they are deleted from the RA server.
- On the server-side, save notes on a **per-account** basis, not a per-game basis. This handles accidental deletion of notes for _literally everyone creating achievements._

In this setup, I believe the default behavior should be for RAIntegration to NOT download notes by default. Instead, a "download online notes" button should be implemented, or something to that effect. Then, when this button is clicked:

1. Query the server to retrieve a list of **users** that have notes for a given game.
2. Download separate lists of notes for those users.
3. Combine the two lists, using the address as the unique key
4. When a duplicate is found, ask the user which of the two they would like to keep.
5. Combine this new, combined list with the local list of notes, if any. Then, either use the local list of notes to override any found in this new list (as suggested above), or ask the user once again which of the two they want to use.

I'm sure this could be fleshed out more. But there really needs to be a discussion about this.

[...]

PR #252, opened 2019-01-13
add confirmation dialog to delete note (supports #115)

Jamiras:
Changes the overwrite prompt from
image
to
image
and adds a prompt for deleting the note
image


SyrianBallaS:
I think this could be simplified with class invariants for the MessageBoxViewModel but would be outside the scope of your PR, other than that I think it's good.

Jamiras:
Not sure what you mean by class invariants, but I couldn't use the existing ShowWarningMessage helper functions because they use OkCancel instead of YesNo.

SyrianBallaS:
Like you could use constructors with default arguments. Don't worry about it, it's not that a big a deal. What you have is fine.

PR #254, opened 2019-01-13
Ensure offscreen rendering is clipped to surface

Jamiras:
Prevents "wraparound" when writing directly to the bitmap memory:

Before:
image

After:
image

The text still writes outside the orange area because it's clipped to the entire surface, not the orange area. That's a possible future improvement.


SyrianBallaS:
Seems good to me.

Issue #253, opened 2019-01-13
Emulator crash after logging in, when launched outside of the executable root directory


rzumer:
Similar to #79, which was fixed in 0.73, `RANes` will simply crash after logging in. I assume that it will also be the case with other emulators. Log below with the user information redacted (the `BaseDirectory` is correct).

```
132909.000|INFO| ================================================================================
132909.000|INFO| Log started at 01/13/19 13:29:09 Eastern Standard Time
132909.000|INFO| BaseDirectory: C:\Users\rzumer\Applications\RANes\
132910.151|INFO| Loading preferences...
132910.181|INFO| Initializing 15 worker threads
132910.182|ERR | Error 3 getting file size: RACache\Badge\00000.png
132910.182|INFO| Downloading http://i.retroachievements.org/Badge/00000.png
132910.182|ERR | Error 3 getting file size: RACache\UserPic\_User.png
132910.182|INFO| Downloading retroachievements.org/UserPic/_User.png
132910.182|INFO| Failed to create "RACache\Badge\00000.png": 2
132910.182|INFO| Failed to create "RACache\UserPic\_User.png": 2
132910.183|INFO| User-Agent: RANes/0.015.0.0 (WindowsNT 10.0) Integration/0.74.75.0
132910.183|INFO| POST to dorequest.php?r=news&c=6
132910.183|INFO| POST to dorequest.php?r=latestclient&e=5
132910.185|INFO| Login Request: ...
132910.678|INFO| -- Login Response: {"Success":true,...}
132910.678|INFO| Received 40 bytes from dorequest.php?r=latestclient: {"Success":true,"LatestVersion":"0.015"}
```

PR #255, opened 2019-01-13
Ensure paths are absolute when interacting with file system

Jamiras:
Addresses #253. The RACache directory was created with a full path, but the images were being written with a relative path. Since the relative RACache directory doesn't exist, an exception occurred.

### DIFF
--- a/src/services/impl/FileLogger.hh
+++ b/src/services/impl/FileLogger.hh
@@ -29,6 +29,12 @@ public:
             pFileSystem.DeleteFile(sOldLogFilePath);
             pFileSystem.MoveFile(sLogFilePath, sOldLogFilePath);
         }
+        else if (nLogSize < 0)
+        {
+            std::wstring sCacheDirectory = pFileSystem.BaseDirectory() + L"RACache";
+            if (!pFileSystem.DirectoryExists(sCacheDirectory))
+                pFileSystem.CreateDirectory(sCacheDirectory);
+        }
 
         m_pWriter = pFileSystem.AppendTextFile(sLogFilePath);
         if (m_pWriter != nullptr)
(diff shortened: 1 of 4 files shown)


src/services/impl/WindowsFileSystem.cpp
```diff
@@ -60,18 +77,26 @@ size_t WindowsFileSystem::GetFilesInDirectory(const std::wstring& sDirectory, _I
 
 bool WindowsFileSystem::DeleteFile(const std::wstring& sPath) const noexcept
 {
-    return static_cast<bool>(DeleteFileW(sPath.c_str()));
+    std::wstring sBuffer;
+    const auto& sAbsolutePath = MakeAbsolute(sBuffer, sPath);
+    return static_cast<bool>(DeleteFileW(sAbsolutePath.c_str()));
```
SyrianBallaS: I'm not totally sure if the type.1b PR will overwrite this but you should probably just use `!= 0` (0 is failure for these File functions) instead of casting.

If it can be done by merging, don't worry about it. There doesn't seem to be anything else that needs improving here.

Jamiras: Since the same line is modified in two different ways, it'll cause a merge conflict, which will have to be resolved manually.

src/services/impl/FileLogger.hh
```diff
@@ -29,6 +29,12 @@ public:
             pFileSystem.DeleteFile(sOldLogFilePath);
             pFileSystem.MoveFile(sLogFilePath, sOldLogFilePath);
         }
+        else if (nLogSize < 0)
+        {
+            std::wstring sCacheDirectory = pFileSystem.BaseDirectory() + L"RACache";
```
SyrianBallaS: I spoke too soon. The `+` should be replaced with a `StringPrintf` to only work on one string instead of needing two to make one.

SyrianBallaS:
This should be done in a different PR but thought I should comment on it.
In the spirit of trying to be cross-platform, have you considered using the standard's filesystem API?
https://en.cppreference.com/w/cpp/filesystem

It's unix based but works on windows.

PR #257, opened 2019-01-17
Retry achievement unlock on network error

Jamiras:
* If the WinHttp library returns one of several identified error codes (three of which I've experienced), the API call with fail with an Incomplete state.
* If the Incomplete state is detected on login, a second login attempt will be made immediately. If that also fails, the error will be reported to the user as the login process is synchronous.
* If the Incomplete state is detected when unlocking an achievement, the request will be made again until it is successful or the application is closed. The first retry will occur 500ms after the failed attempt, then the delay will double up to 2 minutes between attempts: 1s, 2s, 4s, 8s, 16s, 32s, 64s, 120s, 120s, 120s...
* Changes made in #242 ensure the popup is displayed even if the server doesn't acknowledge the unlock.
* In the unlikely event that connectivity is not restored before the application is closed, the unlock will be lost. 
  - This is no different than the current application, which silently swallows the network error.
    - Because the popup does not currently display in this situation, the player may not realize they triggered the achievement and may submit a ticket indicating the achievement did not work. It is impossible for the developer to know the problem was network related without looking at the player's local log file.
  - Because the popup does display now, the player will see that they triggered the achievement, but may not realize the server did not acknowledge it. The assumption is that the client will eventually reach the server and the achievement will be unlocked. 
    - Because the popup does display now, the player may take a screenshot to use as proof (although, doing that for every achievement just in case their network goes out seems inefficient).

### DIFF
--- a/src/api/ApiCall.hh
+++ b/src/api/ApiCall.hh
@@ -42,7 +42,7 @@ protected:
 
 private:
     template<class TRequest, class TCallback>
-    static void DoAsyncWithRetry(const TRequest&& request, TCallback&& callback, std::chrono::milliseconds delay)
+    static void DoAsyncWithRetry(const TRequest& request, TCallback&& callback, std::chrono::milliseconds delay)
     {
         auto response = request.Call();
         if (response.Result != ApiResult::Incomplete)
@@ -97,7 +97,7 @@ struct ApiResponseBase
     /// <returns><c>true</c> if the call was unsuccessful, <c>false</c> if not.</returns>
     bool Failed() const noexcept
     {
-        return (!Succeeded() && Result != ApiResult::None);
+        return !Succeeded() && (Result != ApiResult::None);
     }
 };
 
--- a/src/api/ApiCall.hh
+++ b/src/api/ApiCall.hh
@@ -11,21 +11,62 @@ namespace ra {
 namespace api {
 
 enum class ApiResult
-{    
+{
     None = 0,       // unspecified
     Success,        // call was successful
     Error,          // an error occurred
     Failed,         // call was unsuccessful, but didn't error
     Unsupported,    // call is not supported and was not made
+    Incomplete,     // call could not be made at this time
 };
 
 struct ApiRequestBase
 {
+protected:
     template<class TRequest, class TCallback>
     static void CallAsync(const TRequest& request, TCallback&& callback)
     {
         ra::services::ServiceLocator::GetMutable<ra::services::IThreadPool>().RunAsync(
-            [request, callback = std::move(callback)] { callback(request.Call()); });
+            [request, callback = std::move(callback)]{ callback(request.Call()); });
+    }
+
+    template<class TRequest, class TCallback>
+    static void CallAsyncWithRetry(const TRequest& request, TCallback&& callback)
+    {
+        ra::services::ServiceLocator::GetMutable<ra::services::IThreadPool>().RunAsync(
+            [request, callback = std::move(callback)]
+        {
+            DoAsyncWithRetry(std::move(request), std::move(callback), std::chrono::milliseconds(0));
+        });
+    }
+
+private:
+    template<class TRequest, class TCallback>
+    static void DoAsyncWithRetry(const TRequest&& request, TCallback&& callback, std::chrono::milliseconds delay)
+    {
+        auto response = request.Call();
+        if (response.Result != ApiResult::Incomplete)
+        {
+            callback(response);
+            return;
+        }
+
+        if (delay < std::chrono::milliseconds(500))
+        {
+            delay = std::chrono::milliseconds(500);
+        }
+        else
+        {
+            delay += delay;
+            if (delay > std::chrono::minutes(2))
+                delay = std::chrono::minutes(2);
+        }
+
+        ra::services::ServiceLocator::GetMutable<ra::services::IThreadPool>().ScheduleAsync(delay,
+            [request = std::move(request), callback = std::move(callback), delay]
+        {
+            DoAsyncWithRetry(std::move(request), std::move(callback), delay);
+        });
     }
 };
 
@@ -56,7 +97,7 @@ struct ApiResponseBase
     /// <returns><c>true</c> if the call was unsuccessful, <c>false</c> if not.</returns>
     bool Failed() const noexcept
     {
-        return (Result == ApiResult::Error || Result == ApiResult::Unsupported || Result == ApiResult::Failed);
+        return (!Succeeded() && Result != ApiResult::None);
     }
 };
 
--- a/src/services/impl/WindowsHttpRequester.cpp
+++ b/src/services/impl/WindowsHttpRequester.cpp
@@ -228,6 +228,45 @@ unsigned int WindowsHttpRequester::Request(const Http::Request& pRequest, TextWr
     return nStatusCode;
 }
 
+// these defines are in <wininet.h>
+// cannot include <wininet.h> and <winhttp.h> in the same file
+#define INTERNET_ERROR_BASE                     12000
+#define ERROR_INTERNET_TIMEOUT                  (INTERNET_ERROR_BASE + 2)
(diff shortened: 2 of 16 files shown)


src/api/ApiCall.hh
```diff
@@ -11,21 +11,62 @@ namespace ra {
...
+            [request, callback = std::move(callback)]{ callback(request.Call()); });
+    }
+
+    template<class TRequest, class TCallback>
+    static void CallAsyncWithRetry(const TRequest& request, TCallback&& callback)
+    {
+        ra::services::ServiceLocator::GetMutable<ra::services::IThreadPool>().RunAsync(
+            [request, callback = std::move(callback)]
+        {
+            DoAsyncWithRetry(std::move(request), std::move(callback), std::chrono::milliseconds(0));
```
SyrianBallaS: If you want to make this easier/more intuitive, you can try using the `chrono_literals`.
They have them for all duration helper types. They are all written the same as you would normally except microseconds uses the "us" suffix since the actual character for it is Unicode.

Not a big deal, just something to consider.
https://en.cppreference.com/w/cpp/chrono/duration

src/api/ApiCall.hh
```diff
@@ -11,21 +11,62 @@ namespace ra {
...
+        });
+    }
+
+private:
+    template<class TRequest, class TCallback>
+    static void DoAsyncWithRetry(const TRequest&& request, TCallback&& callback, std::chrono::milliseconds delay)
```
SyrianBallaS: Does request really have to be const like that? I'm going to suggest to get rid of the `const`, or remove one of the `&`. A const temporary doesn't make any sense.

src/api/ApiCall.hh
```diff
@@ -11,21 +11,62 @@ namespace ra {
...
+
+        if (delay < std::chrono::milliseconds(500))
+        {
+            delay = std::chrono::milliseconds(500);
+        }
+        else
+        {
+            delay += delay;
+            if (delay > std::chrono::minutes(2))
+                delay = std::chrono::minutes(2);
```
SyrianBallaS: These look like good opportunities to use `std::chrono_literals`.

src/api/ApiCall.hh
```diff
@@ -56,7 +97,7 @@ struct ApiResponseBase
     /// <returns><c>true</c> if the call was unsuccessful, <c>false</c> if not.</returns>
     bool Failed() const noexcept
     {
-        return (Result == ApiResult::Error || Result == ApiResult::Unsupported || Result == ApiResult::Failed);
+        return (!Succeeded() && Result != ApiResult::None);
```
SyrianBallaS: Up to you but I think you should make the order of evaluation clearer by using parentheses.
Kind of like this: `!(Succeeded() && (Result == ApiResult::None))`
or: `(!Succeeded() && (Result != ApiResult::None))`

src/services/impl/WindowsHttpRequester.cpp
```diff
@@ -228,6 +228,45 @@ unsigned int WindowsHttpRequester::Request(const Http::Request& pRequest, TextWr
...
+#define ERROR_INTERNET_NAME_NOT_RESOLVED        (INTERNET_ERROR_BASE + 7)
+#define ERROR_INTERNET_OPERATION_CANCELLED      (INTERNET_ERROR_BASE + 17)
+#define ERROR_INTERNET_INCORRECT_HANDLE_STATE   (INTERNET_ERROR_BASE + 19)
+#define ERROR_INTERNET_ITEM_NOT_FOUND           (INTERNET_ERROR_BASE + 28)
+#define ERROR_INTERNET_CANNOT_CONNECT           (INTERNET_ERROR_BASE + 29)
+#define ERROR_INTERNET_CONNECTION_ABORTED       (INTERNET_ERROR_BASE + 30)
+#define ERROR_INTERNET_CONNECTION_RESET         (INTERNET_ERROR_BASE + 31)
+#define ERROR_INTERNET_FORCE_RETRY              (INTERNET_ERROR_BASE + 32)
+#define ERROR_HTTP_INVALID_SERVER_RESPONSE      (INTERNET_ERROR_BASE + 152)
+#define ERROR_INTERNET_DISCONNECTED             (INTERNET_ERROR_BASE + 163)
```
SyrianBallaS: Since you can't include `<wininet.h>`, just make them `constexpr`.

tests/RA_UnitTestHelpers.h
```diff
@@ -26,6 +26,12 @@ namespace CppUnitTestFramework {
 
 #pragma warning(push)
 #pragma warning(disable : 4505) // unreferenced inline functions, they are referenced. Must be a bug.
+template<>
+std::wstring ToString<std::chrono::milliseconds>(const std::chrono::milliseconds& t)
+{
+    return std::to_wstring(t.count()) + L"ms";
```
SyrianBallaS: The template arguments don't have to be there for specializations.
You could write it like:
```
template<>
std::wstring ToString(const std::chrono::milliseconds& t)
```

I also think you should use one of the methods in `StringUtils` instead of using `+`

src/services/impl/WindowsHttpRequester.hh
```diff
@@ -19,6 +19,8 @@ public:
 
     unsigned int Request(const Http::Request& pRequest, TextWriter& pContentWriter) const override;
 
+    bool IsRetryable(unsigned int nStatusCode) const noexcept override;
```
SyrianBallaS: Do you want the return values to be ignored? If you do, don't worry about it but I don't think they should.

tests/mocks/MockHttpRequester.hh
```diff
@@ -26,6 +26,11 @@ public:
         return ra::etoi(response.StatusCode());
     }
 
+    bool IsRetryable(_UNUSED unsigned int nStatusCode) const noexcept
```
SyrianBallaS: I found out that not putting a parameter name doesn't trigger the unused formal parameter warning. I'm guessing that's why they called it `maybe_unused`. Parameter name could be commented out the have the same effect.

tests/mocks/MockThreadPool.hh
```diff
@@ -52,6 +52,20 @@ public:
...
+    /// </summary>
+    std::chrono::milliseconds NextTaskDelay() const
+    {
+        if (!m_vTasks.empty())
+            return std::chrono::milliseconds(0);
+
+        if (!m_vDelayedTasks.empty())
+            return m_vDelayedTasks.front().nDelay;
+
+        return std::chrono::hours(10000);
```
SyrianBallaS: Do duration types do implicit casting?

Jamiras: Yes. If this code is reached, it will generate an error like this:
```
Message: Assert failed. Expected:<500ms> Actual:<36000000000ms>
```

src/api/ApiCall.hh
Jamiras: I prefer not to use suffixes for non-primitives. It's very clear that this is comparing against 2 minutes, whereas "2m" could be potentially be meters or something else.

tests/mocks/MockHttpRequester.hh
SyrianBallaS: The main reason I did that was for configurations. Like if for some reason some using that file didn't have C++17, it'll do something else (or nothing if no alternative exists).

Might be better to define as `_MAYBE_UNUSED`?

tests/RA_UnitTestHelpers.h
SyrianBallaS: It's fine

src/services/impl/WindowsHttpRequester.cpp
SyrianBallaS: If you say so.

src/api/ApiCall.hh
SyrianBallaS: Alright

src/api/ApiCall.hh
SyrianBallaS: I got you.

PR #264, opened 2019-01-19
add license

Jamiras:


rzumer:
> I'm not a lawyer (and maybe I'm missing some possible implications), but reading the MIT license looks like it can be summarized as "do whatever you want but don't blame us if some sh*t happens".

That is pretty accurate.

PR #266, opened 2019-01-21
prevent exceptions unloading DLL in XP

Jamiras:
Prevents a crash when closing the application in XP with the Rich Presence dialog open.

### DIFF
--- a/src/ui/drawing/gdi/ImageRepository.cpp
+++ b/src/ui/drawing/gdi/ImageRepository.cpp
@@ -64,6 +64,8 @@ ImageRepository::~ImageRepository() noexcept
     for (auto& userPic : m_mUserPics)
         DeleteBitmap(userPic.second.m_hBitmap);
     m_mUserPics.clear();
+
+    g_pIWICFactory.Release();
 }
 
 std::wstring ImageRepository::GetFilename(ImageType nType, const std::string& sName)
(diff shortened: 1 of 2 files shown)


src/ui/drawing/gdi/ImageRepository.cpp
```diff
@@ -64,6 +64,8 @@ ImageRepository::~ImageRepository() noexcept
     for (auto& userPic : m_mUserPics)
         DeleteBitmap(userPic.second.m_hBitmap);
     m_mUserPics.clear();
+
+    g_pIWICFactory.Release();
```
Jamiras: `CComPtr` must be released before calling `Couninitialize`

src/ui/win32/DialogBase.cpp
```diff
@@ -19,6 +19,10 @@ DialogBase::~DialogBase() noexcept
 {
     if (m_hWnd)
     {
+        // remove the pointer to the instance from the HWND so the StaticDialogProc doesn't try to redirect
+        // the WM_DESTROY message through the instance we're trying to destroy.
+        ::SetWindowLongPtr(m_hWnd, DWLP_USER, 0L);
```
Jamiras: In XP, the destructor was being called before the window was destroyed, which caused line 26 to destroy the window. The window's OnDestroy handler called back into the Presenter to destroy the singleton instance, which was already being destroyed by the destructor, causing the crash. This prevents the message from being forwarded to the dialog that's already being destroyed.

PR #277, opened 2019-01-23
Remove achievement from GameContext, not just global list

Jamiras:


src/RA_Dlg_Achievement.cpp
```diff
@@ -632,27 +633,27 @@ INT_PTR Dlg_Achievements::AchievementsProc(HWND hDlg, UINT nMsg, WPARAM wParam,
...
-                            std::ostringstream oss;
-                            oss << "This achievement exists on " << _RA_HostName() << ".\n\n"
-                                << "*Removing it will affect other games*\n\n"
-                                << "Are you absolutely sure you want to delete this?";
-                            MessageBox(hDlg, NativeStr(oss.str()).c_str(), TEXT("Confirm Delete"),
-                                       MB_YESNO | MB_ICONWARNING);
+                            ra::ui::viewmodels::MessageBoxViewModel::ShowErrorMessage(L"Deleting achievements from the server is not supported.");
+                            break;
+
+                            //vmMessageBox.SetHeader(L"Are you sure that you want to delete this achievement?");
```
rzumer: Are the two commented lines here useful to keep or are they left over from your testing?

src/data/GameContext.cpp
```diff
@@ -504,6 +504,39 @@ Achievement& GameContext::NewAchievement(AchievementSet::Type nType)
     return pAchievement;
 }
 
+bool GameContext::RemoveAchievement(unsigned int nAchievementId) noexcept
+{
+    for (auto pIter = m_vAchievements.begin(); pIter != m_vAchievements.end(); ++pIter)
```
rzumer: How about `for (auto pIter : m_vAchievements)`? I believe this should work as well and it is a little bit less boilerplate-y.

src/RA_Dlg_Achievement.cpp
Jamiras: It's effectively retaining the previous message box text that was displayed, but the user's reaction was ignored.

src/data/GameContext.cpp
SyrianBallaS: It seems he needs the iterator for line 532.

Jamiras: Using the shorthand loop eliminates the iterator variable, which is needed for the erase call  (pIter would be a reference to an element).

rzumer: I see, my bad.

PR #286, opened 2019-01-27
use strtoul instead of stoul because we don't handle exceptions very well

Jamiras:
Legend of Zelda: Link's Awakening DX rich presence contained an invalid lookup entry:
```
oxda=Martha's Bay
```
The letter 'o' is not a valid hex prefix, so `stoul` threw an exception that was not caught. 

This PR replaces most of the instances of `std::stoul` with `std::strtoul`, which does not throw exceptions on parse error. Instead '0' is returned.

### DIFF
--- a/src/RA_CodeNotes.cpp
+++ b/src/RA_CodeNotes.cpp
@@ -36,7 +36,7 @@ size_t CodeNotes::Load(unsigned int nID)
             continue;
 
         const std::string& sAddr{note["Address"].GetString()};
-        const ra::ByteAddress nAddr{std::stoul(sAddr, nullptr, 16)};
+        const ra::ByteAddress nAddr = ra::ByteAddressFromString(sAddr);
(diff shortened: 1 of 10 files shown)


SyrianBallaS:
I guess it's ok but I'd suggest putting some todos about addressing/handling them later because it's still a bug. Mainly so its easier to log the exception to diagnose issues related to thus.

SyrianBallaS:
In c, the errno is set to ERANGE for out_of_range and 0 is returned for invalid_argument. So it's not like the bug magically disappeared.

Jamiras:
The only bug here is that an exception was introduced that was not being caught and crashed the program. In many of these cases, the expected behavior is that an unparseable value be converted to 0.

Look at the implementation of `ByteAddressFromString` in #262, which was attempting to address an isolated instance of this problem. This PR is attempting to address the remaining instances.

True, it would be better if bad data was reported, but we cannot ship something that crashes - especially on data that it didn't use to crash on.

Issue #284, opened 2019-01-26
x64 build


rzumer:
I would like to be able to link with x64 build of emulators.

This requires an x64 build configuration for `RAIntegration` and the ability to fetch the correct build from the server on update.

SyrianBallaS:
I've actually wanted to do something like this for a while. Anything that works on x64 will work on a x86 build. X64 is just short for x86_64. 

Shouldn't be too tough as I already did most of the work via code smell correction (not done yet, the code suffers from range bugs like buffer overrun) prs but will still need a little of work.

PR #288, opened 2019-01-27
Add RAOricutron to the emulator list

rzumer:
Also use the unused console ID of 32 for the Oric platforms.

Issue #287, opened 2019-01-27
Text input does not work when linking with SDL1 application


rzumer:
SDL1 will not handle text input automatically, which means that `RAIntegration` should provide its own event handlers for filling out text boxes, since events cannot be propagated from the client's end.

Keyboard event handlers are processed correctly, because navigation in the memory inspector, for example, works as expected.

While I have not found out anything about this issue in my research, I don't think this can be solved without change to `RAIntegration`, because the key events "belong" to it, i.e. I cannot intercept events on sub-windows with the client application.

rzumer:
Example: https://github.com/rzumer/oricutron/tree/cheevos (use the `RAOricutron.sln` solution in the `msvc\MSVC2017` directory)

Jamiras:
This is also a known issue for Allegro. 

The message loop must call `TranslateMessage` to convert the `WM_KEYDOWN` messages to `WM_CHAR` messages. It's not something we can really do anything about in our code.

https://docs.microsoft.com/en-us/windows/win32/winmsg/using-messages-and-message-queues
> If your application must obtain character input from the user, include the TranslateMessage function in the loop.

See also #758. A customized message loop is required to support tabbing between fields.

PR #289, opened 2019-01-28
x64 Configurations (implements #284)

SyrianBallaS:
The unit tests pass in all configurations but we need a 64-bit build of an RA Emulator to test with this.

**Breakdown**
- Basically to get this build as-is I just had to change the unit-tests a little bit to auto deduce param types because of type mismatches.
- Lots of narrowing conversions had to be done
   - Was unsure where on what "difference_type" I should've put, yet put the ones that made sense to me
   - These types are `std::ptrdiff_t`, `std::intptr_t`, or `gsl::index` (they are all the same, they are just aliases of the same type)
- Lots of adjustments to the math portions because everything's been extended to 8 bytes, the 32-bit is unaffected but the changes are required for the 64-bit build (like result of pointer math is `long long` instead of just `int`).


SyrianBallaS:
Because this is a big change I'm gonna mark it WIP until it's been tested and recommend on not merging until after release.

rzumer:
Thanks for working on it. `RAOricutron` has a 64-bit configuration, but because of #287 it is probably not a good target for testing this yet unless setting up a config compiling against SDL2.

rzumer:
I just went ahead and rolled back to SDL2 for `RAOricutron` after fixing a major issue. x64 is supported. I believe it is stable enough for testing, and all that is missing is an x64 configuration, so feel free to pull my branch at https://github.com/rzumer/oricutron/tree/cheevos and clone the "DebugAchievements" and/or "ReleaseAchievements" configuration(s) to x64.

SyrianBallaS:
Alright, I'll probably do that after work or something.

SyrianBallaS:
@rzumer I can't seem to get it to build. The `vcpkg` submodule refuses to get cloned. When I try to do it manually it gives me an error.
image

rzumer:
@SyrianBallaS I think you are missing the `--init` flag, which is needed to target a revision. If that does not work, try to pull my last commit; I changed the URLs to use HTTPS instead of SSH, maybe that will help. I cloned my repository again and had no problem with the submodule initialization, so if it still does not work after that, maybe you need to update your `git` version.

SyrianBallaS:
recloning worked.

rzumer:
This should only appear when debugging while performing operations like quitting or changing renderers or machines, and the application still functions despite them. Just ignore them. Personally I am not interested in fixing all the bugs in the emulators I am working with, but if you want to raise this issue in the upstream `oricutron` repository, feel free.

SyrianBallaS:
I can't even get it to run in a Win32 configuration with the released version of the toolkit. I'll need to know more requirements. I don't I've seen `RA_Implementation/Interface` anywhere either.

rzumer:
Right, you will need to merge in my branch from #288, sorry for omitting that detail.

SyrianBallaS:
Yeah I don't think I can pull in a branch from another repo, I'll try to see if I can do it with RANes in the mean time. That one doesn't build as-is for x64 because the callback functions have the wrong type (like `BOOL (int)` instead of `INT_PTR (int/long long)`. I'll probably PR a x64 config for RA stuff for that one too.

src/RA_Core.cpp
```diff
@@ -1136,9 +1125,16 @@ std::string GetFolderFromDialog()
...
-
+#if _MBCS
     std::array<TCHAR, RA_MAX_PATH> pDisplayName{};
     bi.pszDisplayName = pDisplayName.data();
+#elif _UNICODE
+    auto pDisplayName = std::make_unique<TCHAR[]>(RA_MAX_PATH);
+    bi.pszDisplayName = pDisplayName.get();
+#else
+#error Unknown Character Set!
+#endif /* _MBCS */
```
Jamiras: isn't the whole point of TCHAR to avoid things like this?

src/RA_Dlg_Memory.cpp
```diff
@@ -29,18 +29,18 @@ HWND Dlg_Memory::m_hWnd = nullptr;
...
+ra::ByteAddress MemoryViewerControl::m_nEditAddress = 0;
 unsigned int MemoryViewerControl::m_nEditNibble = 0;
 bool MemoryViewerControl::m_bHasCaret = 0;
 unsigned int MemoryViewerControl::m_nCaretWidth = 0;
 unsigned int MemoryViewerControl::m_nCaretHeight = 0;
 unsigned int MemoryViewerControl::m_nDisplayedLines = 8;
 unsigned short MemoryViewerControl::m_nActiveMemBank = 0;
 
-unsigned int m_nPage = 0;
+std::size_t m_nPage = 0;
```
Jamiras: can you explain why you use `gsl::index` in some places and `size_t` in others?

src/RA_LeaderboardPopup.cpp
```diff
@@ -171,12 +169,14 @@ _Use_decl_annotations_ void LeaderboardPopup::Render(ra::ui::drawing::ISurface&
             if (!pConfiguration.IsFeatureEnabled(ra::services::Feature::LeaderboardCounters))
                 break;
 
-            const auto& pSurfaceFactory = ra::services::ServiceLocator::Get<ra::ui::drawing::ISurfaceFactory>();
+            using default_isurfacefactory_t = ra::ui::drawing::ISurfaceFactory;
```
Jamiras: this alias is undesirable. it's barely shorted, and abstracts the usage in the lookup.

src/RA_LeaderboardPopup.cpp
```diff
@@ -263,8 +266,8 @@ _Use_decl_annotations_ void LeaderboardPopup::Render(ra::ui::drawing::ISurface&
 
                         const auto sScore = ra::Widen(pLB->FormatScore(lbInfo.m_nScore));
                         const auto szScore = m_pScoreboardSurface->MeasureText(nFontText, sScore);
-                        m_pScoreboardSurface->WriteText(m_pScoreboardSurface->GetWidth() - 4 - szScore.Width - 8, nY,
-                                                        nFontText, nTextColor, sScore);
+                        const auto nX = gsl::narrow_cast<int>(m_pScoreboardSurface->GetWidth()) - 4 - szScore.Width - 8;
```
Jamiras: this should use `iWidth` too.

src/RA_LeaderboardPopup.cpp
```diff
@@ -232,24 +233,26 @@ _Use_decl_annotations_ void LeaderboardPopup::Render(ra::ui::drawing::ISurface&
...
 
                     // background
-                    m_pScoreboardSurface->FillRectangle(0, 0, m_pScoreboardSurface->GetWidth(),
-                                                        m_pScoreboardSurface->GetHeight(), nColorBackground);
-                    m_pScoreboardSurface->FillRectangle(2, 2, m_pScoreboardSurface->GetWidth() - nShadowOffset,
-                                                        m_pScoreboardSurface->GetHeight() - nShadowOffset, nColorBlack);
-                    m_pScoreboardSurface->FillRectangle(0, 0, m_pScoreboardSurface->GetWidth() - nShadowOffset,
-                                                        m_pScoreboardSurface->GetHeight() - nShadowOffset,
+                    const auto iWidth = gsl::narrow_cast<int>(m_pScoreboardSurface->GetWidth());
+                    const auto iHeight = gsl::narrow_cast<int>(m_pScoreboardSurface->GetHeight());
```
Jamiras: suggest `nWidth` and `nHeight`. I don't think we use `i` as a prefix anywhere.

src/RA_LeaderboardPopup.cpp
```diff
@@ -232,24 +233,26 @@ _Use_decl_annotations_ void LeaderboardPopup::Render(ra::ui::drawing::ISurface&
...
                     m_pScoreboardSurface->WriteText(8, 3, nFontTitle, nColorBlack, sResultsTitle);
 
                     // scoreboard
-                    size_t nY = 4 + FONT_SIZE_TITLE + 4;
-                    size_t i = 0;
-                    while (i < pLB->GetRankInfoCount() && nY + FONT_SIZE_TEXT < m_pScoreboardSurface->GetHeight())
+                    auto nY = 4 + FONT_SIZE_TITLE + 4;
+                    gsl::index i = 0;
+                    while (i < ra::to_signed(pLB->GetRankInfoCount()) &&
+                           gsl::narrow_cast<std::ptrdiff_t>(nY + FONT_SIZE_TEXT) < m_pScoreboardSurface->GetHeight())
```
Jamiras: this should use `iHeight`

src/RA_StringUtils.h
```diff
@@ -309,8 +309,8 @@ class StringBuilder
         }
         else
         {
-            const int nLength = strlen(arg);
-            int nPadding = std::stoi(sFormat.c_str());
+            const auto nLength = std::strlen(arg);
+            std::ptrdiff_t nPadding{std::stoi(sFormat)};
```
Jamiras: another data type? why not `size_t`?

src/rcheevos.vcxproj
```diff
@@ -1,14 +1,46 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="DebugUnicode|Win32">
+      <Configuration>DebugUnicode</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
```
Jamiras: it doesn't make sense to have Unicode builds of this project. the solution configuration should just reference the Debug/Release configurations of this project.

src/services/SearchResults.cpp
```diff
@@ -402,20 +403,20 @@ void SearchResults::Initialize(const SearchResults& srSource, ComparisonType nCo
...
+    const auto nPadding = Padding(m_nSize);
+    unsigned int nCount{};
     for (auto& block : m_vBlocks)
         nCount += block.GetSize() - nPadding;
 
     if (m_nSize == MemSize::Nibble_Lower)
         nCount *= 2;
 
-    return nCount;
+    return {nCount}; // NB: The braces are required
```
Jamiras: why? just make `nCount` a `size_t`.

src/ui/drawing/ISurface.hh
```diff
@@ -23,12 +23,12 @@ public:
     /// <summary>
     /// Gets the width of the surface.
     /// </summary>
-    virtual size_t GetWidth() const = 0;
+    virtual std::ptrdiff_t GetWidth() const = 0;
```
Jamiras: why not `size_t`?

src/ui/drawing/gdi/GDIBitmapSurface.hh
```diff
@@ -89,9 +89,9 @@ public:
     GDIAlphaBitmapSurface& operator=(GDIAlphaBitmapSurface&&) noexcept = delete;
 
     void FillRectangle(int nX, int nY, int nWidth, int nHeight, Color nColor) noexcept override;
-    void WriteText(int nX, int nY, int nFont, Color nColor, const std::wstring& sText) override;
+    void WriteText(int nX, int nY, gsl::index nFont, Color nColor, const std::wstring& sText) override;
 
-    void Blend(HDC hTargetDC, int nX, int nY) const;
+    void Blend(HDC hTargetDC, std::ptrdiff_t nX, std::ptrdiff_t nY) const;
```
Jamiras: `nX` and `nY` should have a consistent data type. you've changed this to `ptrdiff_t`, but not `FillRectangle` or `WriteText` (probably others). 

is there a better type for coordinates? there's no pointers involved.

src/ui/drawing/gdi/ImageRepository.cpp
```diff
@@ -325,14 +325,18 @@ HBITMAP ImageRepository::LoadLocalPNG(const std::wstring& sFilename, size_t nWid
         if (SUCCEEDED(hr))
         {
             if (nWidth == 0 || nHeight == 0)
-                hr = pOriginalBitmapSource->GetSize(&nWidth, &nHeight);
+            {
+                auto nWidth32 = gsl::narrow_cast<std::uint32_t>(nWidth);
+                auto nHeight32 = gsl::narrow_cast<std::uint32_t>(nHeight);
+                hr = pOriginalBitmapSource->GetSize(&nWidth32, &nHeight32);
```
Jamiras: `GetSize` parameters are both out. the conversion should be from `nWidth32` to `nWidth`, not the other way around. this code is attempting to determine the width if it's not already known.

tests/Exports_Tests.cpp
```diff
@@ -244,7 +244,7 @@ TEST_CLASS(Exports_Tests)
         Assert::IsFalse(mockUserContext.IsLoggedIn());
         Assert::AreEqual(std::string(""), mockUserContext.GetUsername());
         Assert::AreEqual(std::string(""), mockUserContext.GetApiToken());
-        Assert::AreEqual(0U, mockUserContext.GetScore());
+        Assert::AreEqual({0U}, mockUserContext.GetScore());
```
Jamiras: why is this necessary?

tests/RA_StringUtils_Tests.cpp
```diff
@@ -285,11 +285,15 @@ TEST_CLASS(RA_StringUtils_Tests)
...
-        Assert::AreEqual(15, tokenizer.GetPointer(16U) - pStart); // string is 15 characters
+
+        // ptrdiff_t is 8 bytes but a ToString overload isn't working
+        Assert::AreEqual(0, gsl::narrow_cast<int>(tokenizer.GetPointer(0U) - pStart));
+        Assert::AreEqual(4, gsl::narrow_cast<int>(tokenizer.GetPointer(4U) - pStart));
+        Assert::AreEqual(8, gsl::narrow_cast<int>(tokenizer.GetPointer(8U) - pStart));
+        Assert::AreEqual(12, gsl::narrow_cast<int>(tokenizer.GetPointer(12U) - pStart));
+
+        // TODO: This is an error in logic and should not be considered success
+        Assert::AreEqual(15, gsl::narrow_cast<int>(tokenizer.GetPointer(16U) - pStart)); // string is 15 characters
```
Jamiras: this is validating intentional behavior. if you ask for a pointer beyond the end of the string, the end of the string should be returned.

src/RA_Core.cpp
SyrianBallaS: Ra_max_path has a different value for unicode, 32,767 characters, which is too big for the stack.

src/RA_Dlg_Memory.cpp
SyrianBallaS: In 64-bit based architectures (usually still varies a bit by micro), a user-mode application can't index past `PTRDIFF_MAX`. The result of pointer math is also `ptrdiff_t`. `gsl::index` is an alias of `ptrdiff_t `(which is either `int` or `long long` depending on the platform). I wasn't sure if this was supposed to be an index or not. It was already unsigned, because of truncation it needed to be different but still preserve the type for x86/Win32.

src/RA_LeaderboardPopup.cpp
SyrianBallaS: Sorry, I must have forgotten to get rid of it. I was originally trying to make the class a class template that accepted any integral type for the dimensions.

src/RA_LeaderboardPopup.cpp
SyrianBallaS: Sure, I guess `iWidth` should be `nWidth` as well.

src/RA_LeaderboardPopup.cpp
SyrianBallaS: sure

src/RA_LeaderboardPopup.cpp
SyrianBallaS: sure

src/RA_StringUtils.h
SyrianBallaS: because `stoi` returns int but a narrowing error might happen below it if it's not 8 bytes.

src/rcheevos.vcxproj
SyrianBallaS: sure, for the dropdown I'll get rid of the Unicode. I guess devs will have to turn it on in the configuration manager if they want to build a Unicode version. The Unicode builds where there just not in the dropdown near the task bar.

src/services/SearchResults.cpp
SyrianBallaS: The result of `block.GetSize() - nPadding` is `unsigned int`. The += with nCount as size_t would cause a narrowing error. size_t is unsigned long long in x64 based archectures. The way it was done was to satisfy both platforms. The braces disable narrowing. In Win32 it will be the same, in x64 it will be returned as a size_t without needing a narrowing conversion.

src/ui/drawing/ISurface.hh
SyrianBallaS: Because  size_t is not an appropriate type for dimensions or coordinates. Plus in all derived classes, the dimensions are signed.

src/ui/drawing/gdi/GDIBitmapSurface.hh
SyrianBallaS: There's `intptr_t`, or should we just make a `Point` structure? Or maybe alias Point from Position in types.hh. It does make sense to use to use int, I'll see if I can do that without narrowing, I do think pairs can abstracted.

`using Point = Position;`

tests/Exports_Tests.cpp
SyrianBallaS: To preserve the Win32 version. GetScore return size_t, but U is a literal for unsigned int, there exists no literal for size_t. If the braces aren't there it won't compile.

tests/RA_StringUtils_Tests.cpp
SyrianBallaS: Alright, I'll get rid of that comment when I get a chance.

src/RA_Core.cpp
SyrianBallaS: It has nothing to do with TCHAR

src/ui/drawing/gdi/ImageRepository.cpp
Jamiras: You don't understand. The point of calling GetSize is to set nWidth and nHeight. You've changed the code to pass in values and ignore the changes, so on line 339 below, nWidth and nHeight will both still be 0!

https://docs.microsoft.com/en-us/windows/desktop/api/wincodec/nf-wincodec-iwicbitmapsource-getsize

tests/Exports_Tests.cpp
Jamiras: `GetScore` returns `unsigned int`, and I don't see anything in this PR that changes that.

tests/api/ConnectedServer_Tests.cpp
```diff
@@ -164,8 +164,8 @@ TEST_CLASS(ConnectedServer_Tests)
         Assert::AreEqual(std::string(""), response.ErrorMessage);
         Assert::AreEqual(std::string(""), response.Username);
         Assert::AreEqual(std::string(""), response.ApiToken);
-        Assert::AreEqual(0U, response.Score);
-        Assert::AreEqual(0U, response.NumUnreadMessages);
+        Assert::AreEqual({0U}, response.Score);
+        Assert::AreEqual({0U}, response.NumUnreadMessages);
```
Jamiras: `Score` and `NumUnreadMessages` are unsigned ints. Please don't wrap stuff in brackets unless it's really necessary. It makes the code harder to read as it implies a conversion that is not occurring, and it makes the diff harder to read as there's more stuff to sift through.

tests/data/EmulatorContext_Tests.cpp
```diff
@@ -386,7 +386,7 @@ TEST_CLASS(EmulatorContext_Tests)
         emulator.mockServer.HandleRequest<ra::api::FetchUserUnlocks>([&bUnlocksRequested](const ra::api::FetchUserUnlocks::Request& request, ra::api::FetchUserUnlocks::Response& response)
         {
             bUnlocksRequested = true;
-            Assert::AreEqual(1U, request.GameId);
+            Assert::AreEqual({1U}, request.GameId);
```
Jamiras: `GameId` is an unsigned int.

tests/data/GameContext_Tests.cpp
```diff
@@ -192,7 +192,7 @@ TEST_CLASS(GameContext_Tests)
         Assert::AreEqual(5, pAch2->Category());
         Assert::AreEqual(1234567890, (int)pAch2->CreatedDate());
         Assert::AreEqual(1234599999, (int)pAch2->ModifiedDate());
-        Assert::AreEqual(15U, pAch2->Points());
+        Assert::AreEqual({15U}, pAch2->Points());
```
Jamiras: `Points` returns an unsigned int

tests/data/GameContext_Tests.cpp
```diff
@@ -351,13 +351,13 @@ TEST_CLASS(GameContext_Tests)
...
         Assert::AreEqual(1234500000, (int)pAch->ModifiedDate());
-        Assert::AreEqual(10U, pAch->Points());
+        Assert::AreEqual({10U}, pAch->Points());
         Assert::AreEqual(std::string("1=1"), pAch->CreateMemString());
 
         // new achievement should be allocated an ID higher than the largest existing local
         // ID, even if intermediate values are available
         const auto& pAch2 = game.NewAchievement(AchievementSet::Type::Local);
-        Assert::AreEqual(999000004U, pAch2.ID());
+        Assert::AreEqual({999000004U}, pAch2.ID());
```
Jamiras: ID is an AchievementId, which is a unit32_t

src/base.props
```diff
@@ -12,8 +12,8 @@
     <XPDeprecationWarning>false</XPDeprecationWarning>
   </PropertyGroup>
   <PropertyGroup>
-    <OutDir>$(SolutionDir)bin\$(Configuration)\</OutDir>
-    <IntDir>$(SolutionDir)obj\$(Configuration)\</IntDir>
+    <OutDir>$(SolutionDir)bin\$(Configuration)\$(Platform)\$(ProjectName)\</OutDir>
+    <IntDir>$(SolutionDir)obj\$(Configuration)\$(Platform)\$(ProjectName)\</IntDir>
```
Jamiras: Many of the projects override these, so they don't have any effect. Building Win32 debug creates a Win32\RAIntegration folder, but tests and rcheevos still end up at the top level.

And frankly, I prefer that structure. The dll should be at the top level, with the tests and rcheevos files in subdirectories. I'm fine with adding the $(Platform), but please remove $(ProjectName) - at least for RA_Integration.

src/services/SearchResults.cpp
Jamiras: This comment doesn't explain any of that, and as such, is not a very good comment.

I was unable to generate the error you describe. It did not appear in either the Win32 or x64 builds of the Debug or Analysis projects. There were two warnings that only appeared in the Win32 builds that should be addressed:
```
1>e:\source\raintegration\src\ra_achievementoverlay.cpp(837): warning C4018: '<': signed/unsigned mismatch
1>e:\source\raintegration\src\ra_core.cpp(1093): warning C4244: 'argument': conversion from 'std::streamsize' to 'size_t', possible loss of data
```
And one warning that only appeared in the x64 builds:
```
1>e:\source\raintegration\src\ra_dlg_acheditor.cpp(1750): warning C4267: '=': conversion from 'size_t' to 'unsigned int', possible loss of data
```

src/pch.h
```diff
@@ -39,6 +39,7 @@
 /* STL Stuff */
 #include <array> // algorithm, iterator, tuple
 #include <atomic>
+#include <filesystem>
```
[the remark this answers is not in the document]
SyrianBallaS: Got you, I have a different branch that uses standard api instead of Windows api. I guess this should go in IFileSystem implementations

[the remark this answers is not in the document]
SyrianBallaS: Standard filesystem api works on Windows, Linux, and Mac. Sure it's the only thing that make this portable but it's start.

src/ui/Types.hh
```diff
@@ -7,12 +7,28 @@ namespace ui {
...
+
 struct Size
 {
+#ifdef _WIN32_WINNT
+    [[nodiscard]] inline constexpr operator SIZE() const noexcept { return {Width, Height}; }
+    // can't constexpr as it needs automatic storage duration (stack storage)
+    [[nodiscard]] inline operator LPSIZE() const noexcept 
+    {
+        static SIZE sz{Width, Height};
+        return &sz;
```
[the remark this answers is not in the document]
SyrianBallaS: You're right, I'll get rid of this. For the `SIZE` one I'll add `explicit` to force a `static_cast`.

src/ui/drawing/ISurface.hh
```diff
@@ -48,43 +46,39 @@ public:
...
 
     /// <summary>
     /// Draws a secondary surface onto the surface.
     /// </summary>
-    /// <param name="nX">The x coordinate to draw at.</param>
-    /// <param name="nY">The y coordinate to draw at.</param>
+    /// <param name="nX">The point to draw at.</param>
     /// <param name="pImage">The surface to draw.</param>
-    virtual void DrawSurface(int nX, int nY, const ISurface& pSurface) = 0;
+    virtual void DrawSurface(const Point& nXY, const ISurface& pSurface) = 0;
```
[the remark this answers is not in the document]
SyrianBallaS: I haven't looked at this in awhile, built fine last time I did but the code did change.

src/ui/drawing/ISurface.hh
```diff
@@ -108,12 +102,12 @@ public:
...
     /// Creates a new offscren surface.
     /// </summary>
-    virtual std::unique_ptr<ISurface> CreateSurface(int nWidth, int nHeight) const = 0;
+    virtual std::unique_ptr<ISurface> CreateSurface(const Size& nWH) const = 0;
 
     /// <summary>
     /// Creates a new offscren surface with an alpha channel.
     /// </summary>
-    virtual std::unique_ptr<ISurface> CreateTransparentSurface(int nWidth, int nHeight) const = 0;
+    virtual std::unique_ptr<ISurface> CreateTransparentSurface(const Size& nWH) const = 0;
```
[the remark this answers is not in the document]
SyrianBallaS: It was to reduce number of parameters needed. Not a big deal. If you want they could be just passed by value because they are literal types.

To save headache I'll just put it back.

src/RA_Dlg_Memory.h
```diff
@@ -59,12 +59,12 @@ struct SearchResult
 {
     ra::services::SearchResults m_results;
 
-    std::vector<unsigned int> m_modifiedAddresses;
-    bool m_bUseLastValue = false;
-    unsigned int m_nLastQueryVal = 0;
-    ComparisonType m_nCompareType = ComparisonType::Equals;
+    std::vector<ra::ByteAddress> m_modifiedAddresses;
+    bool m_bUseLastValue{};
+    ra::ByteAddress m_nLastQueryVal{};
```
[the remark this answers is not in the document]
SyrianBallaS: Well it's a bit confusing when `ra::ByteAddress` is assigned to it. It'll change back.

src/ui/drawing/gdi/ImageRepository.cpp
SyrianBallaS: OK I got you.

SyrianBallaS:
I think this is ready, unless I missed a few things. 
Someone else will need to make a x64 build of an RA Emulator if they want to use it, don't have enough time to guess.

SyrianBallaS:
I'll reopen this later

PR #292, opened 2019-01-31
validate both source and target values

Jamiras:
Another crash caused by `stoul`. I missed it in the last pass because the `target` value field was catching the exception, so I assumed the `source` field was too. However, it was not, so it can crash (and was reported).

Modified `source` and `target` handlers to call a helper function to perform the same validation for both fields.

### DIFF
--- a/src/RA_Dlg_AchEditor.cpp
+++ b/src/RA_Dlg_AchEditor.cpp
@@ -1665,66 +1665,14 @@ INT_PTR Dlg_AchievementEditor::AchievementEditorProc(HWND hDlg, UINT uMsg, WPARA
 
                         case CondSubItems::Value_Src:
                         {
-                            int nBase = 16;
-                            if (rCond.CompSource().GetType() == CompVariable::Type::ValueComparison)
-                            {
-                                auto& pConfiguration =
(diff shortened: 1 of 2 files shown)


src/RA_Dlg_AchEditor.cpp
```diff
@@ -1783,6 +1731,80 @@ INT_PTR Dlg_AchievementEditor::AchievementEditorProc(HWND hDlg, UINT uMsg, WPARA
...
+    unsigned int nMax = 0xFFFFFFFF;
+
+    int nBase = 16;
+    if (nType == CompVariable::Type::ValueComparison)
+    {
+        auto& pConfiguration = ra::services::ServiceLocator::Get<ra::services::IConfiguration>();
+        if (pConfiguration.IsFeatureEnabled(ra::services::Feature::PreferDecimal))
+            nBase = 10;
+
+        // TODO: limit nMax to size of memory being read
```
rzumer: Why is nMax different with nBase = 10? Isn't the size is base 10 either way?

Jamiras: future feature. if the field is 8-bit, nMax should be 0xFF. 0xFFFF for 16-bit, etc.

Issue #297, opened 2019-02-05
Status code 12007 when updating the toolkit while it exists


rzumer:
I was getting this error on attempting an update from 0.74 to 0.75 using WINE and RAQUASI88.

meleu:
I got 12005 on attempting to update 0.74 to 0.75 using VirtualBox+Windows7 and RALibretro, also with RANes.

![error](https://user-images.githubusercontent.com/8508804/52263403-3a8f4d00-2916-11e9-8b3c-7f40c34a1f1d.png)

Worked fine when I deleted the old RA_Integration and launched the emulator again.

rzumer:
12005 is `ERROR_INTERNET_INVALID_URL` and 12007 is `ERROR_INTERNET_NAME_NOT_RESOLVED`.

Jamiras:
12007 usually occurs when a network hiccup occurs trying to perform the DNS lookup. It's in the list of automatic retry codes _within_ the DLL. The download logic for the DLL is not in the DLL, so doesn't have any retry logic. 12007 occasionally happens with the older DLLs as well.

12005 indicates the hostname is not parseable. I have never seen that.

Regarding WINE: I don't plan to support this, if it works, great, if not, well, tough luck. There's a lot more important stuff to be done first.

PR #300, opened 2019-02-05
make local copy of internal DLL data so its still valid if the DLL is unloaded

Jamiras:
also added retry on network fail logic

### DIFF
--- a/src/RA_Interface.cpp
+++ b/src/RA_Interface.cpp
@@ -368,6 +368,39 @@ static BOOL DoBlockingHttpGet(const char* sHostName, const char* sRequestedPage,
     return bSuccess;
 }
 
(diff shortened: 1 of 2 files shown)


src/RA_Interface.cpp
```diff
@@ -523,13 +556,13 @@ void RA_Init(HWND hMainHWND, int nConsoleID, const char* sClientVersion)
 {
     const char* sVerInstalled = _RA_InstallIntegration();
 
-    const char* sHostName = nullptr;
+    char sHostName[256] = "";
     if (_RA_HostName != nullptr)
-        sHostName = _RA_HostName();
```
Jamiras: This is what was causing the 12005 error. `_RA_HostName()` was returning a pointer to data in the DLL, so when the DLL gets unloaded to be replaced by the new one, the `sHostName` pointer was no longer pointing at valid data. If the valid data is replaced between lines 575 and 588, the host name may be garbage, which results in the 12005 error.

Jamiras:
fixes #297. The 12005 error is the result of pointing to memory that's been unloaded. The 12007 error will be handled by the automatic retry logic.

PR #303, opened 2019-02-06
Always Check Return of CRT Functions

SyrianBallaS:
This addresses a lot of different things, but not completely (bounds/ranges/errors).
This basically just tells the compiler to tell us that a return value was ignored even if it was optional in the C library. In C11, the safe functions call a constraint handler (function pointer callback) to handle any runtime errors that happen (similar to exception handling).

Even though the "safe" functions are somewhat similar to the C11 spec, it doesn't handle the error at all and just crashes a module/app if not handled (usually leading to segmentation faults).

Most of the time there are functions being replaced but sometimes we just checked the value since it didn't make sense to replace it.

### DIFF
--- a/RA_Integration.sln
+++ b/RA_Integration.sln
@@ -19,7 +19,6 @@ EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Analysis|Win32 = Analysis|Win32
-		AnalysisUnicode|Win32 = AnalysisUnicode|Win32
 		Debug|Win32 = Debug|Win32
 		Release|Win32 = Release|Win32
 		ReleaseWithDebug|Win32 = ReleaseWithDebug|Win32
@@ -27,8 +26,6 @@ Global
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{825CB292-F069-4B27-82CF-3417746ACDF3}.Analysis|Win32.ActiveCfg = Analysis|Win32
 		{825CB292-F069-4B27-82CF-3417746ACDF3}.Analysis|Win32.Build.0 = Analysis|Win32
-		{825CB292-F069-4B27-82CF-3417746ACDF3}.AnalysisUnicode|Win32.ActiveCfg = AnalysisUnicode|Win32
-		{825CB292-F069-4B27-82CF-3417746ACDF3}.AnalysisUnicode|Win32.Build.0 = AnalysisUnicode|Win32
 		{825CB292-F069-4B27-82CF-3417746ACDF3}.Debug|Win32.ActiveCfg = Debug|Win32
 		{825CB292-F069-4B27-82CF-3417746ACDF3}.Debug|Win32.Build.0 = Debug|Win32
 		{825CB292-F069-4B27-82CF-3417746ACDF3}.Release|Win32.ActiveCfg = Release|Win32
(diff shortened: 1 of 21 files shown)


RA_Integration.sln
```diff
@@ -19,16 +19,13 @@ EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Analysis|Win32 = Analysis|Win32
-		AnalysisUnicode|Win32 = AnalysisUnicode|Win32
```
SyrianBallaS: Noticed this solution config was here, remembered in another PR that we don't need solution-wide unicode configurations.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -83,10 +83,15 @@ INT_PTR CALLBACK AchProgressProc(HWND hDlg, UINT nMsg, WPARAM wParam, _UNUSED LP
...
+    gsl::index i = -1;
+    for (auto& name : m_lbxGroupNames)
     {
-        const auto i = std::distance(std::next(m_lbxGroupNames.begin()), it);
-        *it = ra::StringPrintf(_T("Alt %02t"), i);
+        i++;
+        if (i == 0)
+            continue;
+        else
+            name = ra::StringPrintf(_T("Alt %02t"), i);
```
SyrianBallaS: This was giving me a weird warning. Something about the offset being negative which didn't make any sense. So I did this way instead.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -235,8 +241,8 @@ LRESULT CALLBACK EditProc(HWND hwnd, UINT nMsg, WPARAM wParam, LPARAM lParam) no
             lvDispinfo.hdr.idFrom = GetDlgCtrlID(hwnd);
             // inline suppression not working, and by function not working, have to disable it by code
 
-#pragma warning(suppress: 26454)
-            GSL_SUPPRESS(io.5) lvDispinfo.hdr.code = LVN_ENDLABELEDIT;
+#pragma warning(suppress : 26454)
+            GSL_SUPPRESS(io .5) lvDispinfo.hdr.code = LVN_ENDLABELEDIT;
```
SyrianBallaS: I was doing this on a different computer, I'll fix these as soon as I can.

src/RA_Dlg_AchievementsReporter.cpp
```diff
@@ -45,34 +45,39 @@ void Dlg_AchievementsReporter::SetupColumns(HWND hList)
...
     // We aren't actually using the value so we're using iterators
     for (auto it = COL_TITLE.cbegin(); it != COL_TITLE.cend(); ++it)
     {
         const auto nPos{std::distance(COL_TITLE.cbegin(), it)};
+
         switch (ra::itoe<Column>(nPos))
         {
             case Column::Checked:
-                sprintf_s(LbxDataAt(ms_nNumOccupiedRows, nPos), MAX_TEXT_LEN, "%s", "");
+                check = sprintf_s(LbxDataAt(ms_nNumOccupiedRows, nPos), MAX_TEXT_LEN, "%s", "");
```
SyrianBallaS: Since they array uses `char*` instead of `std::string`, we just checked the value.

src/RA_Dlg_GameLibrary.cpp
```diff
@@ -76,8 +76,7 @@ namespace ra {
 inline static void LogErrno() noexcept
 {
     std::array<char, 2048> buf{};
-    strerror_s(buf.data(), sizeof(buf), errno);
-    // TODO: Make StringPrintf support std::array
+    GSL_SUPPRESS_F6 Ensures(strerror_s(buf.data(), sizeof(buf), errno) == 0);
```
SyrianBallaS: Don't know if there's a C++ version of this function so it's still using the same one. This just needs to make sure the function only returns an `errno` of 0 (no error)

src/RA_Integration.vcxproj
```diff
@@ -462,7 +462,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='AnalysisUnicode|Win32'">
     <ClCompile>
-      <PreprocessorDefinitions>_DEBUG;RA_EXPORTS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_DEBUG;RA_EXPORTS;_PREFAST_;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
SyrianBallaS: Visual Studio wasn't automatically defining `_PREFAST_` even though the Microsoft Recommended Ruleset is used..

src/RA_md5factory.cpp
```diff
@@ -2,65 +2,56 @@
...
-
-    delete[] pDataBuffer;
-    pDataBuffer = nullptr;
-
-    return buffer;
+    md5_append(&pms, pDataBuffer.get(), sStringToMD5.length());
+    md5_finish(&pms, digest.data());
+
+    std::array<char, MD5_STRING_LEN + 1> buffer{};
+    Ensures(sprintf_s(buffer.data(), buffer.size(),
```
SyrianBallaS: Using `StringPrintf` with `std::string` seems to dangle the stored pointer, using `std::array` didn't dangle it though.

SyrianBallaS:
I'll reopen later

PR #306, opened 2019-02-08
support patch numbers in version

Jamiras:
Allows for tagging a release with a patch version: i.e. "0.75.1"

Previously the patch field was used to indicate the number of commits since the previous tag. Now that data is in the revision field (fourth field). The revision field was previously used to indicate the presence of non-committed changes. This change includes that as part of the revision, so "0.75.1.12" would indicate either 12 commits since "0.75.1" or 11 commits since "0.75.1" and local changes.


Jamiras:
I'm going to merge this early, I want to try to get the performance improvements available before the weekend rush.

PR #307, opened 2019-02-13
only update list view subitems that change

Jamiras:
Reported by @kdecks: 
>While an achievement is active and the window is open various conditions flicker white without pattern 2-3 times per second.
> About performance, the game is where the performance struggles. There is noticeable slowdown even with 15 conditions, and it increases from there.

Something between 0.74 and 0.75 significantly increased the time spent updating the achievement editor listview when the selected achievement was active. The editor was blindly repainting the listview for every frame in case the achievement's hitcounts changed.

In 0.74, the average time to repaint the listview (on my machine, release build) with 40 conditions was 10.8ms. In 0.75, it's 23.7ms (119% slower). 

Assuming an emulator runs at 60fps, that only gives us 16ms per frame to do whatever needs to be done for each frame, and that includes everything the emulator is doing for that frame as well. As we start to exceed that threshhold, we actually slow down the emulator.

I couldn't narrow down exactly which change introduced the overhead - I think it was a combination of several: spectre mitigation, don't use std library functions that aren't bounds checked, and replacing sprintf(buffer) with StringPrintf are the top contenders.

With these changes, it only takes 0.4ms to update the listview (96% faster than 0.74). The basic principal is we already keep a copy of the stringified data in memory, so we can tell when the string changes, and the UI should only be notified to repaint when that occurs.

### DIFF
--- a/src/RA_Dlg_AchEditor.cpp
+++ b/src/RA_Dlg_AchEditor.cpp
@@ -185,7 +185,7 @@ void Dlg_AchievementEditor::SetCell(HWND hList, LV_ITEM& lvItem, int nRow, CondS
     }
 }
 
-void Dlg_AchievementEditor::SetCell(HWND hList, LV_ITEM& lvItem, int nRow, CondSubItems nColumn, const std::string&& sNewValue)
+void Dlg_AchievementEditor::SetCell(HWND hList, LV_ITEM& lvItem, int nRow, CondSubItems nColumn, std::string&& sNewValue)
 {
     std::string& sCell = LbxDataAt(nRow, nColumn);
     if (sCell != sNewValue)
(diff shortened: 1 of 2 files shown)


rzumer:
Seems good, I suspect a lot of the extra safety checks added recently have a significant performance cost. Would it be possible to do the same for the memory viewer in a future PR? I'm not sure how similar the refresh process is, or what the performance impact is for having it open, but it also flickers a lot for me during updates.

src/RA_Dlg_AchEditor.cpp
```diff
@@ -156,6 +173,30 @@ const int Dlg_AchievementEditor::AddCondition(HWND hList, const Condition& Cond,
...
+    if (sCell != sNewValue)
+    {
+        sCell.assign(sNewValue);
+        lvItem.iSubItem = ra::etoi(nColumn);
+        lvItem.pszText = sCell.data();
+        ListView_SetItem(hList, &lvItem);
+    }
+}
+
+void Dlg_AchievementEditor::SetCell(HWND hList, LV_ITEM& lvItem, int nRow, CondSubItems nColumn, const std::string&& sNewValue)
```
SyrianBallaS: `sNewValue` shouldn't be `const` here.
Also in the declaration.

PR #311, opened 2019-02-16
Create ConsoleContext for console-specific logic

Jamiras:
Starting with memory regions.

Also eliminates `g_ConsoleID`

### DIFF
--- a/src/Exports.cpp
+++ b/src/Exports.cpp
@@ -21,6 +21,10 @@
 #include "ui\viewmodels\MessageBoxViewModel.hh"
 #include "ui\viewmodels\OverlayManager.hh"
 
+#ifndef RA_UTEST
+#include "RA_Dlg_Memory.h"
+#endif
+
 API const char* CCONV _RA_IntegrationVersion() { return RA_INTEGRATION_VERSION; }
 
 API const char* CCONV _RA_HostName()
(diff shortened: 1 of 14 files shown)


rzumer:
That's a lot of boilerplate to define every context. It would be nice to use macros and/or templates to reduce that considering a lot of it is the same.

rzumer:
Going back to my previous comment, is there a reason why we can't just fill the memory regions parameter in a struct for each platform instead of creating a concrete class with empty constructor and accessor functions?

Jamiras:
1) encapsulation - all data related to a single console is in one class instead of spread out over multiple switch statements. Functions that apply to that data are also in the class so you don't have to remember where they are (i.e. `GetMemoryRegion(address)`).
2) subclassing - default behavior (such as not defining a memory map) is handled by the base class without having to explicitly provide the default value. Additionally, similar consoles can be handled through subclass-chains. It wouldn't be unexpected to have `GameBoyColorContext` inherit from `GameBoyContext` since it's basically just an upgraded version of the same hardware.

It may seem like redundant code now, but it simplifies expansion. You can quickly see what behaviors can be modified by looking at the virtual functions in the header file and determine which apply to the new system. 

Right now, the base class only provides a few getters and one method (`GetMemoryRegion(address)`).

At some point, I'd like to add "real address mapping". A method would be called on the `EmulatorContext` with the RA address and it would return the actual memory address (and vice-versa). The base class would just return the input, but the other classes could do whatever logic is appropriate for their systems. A similar function could also provided to normalize mirrored addresses. 

Another idea would be to put the hashing code in the DLL. The base class would hash the whole file, and subclasses could handle special cases. 

The inverse is something along the lines of `EmulatorContext` where a switch statement initializes all the fields of the class. The downside of this approach is things are initialized even if they're not needed, and it's a lot harder to provide unique code implementations per case - you either have to use function pointers or introduce more independent switch statements.

The emulator implementation code is mostly in the emulator, so `EmulatorContext` is implemented using function pointers, and it makes sense to implement it solely as a container of variables. `ConsoleContext` is internalized and shared between emulators, so anything console-specific will be implemented internally, and there's a much higher chance of having unique implementations per instance. As such, it feels better suited to a subclassing model where the function pointers are managed by the vtable.

I'm not going to argue that one way is better than the other, as I clearly use both. I would argue that either solution is better than having switch statements in each location where case-specific logic is required.

PR #315, opened 2019-02-18
Unregister custom class handler when Shutdown is called (fixes #314)

### LINKED ISSUE #314: Heap allocation(s) leaked on shutdown
In RAppleWin I need to be able to reset RA in order to switch window handles, because of the way the emulator handles restarts. For this I call `RA_Shutdown()` and reinitialize the system. However RAIntegration seems to be leaking heap allocations which cause subsequent crashes or unstable behavior. For example when switching systems (prompting a restart) and then trying to open the Memory Inspector dialog after the restart is completed my application will just crash. In addition when pressing a button on the right-hand side it will get stuck, which does not happen when I comment out the shutdown/init calls on restart.

Below is debug output from Application Verifier/WinDbg. `du` confirms that the owner of the leaked allocation is `RA_Integration.dll` and `dps` shows the allocation stack, which points to `RA_InstallSharedFunctions()`.

```
0:000:x86> du 13463FD8
13463fd8  "RA_Integration.dll"

0:000:x86> dps 054479C4
054479c4  053570fc
054479c8  00009002
054479cc  00120000
054479d0  0f46abb0 verifier_f460000!AVrfDebugPageHeapAllocate+0x240
054479d4  7761246b ntdll_77530000!RtlDebugAllocateHeap+0x39
054479d8  77576dd9 ntdll_77530000!RtlpAllocateHeap+0xf9
054479dc  77575ec9 ntdll_77530000!RtlpAllocateHeapInternal+0x179
054479e0  77575d3e ntdll_77530000!RtlAllocateHeap+0x3e
054479e4  0f54aaaf vrfcore!VfCoreRtlAllocateHeap+0x1f
054479e8  0fad71ec vfbasics!AVrfpRtlAllocateHeap+0xdc
054479ec  136cad74*** WARNING: Unable to verify checksum for C:\Users\rzumer\Projects\AppleWin\Debug\RA_Integration.dll
*** ERROR: Symbol file could not be found.  Defaulted to export symbols for C:\Users\rzumer\Projects\AppleWin\Debug\RA_Integration.dll - 
 RA_Integration!RA_InstallSharedFunctionsExt+0x92a64
054479f0  1369e75c RA_Integration!RA_InstallSharedFunctionsExt+0x6644c
054479f4  1365e3a4 RA_Integration!RA_InstallSharedFunctionsExt+0x26094
054479f8  1367b917 RA_Integration!RA_InstallSharedFunctionsExt+0x43607
054479fc  1367d1b2 RA_Integration!RA_InstallSharedFunctionsExt+0x44ea2
05447a00  1364fdd8 RA_Integration!RA_InstallSharedFunctionsExt+0x17ac8
05447a04  136c5b17 RA_Integration!RA_InstallSharedFunctionsExt+0x8d807
05447a08  0fad2d78 vfbasics!AVrfpStandardThreadFunction+0x48
05447a0c  75228484 KERNEL32!BaseThreadInitThunk+0x24
05447a10  775941c8 ntdll_77530000!__RtlUserThreadStart+0x2f
05447a14  77594198 ntdll_77530000!_RtlUserThreadStart+0x1b
05447a18  00000000
05447a1c  05415c8c
05447a20  0000c000
05447a24  00180000
05447a28  0541c620
05447a2c  7761246b ntdll_77530000!RtlDebugAllocateHeap+0x39
05447a30  77576dd9 ntdll_77530000!RtlpAllocateHeap+0xf9
05447a34  77575ec9 ntdll_77530000!RtlpAllocateHeapInternal+0x179
05447a38  77575d3e ntdll_77530000!RtlAllocateHeap+0x3e
05447a3c  0f54aaaf vrfcore!VfCoreRtlAllocateHeap+0x1f
05447a40  0fad71ec vfbasics!AVrfpRtlAllocateHeap+0xdc
```

In case it matters, the shutdown->init process calls the following RA interface functions in order:

1. RA_Shutdown()
2. RA_Init()
3. RA_InstallSharedFunctions()
4. RA_AttemptLogin()
5. RA_CreatePopupMenu()
6. RA_ClearMemoryBanks()
7. RA_InstallMemoryBank() (twice)

Besides `RA_Shutdown()`, which is called as soon as a restart is confirmed, each of these functions is called in the exact same order, as part of the same boot code, as on an initial start.

Jamiras:
The memory viewer uses a custom `WNDCLASS`, which has a pointer to a `WndProc`. When the DLL is unloaded and reloaded, the attempt to register the `WNDCLASS` fails as the previous one is still registered. Despite associating it with the DLL instance, it only automatically gets unregistered when the application terminates. As a result, the _new_ memory viewer that is created has a `WndProc` pointing at some random place in memory.

Solution: during `RA_Shutdown`, unregister the custom `WNDCLASS`, so it can be correctly re-registered if the DLL is reloaded.

### DIFF
--- a/src/RA_Core.cpp
+++ b/src/RA_Core.cpp
@@ -56,9 +56,11 @@ bool g_bRAMTamperedWith = false;
 BOOL APIENTRY DllMain(HMODULE hModule, DWORD dwReason, _UNUSED LPVOID)
 {
     if (dwReason == DLL_PROCESS_ATTACH)
+    {
         g_hThisDLLInst = hModule;
 
-    ra::services::Initialization::RegisterCoreServices();
+        ra::services::Initialization::RegisterCoreServices();
+    }
 
     return TRUE;
 }
@@ -165,6 +167,7 @@ API int CCONV _RA_Shutdown()
         DestroyWindow(g_MemoryDialog.GetHWND());
         g_MemoryDialog.InstallHWND(nullptr);
     }
+    g_MemoryDialog.Shutdown();
 
     if (g_GameLibrary.GetHWND() != nullptr)
     {
--- a/src/RA_Dlg_Memory.cpp
+++ b/src/RA_Dlg_Memory.cpp
@@ -765,6 +765,11 @@ void Dlg_Memory::Init() noexcept
     ASSERT(checkAtom != 0);
 }
 
+void Dlg_Memory::Shutdown() noexcept
+{
+    ::UnregisterClass(TEXT("MemoryViewerControl"), g_hThisDLLInst);
+}
+
 // static
 INT_PTR CALLBACK Dlg_Memory::s_MemoryProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
 {
--- a/src/RA_Dlg_Memory.h
+++ b/src/RA_Dlg_Memory.h
@@ -74,6 +74,7 @@ class Dlg_Memory
 {
 public:
     void Init() noexcept;
+    void Shutdown() noexcept;
 
     void ClearLogOutput() noexcept;
 


src/RA_Core.cpp
```diff
@@ -56,9 +56,11 @@ bool g_bRAMTamperedWith = false;
 BOOL APIENTRY DllMain(HMODULE hModule, DWORD dwReason, _UNUSED LPVOID)
 {
     if (dwReason == DLL_PROCESS_ATTACH)
+    {
         g_hThisDLLInst = hModule;
 
-    ra::services::Initialization::RegisterCoreServices();
+        ra::services::Initialization::RegisterCoreServices();
+    }
```
Jamiras: This change is not directly related to the issue. It's just a minor optimization, as the services only need to be registered once on DLL load. When `DllMain` is called for `DLL_THREAD_ATTACH` or `DLL_PROCESS_DETACH`, this call does nothing as internally, it short-circuits if it's already been called.

PR #320, opened 2019-02-23
move emulator callbacks into EmulatorContext

Jamiras:
* deprecates `fpIsActive` callback. All previous calls to `fpIsActive` now check `GameId != 0`
* `fpLoadROM` is associated to the game library dialog, which is currently disabled, but not removed.
* other "shared" functions are loaded into and called through the `EmulatorContext` class.

### DIFF
--- a/src/Exports.cpp
+++ b/src/Exports.cpp
@@ -27,6 +27,7 @@
 #ifndef RA_UTEST
 #include "RA_Dlg_Achievement.h"
 #include "RA_Dlg_AchEditor.h"
+#include "RA_Dlg_GameLibrary.h"
 #include "RA_Dlg_Memory.h"
 #endif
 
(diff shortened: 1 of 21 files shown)


rzumer:
I didn't realize the implication of this change when it was posted, but it adds an additional dependency on the emulator side, which I don't like. I think only headers should be needed when compiling an emulator, but now even with the full interface (header and source files) included with emulator sources I get compilation errors. I would rather revert this to avoid having to add extra files from RAI as project dependencies.

Jamiras:
The intent was solely to move the function pointers from global pointers to class members. The global pointers just happened to be in `RA_Interface.cpp`. Those pointers point to the methods _you_ registered when you called `RA_InitSharedFunctions`. If you are attempting to call them through the global pointer, I recommend just calling the function that you registered.

`RA_Shutdown` was not touched. Why do you always have to create such vague issue reports? How hard is it to copy/paste the actual error you're getting?

Jamiras:
Ah, changes in #335 make sense. As I mentioned, the global pointers were eliminated. Find All References must not have found those because of the `#ifndef RA_EXPORTS`.

PR #321, opened 2019-02-23
don't clear mem viewer when game unloaded (fixes #302)

### LINKED ISSUE #302: Enable/Disable clearing Memory Viewer when game unloaded
The new RAIntegration have a lot of cool new features, ones that I would love to give it a try, thanks Jamiras and Syrian 😁

When I was reading the new features I saw this "bug":
• clear out memory viewer when game is unloaded

I propose this option to be a toggle if the user want it activated or not, or maybe to be like before, why?

Before the update I ever thought wasn't a bug, was very useful to be honest I thought it was a feature lol

There's is times when I'm searching a value in memory and the emulator crash and I lost all the progress I have in the memory viewer (and memory bookmarker), this is a very normal situation to be honest, especially when I'm searching a variable by freezing other ones in the memory bookmarker and the emulator can handle it and crash

I would like to propose too, if a "feature" like this can be implemented in the memory bookmarker too?

Will be very useful and I will no longer need saving a .txt of the bookmark list every 5segs if the emu crash 😅

Grettings ✌️

Jamiras:
I've changed the behavior of the memory viewer when a game is unloaded. Instead of clearing out the memory (resetting to the initial state when the emulator is started), it now just disables the Notes section. All of the notes/bookmarks/search results from the last loaded game will remain until a new game is loaded. They will also remain if the same game is reloaded.

However, reading/writing the memory is dependent on the emulator, and some of the emulators free the memory associated to the game when the game is unloaded. Attempting to read or write this memory while a game is not loaded _will crash the emulator_. Use at your own risk!

### DIFF
--- a/src/RA_Dlg_MemBookmark.cpp
+++ b/src/RA_Dlg_MemBookmark.cpp
@@ -674,8 +674,7 @@ unsigned int Dlg_MemBookmark::GetMemory(unsigned int nAddr, int type)
 
 void Dlg_MemBookmark::ExportJSON()
 {
-    const auto& pGameContext = ra::services::ServiceLocator::Get<ra::data::GameContext>();
-    if (pGameContext.GameId() == 0)
+    if (m_nGameId == 0)
     {
         ra::ui::viewmodels::MessageBoxViewModel::ShowErrorMessage(L"ROM not loaded: please load a ROM first!");
         return;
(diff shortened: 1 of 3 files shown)


rzumer:
Thank you.

Which emulators free allocated RAM? I know that Oricutron has variable-sized RAM and reallocates it when changing machines, but other than that all emulators I have seen allocate the maximum possible amount of memory and leave any unused RAM uninitialized.

Jamiras:
At a minimum, RANes does. If you try to do a search after the game is unloaded, it crashes. The other emulators where I could figure out how to unload the game seem to survive through that specific test, but it doesn't guarantee that the memory contents are still valid, only that's the memory block is still accessible to the program.

rzumer:
I don't mind making necessary changes like adding bounds checking to the memory reader, but after a cursory search I could not find where it is set up in those older emulators. I wanted to look into memory mapping in RAVBA as well but the interface implementations for both seem incomplete or broken up into multiple parts.

PR #328, opened 2019-03-03
Show memory region watermark in search results if no code note exists

Jamiras:
[Old implementation](http://docs.retroachievements.org/Memory-Inspector-Overview/#4-results-window):
image
* Notes were wrapped in parenthesis
* Entries without notes were just left blank
----------
New implementation (in RANes):
image
* Entries without notes have watermarks identifying how the address is used ([if defined in toolkit](https://github.com/RetroAchievements/RAIntegration/blob/master/src/data/ConsoleContext.cpp#L272)).
* If highest known address is less than 0x010000, only four digits are displayed in the address field.

Another example, showing highlights for existing notes (blue) and bookmarked items (green):
image
* Notes are no longer wrapped in parenthesis

Example from RASnes9x showing data that no longer matches the filter (red):
image


kdecks:
This is really cool

Issue #230, opened 2019-01-02
Linking does not work for games that contain non-ASCII characters


rzumer:
I used a regular apostrophe, because the website currently duplicates typewrite apostrophes in game titles. On linking, the title is garbled (as expected). What is unexpected is that when trying to link, a second entry is created with different garbled text, presumably creating a new entry for the game based on the incorrectly-decoded name of the first one.
![example](https://user-images.githubusercontent.com/7488362/50599508-4d78b280-0e7c-11e9-9446-ace79a4d38aa.png)

Jamiras:
Which version of the toolkit were you using? The 0.74 (official release) is known to not be escaping URL parameters correctly. 

Using the test string 
```Tést ' / © ½ Test```
I get this URL parameter in 0.74:
```i=Tést+'+/+©+½+Test```
In current master, I get this:
```i=T%E9st+%27+%2F+%A9+%BD+Test```

Note that %E9, %A9, and %BD are Latin-1 sequences and should be sent to the server as UTF-8 sequences, so even properly escaped, I wouldn't expect them to behave perfectly. Unicode support is currently known to be very buggy and minimally supported.

But that doesn't explain why "/" would be getting converted to "_" - that must be happening on the server. Same with the apostrophe getting converted into a double apostrophe.

As for the unicode characters - the dropdown is populated using single-byte Windows APIs. Similarly, the text box is read using single-byte APIs. I wouldn't expect either to support Unicode characters at this time. I am unable to paste Japanese into the test field (it gets converts to ???? on paste). If you're getting them in there some other way, I'm not surprised the code trying to read it out is confused.

rzumer:
I'm quite sure I was using a recent master version compiled this week. Reproduction should be straightforward, but it is possible that it's a back-end issue. I'm just surprised that linking relies on any component being able to decode a string correctly.

Jamiras:
The API inputs are console ID, game title, and MD5. The response is the game ID. The association to an existing game ID is purely reliant on the title matching: https://github.com/RetroAchievements/RAWeb/blob/1881a9be7aaa2df503f35282fd8e0c3364998b80/lib/database/game.php#L1097

Jamiras:
And you can see the replacing of "/" with "-" here: https://github.com/RetroAchievements/RAWeb/blob/1881a9be7aaa2df503f35282fd8e0c3364998b80/lib/database/game.php#L1102

rzumer:
[...]

Issue #140, opened 2018-10-16
show ConsoleName in "Unknown Title" dialog box


meleu:
As a RALibretro (multi-emulator) user I would like to know in what console library I'm about to add a game title when this dialog appears:
![image](https://camo.githubusercontent.com/be4a9d2919e4ac631ad78c6bb125eed24218c3c0/68747470733a2f2f692e696d6775722e636f6d2f7a4a4f6d6c58432e706e67)

Suggestion: put in the window title something like this: `Unknown Title for ConsoleName`.

salvadorc17:
This is not Integration issue, or just Ralibretro dont send console name to integration, when creating a rom link with RA_Onloadnewrom function

meleu:
:gitgud:

Jamiras:
>This is not Integration issue, or just Ralibretro dont send console name to integration, when creating a rom link with RA_Onloadnewrom function

I'm not sure what you mean by this. `RA_OnLoadNewRom` only takes the ROM pointer and size. `RA_SetConsoleID` is used to indicate which console is running, but that doesn't currently have any effect on this particular dialog. Nor does anything in the dialog look at the current console (or current client). 

I do believe this is an integration issue.

SyrianBallaS:
I can volunteer to do this. I tried it just for RANes and it shows "NES" instead of "Unknown Title". It's pretty simple I'll just do the same thing for all the ConsoleIDs.

Jamiras:
When that gets converted to a viewmodel, the title will be the emulator name. I was planning to change "Unrecognized Title" to "Unrecognized Title: NES" or something along those lines. 

If you're ready to start working on features, I should probably formalize and share my roadmap. I'll try to get something put together this weekend.

If you want something to start looking at now, you could try to convert the RequestGamesList and SubmitNewTitle APIs, which are used by this dialog.

SyrianBallaS:
There's a few rule sets I want to apply before working on large scale things so the analysis can be more of a maintenance thing. Having a roadmap would help though.

PR #334, opened 2019-03-14
Update Unknown Game dialog

Jamiras:
Old dialog: 

image

New dialog:

image

List of changes:
* Emulator provided name for the game is displayed at all times. Previously, it was initially displayed in the text box, but that could be changed by typing or selecting an item from the dropdown.
* Associated system is displayed (addresses #140)
* Checksum is moved up to the game details section, and a button is added to copy the checksum to the clipboard.
* Test button added - non-functional at this point. Plan is to allow loading achievements for an existing game without actually linking the hash to allow for compatibility testing. More details on that functionality will be in the related PR.
* Submit button changed to "Link" and moved to the left side of the dialog. Link button now prompts the user to confirm that they meant to associate the checksum to an existing entry or create a new entry.
image
image

Additionally, when associating the hash to an existing game, the game ID is now passed to the server as the 'g' parameter. This will allow the server to make the correct association when certain characters are present in the title (see #230) (requires server changes).

### DIFF
--- a/src/RA_Core.cpp
+++ b/src/RA_Core.cpp
@@ -267,8 +267,10 @@ static unsigned int IdentifyRom(const BYTE* pROM, unsigned int nROMSize, std::st
 
                     ra::ui::viewmodels::UnknownGameViewModel vmUnknownGame;
                     vmUnknownGame.InitializeGameTitles();
+                    vmUnknownGame.SetSystemName(ra::services::ServiceLocator::Get<ra::data::ConsoleContext>().Name());
                     vmUnknownGame.SetChecksum(ra::Widen(sCurrentROMMD5));
-                    vmUnknownGame.SetNewGameName(ra::Widen(sEstimatedGameTitle));
+                    vmUnknownGame.SetEstimatedGameName(ra::Widen(sEstimatedGameTitle));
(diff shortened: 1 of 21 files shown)


meleu:
Hey @Jamiras , could you please open an issue on RAWeb's issue tracker listing the changes you have been doing in RAIntegration that need server changes. And then we can use the same thread to add new ones that may come.

I remember you added an argument with the md5 hash for `awardachievement` and now this `g` for `submitgametitle`... As I'm seeing RAIntegration evolving faster than RAWeb, I'm afraid we can lose track of all needed changes for RAWeb.

Jamiras:
You handled the MD5 one for me. I was planning to create one for this after it was merged. I think individual issues is better than one monster one as they can be resolved separately.

meleu:
> You handled the MD5 one for me.

wut me?!

IIRC in some of your PRs you added a `m` with the ROM's md5 when the emu sends an `awardachievement` to the server (so we can create statistics about compatibility, etc.). It demands not only change on the PHP code but also in the database design. I don't remember doing any of those (it's always itching in my mind, though).

Maybe you are mistakenly remembering what we did in relation to emulators (console ID x emulator ID), no?

Jamiras:
https://github.com/RetroAchievements/RAWeb/issues/96#issuecomment-455389527

PR #340, opened 2019-03-17
use rcheevos for Rich Presence processing

Jamiras:
Leverages https://github.com/RetroAchievements/rcheevos/pull/10, which still hasn't made its way to rcheevos/master, but we're already pointing at the develop branch to pick up bug fixes that were required for the 0.75 release.

As RetroArch doesn't currently support Rich Presence, there's no cross-dependency on making this change, and it gives us the chance to find any issues that may exist in the rcheevos implementation.

Also fixes an issue displaying ampersands in the Rich Presence Monitor.


Jamiras:
I've been using this code for over a month, and only found one issue with a malformed Rich Presence (https://github.com/RetroAchievements/rcheevos/pull/17). I'm going to merge this now and proceed with the changes in my submodule, expecting that PR to be merged (at least to develop, if not all the way to master) before we release 0.76.

PR #347, opened 2019-03-24
Show emulator-specific navigation buttons

Jamiras:
Default:
image

RAGens:
image

RAQuasi88/RAppleWin:
image


rzumer:
I like the result, but since the bindings are defined on the emulator side it would be best to also assign the label from there, maybe as a future improvement since the binds are not likely to change frequently.

PR #353, opened 2019-04-02
Remove /w45045 from test project configuration

rzumer:
I can see value in displaying warnings for code where Spectre mitigations can be inserted in the distributed integration code, but for tests it is irrelevant. This avoids displaying the C5045 informational warning by default for that project only.

### DIFF
--- a/tests/RA_Integration.Tests.vcxproj
+++ b/tests/RA_Integration.Tests.vcxproj
@@ -153,7 +153,6 @@
       <UseFullPaths>true</UseFullPaths>
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
       <DisableSpecificWarnings>4201</DisableSpecificWarnings>
-      <AdditionalOptions>/w45045 %(AdditionalOptions)</AdditionalOptions>
     </ClCompile>
     <Link>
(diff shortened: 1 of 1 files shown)


Jamiras:
Is this actually causing some problem for you? What prompted this change?

rzumer:
It's not much of a problem. I just found the warning confusing, and after looking into it, felt that it is better to leave this flag out of the test build, as there is no benefit to inserting Spectre mitigations there. The warning will still be displayed in the library build.

Jamiras:
I've finally installed VS2019 this week, and I don't see any additional warnings being generated, let alone any specifically due to this setting. 

In testing this change, I've removed the warning flag from the project, added the sample code [here](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5045?view=vs-2019), and tried building in release mode in both vs2017 and vs2019 and could not generate the warning. I also tried "/w35045" as mentioned in the link, and could not generate the warning.

The warning is currently only specified for the release build of the unit tests, whereas it's specified for both builds of the main project. At a minimum, I believe the warning should not be build-specific, and as it doesn't seem like it's actually doing anything for us at this time, I'm going to merge this to synchronize the flag across build versions.

PR #354, opened 2019-04-03
ViewModel/unit tests for Overlay

Jamiras:
* also adds ability to specify colors for the overlay elements (theming):
image
* eliminates the News, Messages, and Friends overlay pages. Messages was never implemented, News displayed raw HTML, and Friends will be restored later. I didn't want to make this PR any larger by including the friends list API changes.
* eliminates the exported `_RA_RenderPopups` and `_RA_UpdatePopups` methods. Popups are updated/rendered as part of the Overlay logic if the Overlay is not visible. The first two checks here will always be null in the existing emulator and the lines have been removed for future emulator builds: https://github.com/RetroAchievements/RAIntegration/blob/master/src/RA_Interface.cpp#L60-L74
* timestamp format changed from "Fri Mar 15 21:14:19 2019" to "Fri 15 Mar 2019 21:14:19"
* recent winner timestamps now include age
image


kdecks:
Awesome work and I think you made the right call here: 

>eliminates the News, Messages, and Friends overlay pages. Messages was never implemented, News displayed raw HTML, and Friends will be restored later. I didn't want to make this PR any larger by including the friends list API changes.

PR #357, opened 2019-04-06
Add banked RAM to the GB/C memory map

rzumer:
Note that the actual size is different depending on the system (GB/GBC). I prefer avoiding to create two distinct memory maps, but if there are issues caused by defining a memory region that can be larger than the emulator's map, I can change it.

### DIFF
--- a/src/data/ConsoleContext.cpp
+++ b/src/data/ConsoleContext.cpp
@@ -122,6 +122,7 @@ const std::vector<ConsoleContext::MemoryRegion> GameBoyConsoleContext::m_vMemory
     { 0xFF00U, 0xFF7FU, ConsoleContext::AddressType::HardwareController, "Hardware I/O"},
     { 0xFF80U, 0xFFFEU, ConsoleContext::AddressType::SystemRAM, "Quick RAM"},
     { 0xFFFFU, 0xFFFFU, ConsoleContext::AddressType::HardwareController, "Interrupt enable"},
+    { 0x10000U, 0x16FFFU, ConsoleContext::AddressType::SystemRAM, "System RAM (static pages)" }, // 7 banks, GBC only
 };
 
 // ===== GameBoy Advance =====


src/data/ConsoleContext.cpp
```diff
@@ -109,7 +109,7 @@ const std::vector<ConsoleContext::MemoryRegion> GameBoyConsoleContext::m_vMemory
     { 0x0000U, 0x00FFU, ConsoleContext::AddressType::HardwareController, "Interrupt vector" },
     { 0x0100U, 0x014FU, ConsoleContext::AddressType::ReadOnlyMemory, "Cartridge header" },
     { 0x0150U, 0x3FFFU, ConsoleContext::AddressType::ReadOnlyMemory, "Cartridge ROM (fixed)" }, // bank 0
-    { 0x4000U, 0x7FFFU, ConsoleContext::AddressType::ReadOnlyMemory, "Cartridge ROM (paged)" }, // bank 1-XX (switchable)
+    { 0x4000U, 0x7FFFU, ConsoleContext::AddressType::ReadOnlyMemory, "Cartridge ROM (paged)" }, // bank 1-7 for GBC (switchable)
```
Jamiras: There can be a lot more than 7 ROM banks: http://gbdev.gg8.se/wiki/articles/Memory_Bank_Controllers#4000-7FFF_-_ROM_Bank_00-1FF_.28Read_Only.29

rzumer: Misread ROM as RAM, will revert.

Jamiras:
With the current interface, there's no way to access this new entry as the memory mapped from the emulator shouldn't extend into it for non-GBC mode, and the search for "System Memory" will stop after the $C000-$DFFF range. This means there won't be a quick filter for the banked memory, but the user can still create a custom filter.

As such, I'm fine with sharing the map. If we need to diverge later, we can worry about duplicating things then.

PR #358, opened 2019-04-06
switch to three-line popups

Jamiras:
Default is classic orange:
image
image
NOTE: leaderboard description is shown on the third line, this leaderboard doesn't have a description.

Theming allows for custom colors. I propose changing the default to grey:
image
image

More Popups (in grey):
image
image
image


rzumer:
LGTM, and agree with defaulting to grey.

kdecks:
My preference is first to (bordered) Colored Grey as seem in the pop-up designer. Second would be Grey, as show here.

@Jamiras I'd suggest giving a few options in a community poll using the #poll-me channel, for what is used as the default color scheme.

PR #376, opened 2019-05-04
ensure player score appears in scoreboard

Jamiras:
The `submitlbentry` returns the player's best score and rank, so while it's not currently possible to show the [neighboring scores](https://github.com/RetroAchievements/RAWeb/issues/281), it is possible to show the player's rank/score:
image

Additionally, if the player's score does not improve, we can display the submitted score along with the player's best:
image


kdecks:
I've been looking forward to this for a long time. Awesome work!

PR #400, opened 2019-06-21
Remove limit on amount of conditions per group in the achievement editor

### LINKED ISSUE #399: Limited number of conditions per group in achievement editor
I was trying to develop an achievement for Super Mario 64. For this achievement to work I had to check the state of a game object whose position in memory is not fixed. Luckily, however, the position is not random, it can only be one of 240 positions in an array.

Given that knowledge, I thought I could just check all 240 positions. Each individual check required 3 conditions, causing the achievement to have over 700 conditions in a single alt group. This would have been quite a pain to make in the achievement editor (and very error prone), so I generated the achievement string and put it in the XXXXX-User.txt file that contains local achievements.

The achievement worked and there was no noticeable lag, so the engine handled it fine. The issue only showed up when I tried to look at the achievement in the editor, as that caused the emulator to crash. I looked at the code and found out that the editor limits the maximum amount of conditions per group to 200. It doesn't allow the creation of more conditions when the limit is reached, so the crash is not a problem, it was caused by me editing the achievement through the file.

Having now explained my motivation for having an achievement with "a lot" of conditions, and established that it seemed to work fine, my question is: why does the achievement editor impose the 200 limit per group? Is it a purposeful limit to keep achievements small, or is it just a side-effect of how the editor's code was written (it uses a fixed std::array instead of a dynamic container)? If it is a side-effect, would it be ok to remove the limit, or raise it quite a bit?

gdeOo:
The amount of conditions per group was being artificially limited by the
achievement editor, capped at 200. The limit has been lifted by replacing the std::array-based implementation with an std::vector-based one. There should be
no noticeable/measurable performance impact, since the vector is shared by all
uses of the achievement editor, making any allocation-related cost negligible.

Also removed some ad-hoc initialization of the std::strings in the container
that was a workaround for the fact that a preceeding ZeroMemory call messed up
their post-constructor state.

Closes #399.

### DIFF
--- a/src/RA_Dlg_AchEditor.cpp
+++ b/src/RA_Dlg_AchEditor.cpp
@@ -114,9 +114,7 @@ void Dlg_AchievementEditor::SetupColumns(HWND hList)
         i++;
     }
 
-    ZeroMemory(&m_lbxData, sizeof(m_lbxData));
-
-    m_nNumOccupiedRows = 0;
+    m_lbxData.clear();
 
     BOOL bSuccess = ListView_SetExtendedListViewStyle(hList, LVS_EX_FULLROWSELECT);
     bSuccess = ListView_EnableGroupView(hList, FALSE);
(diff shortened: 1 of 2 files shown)


gdeOo:
Yes, I tested it manually, including with the larger achievement I mentioned in the issue. Went through a bunch of achievements and their groups. No problems showed up.

SyrianBallaS:
Alright, I don't have merge authority so some else will have to look at this (@jamiras). Dont know why contributors cant request reviewers here.

PR #408, opened 2019-07-03
fix analysis warnings

Jamiras:
Apparently my `analysis` build has been broken for some time. Despite not reporting any errors they apparently exist as I noticed after checking out a fresh clone for starting 0.77 work. This PR addresses all the current warnings.


Jamiras:
It appears that the analysis warnings aren't being generated (or at least reported) in VS2019. I accidentally opened the new clone in VS2017.

It seems Code Analysis is silently _**disabled**_ when loading a VS2017 project in VS2019. Changing the Platform Toolset to "Visual Studio 2019 (v142)" does make the analysis warnings show up in VS2019, but then the project won't load in VS2017.

Note that that `analysis` build currently targets v141, whereas the `debug` and `release` builds target v141_xp. Attempting to build `analysis` with v141_xp in both VS2017 and VS2019 results in this error:
```error MSB8026: Static analysis is not supported with the current platform toolset.
4>Done building project "RA_Integration.vcxproj" -- FAILED.```

As this only affects the `analysis` build, I'm going to bump the Platform Toolset to v142. This means that the `analysis` build will only work in VS2019. The `debug` and `release` builds will still target `v141_xp` and build in both VS2017 and VS2019.

Jamiras:
I've found a way to make the project select an appropriate toolset in both vs2017 and vs2019. Instead of fixing the version at v141 or v142, I've set it to "$(DefaultPlatformToolset)" which is v141 in vs2017 and v142 in vs2019. Both versions of Visual Studio reported a test warning.

PR #448, opened 2019-09-17
framework support for taking screenshots on achievement trigger

Jamiras:
Currently requires manually editing the prefs file to enable. A future PR will will allow the user to enable the functionality and specify a location where the screenshots should be saved. Screenshots will not be taken for local unlocks or unlocks for modified achievements.

The screenshot captures the _client area_ of the window and the overlay. It does not capture the window frame/titlebar. It also currently uses the achievement ID as the filename. I want to change that to use the achievement title. However, there may be issues with validating the characters in the title, or uniqueness between games.

### DIFF
--- a/src/Exports.cpp
+++ b/src/Exports.cpp
@@ -391,6 +391,9 @@ API void CCONV _RA_DoAchievementsFrame()
     auto& vmBookmarks = ra::services::ServiceLocator::GetMutable<ra::ui::viewmodels::WindowManager>().MemoryBookmarks;
     vmBookmarks.DoFrame();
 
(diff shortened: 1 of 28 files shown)


luchaos:
Very nice! Would prefixing them with the game id either in path or in the filename work?

meleu:
Cool!

Just FYI, [in my PR with a similar feature for RetroArch](https://github.com/libretro/RetroArch/pull/6388), the screenshot filename is `RomName-cheevo-CheevoID.png`. Example: `Metroid (USA)-cheevo-3507.png`.

PR #475, opened 2019-11-03
Hook up the UI for the Measured flag

Jamiras:
Active achievements with a Measured field will be reflected in the overlay:

image

Inactive (previously triggered) achievements with a Measured flag will not show a progress bar. Achievements without a Measured flag will not show a progress bar. 

Only one Measured field may be present in an achievement (this includes across alts). If more than one is present, a "Parse Error -22" will occur.


Jamiras:
Implements #4 - though the only progress indicator is in the achievement list. There aren't (at least not yet) periodic notifications as the user makes progress, or a persistent indicator for whichever progression is the closest to complete.

Issue #60, opened 2018-06-18
Indirect memory addresses: pointers, etc


Jamiras:
As a developer, I want to be able able to reference memory that may not always be at the same address.

To simplify the explanation, I use the following conventions:
* `$` is a memory reference
  * `$1234` means the memory at address `0x1234`. 
* `0x` is a constant
* both are hex encoded
* `()` indicates order of operations: 
  * `$(0x5678 + $1234)` means: take the memory at address `0x1234`, add `0x5678`, and get the memory at the resulting address.

There are four scenarios to consider:
* **Direct pointer**: A direct pointer is a memory address that points at another memory address of interest.
  * `$($1234)`
  * `$1234` contains `0x5678`, condition should look at `$5678`.
* **Indirect pointer**: An indirect pointer is a memory address that points at a chunk of memory where the memory of interest is always N bytes from the pointed-to memory. N is typically the offset of a field within a structure, but it could also be a constant index into an array.
  * `$($1234 + 12)`
  * `$1234` contains `0x5678`, condition should look at `$(0x5678 + 12)`, i.e.`$5684`
* **Array index**: The memory address specifies the number of bytes to offset a known address by.
  * `$(0x5678 + $1234)`, alternately notated as `$(0x5678[$1234])`
  * `$1234` contains `6`, condition should look at `$(0x5678 + 6)`, i.e. `$567E`
* **Scaled array index**: The memory address specifies a number of units to offset a known address by. A unit is some number of bytes that may represent a simple data type (2 for 16-bit values, 4 for 32-bit values) or a complex data type (i.e. 48 for a character sheet in an RPG)
  * `$(0x5678 + $1234 * 16)`, alternately notated as `$(0x5678[$1234*16])`
  * `$1234` contains `6`, condition should look at `$(0x5678 + 6 * 16)`, i.e. `$56D8`

In all of these scenarios, a known memory location (`$1234`) can be used to determine a dynamic address that the condition depends on. All of the scenarios also fit into the formula: "read memory at address, multiply by scalar, add constant, and read memory at result":

* **Direct pointer**: `$($1234 * 1 + 0x0000)`
* **Indirect pointer**: `$($1234 * 1 + 0x000C)`
* **Array index**: `$($1234 * 1 + 0x5678)`
* **Scaled array index**: `$($1234 * 16 + 0x5678)`

As such, the proposal below attempts to fit that formula into the existing framework with the smallest amount of overhead.

[...]

PR #480, opened 2019-11-10
add Add Address condition type

Jamiras:
A simple proof-of-concept example:
image

The 32-bit value at 0x0112F8 is a pointer to the data for the character in the first position of the party. Changing the party order simply shuffles the pointers in the array at 0x0112F8. 

Note that the 32-bit value is actually a real address in the emulated system's memory, and the memory inspector's view of memory doesn't cover the entire memory space. By only reading a 24-bit value, we get an address in the memory inspector's memory space. Then we add a constant (0x5C) to adjust the value to the actual address containing the memory we're interested in. At that address, we read an 8-bit value and compare it to 0x15. If they're equal, the condition is true.

In a bit more detail:
* 32-bit value at 0x112F8 is 0x800110B8
* 24-bit value at 0x112F8 is 0x0110B8
* The data we're interested in is at 0x011114 in the memory inspector.
* 0x011114 is 0x0110B8 + 0x00005C
* By putting 0x00005C as the address in the second condition, the result of the first condition will be added to 0x00005C before doing the memory read for the second condition.
* if 8-bit value at 0x011114 is 0x15, then the condition is true

The same logic can be applied to leaderboards and rich presence by using the "Measured" flag:
```
I:0xW0112F8_M:0xH00005C
```
```
Level @Level(I:0xW0112F8_M:0xH00005C)
```
`I` is the flag for AddAddress, `M` is the flag for measured. `W` is a 24-bit read, and `H` is an 8-bit read.
By using Measured, we can leverage flags normally not available to values, such as AddAddress, AddSource, and AddHits. Without a HitCount or comparison, the Measured value is the resulting left-hand value for the final condition.


Jamiras:
This partially implements #60. The multiplication functionality is not yet available, so "Scaled Array Index" is not currently possible. 

Also note that this solution adds the address to both sides. This allows for delta support, but makes comparing a pointed-at memory value to a non-pointed at memory value virtually impossibly. As I can't think of a single situation where I've compared two separate memory addresses, I don't think this limitation is going to be problematic. Not having delta support would be much worse.

PR #489, opened 2019-11-29
display dummy background if overlay background is not available

Jamiras:
Fixes a rendering issue when the `overlay` directory is missing.

Before:
image

After:
image


rzumer:
Nice catch.

PR #491, opened 2019-12-01
include playtime in mastery notification

Jamiras:
Adds a third line to the mastery notification that includes the player name and captured play time.

The play time value matches the value shown in the achievements overlay and reflects the amount of time the player has played the game in the current emulator (times are stored locally in the RACache directory).

image

Also moved the `UpdateRenderImage` call out of the `IsAnimationStarted` check to avoid a race condition where another thread was between starting the animation and rendering the image. In that situation, the image may be rendered a second time, instead of crashing the screenshot thread.

### DIFF
--- a/src/data/GameContext.cpp
+++ b/src/data/GameContext.cpp
@@ -17,6 +17,7 @@
 #include "api\UpdateCodeNote.hh"
 
 #include "data\EmulatorContext.hh"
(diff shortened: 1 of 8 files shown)


Jamiras:
The mastery popup now has a custom themeable color. The default is gold-ish:
image

PR #505, opened 2019-12-19
force memory viewer to repaint in Oric emulator

Jamiras:
Fixes an issue where the memory viewer window only updates when moving the mouse over the main emulator window. 

For reasons that are completely unclear, the `InvalidateRect` calls for the memory viewer and search results are only promoted to a `WM_PAINT` event when other events are in the queue (i.e. `WM_MOUSEMOVE`). This behavior appears to be SDL-specific, and I can't figure out why. In the past, we've explicitly forced a repaint of these controls for RALibRetro to work around the issue. I've extended that behavior for RAOricutron.

Without this change, there are several oddities in the tool windows. The most obvious one is the memory viewer doesn't update every frame. Other symptoms include radio buttons not appearing to be selected or deselected (including the radio buttons on the achievement sets dialog), and focus rectangles not appearing on buttons.

### DIFF
--- a/src/RA_Dlg_Memory.cpp
+++ b/src/RA_Dlg_Memory.cpp
@@ -270,12 +270,17 @@ void MemoryViewerControl::Invalidate()
     {
         InvalidateRect(hOurDlg, nullptr, FALSE);
 
-        // In RALibRetro, s_MemoryDrawProc doesn't seem to be getting trigger by the InvalidateRect, so explicitly force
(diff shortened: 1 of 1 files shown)


rzumer:
I used to force push an SDL event in the emulator that would redraw the window every frame, but removed it later because I forgot that it affected the memory viewer and thought it had been solved by another change. Better address it here anyway.

PR #533, opened 2020-04-12
have rich presence monitor watch for game and visibility changes

Jamiras:
eliminates some hard-coded calls to refresh the rich presence monitor with an event-driven mechanism.


Keltron3030:
Cool

Issue #107, opened 2018-09-15
Continuous Filter for Memory Inspector


kdecks:
The concept is to create a toggle switch button, alongside the ``filter!`` button that would filter every frame and check your comparison until it's turned off again.

Normally when I'm filtering, I will find myself pressing the filter button sometimes dozens of times before I change the conditions of my filter. This switch would be is a big time saver for achievement devs and make for more accurate results by clearing out addresses that don't fit in the filter for one frame and could otherwise easily be missed.

Already talked to @Jamiras about it briefly, he noted that would probably be necessary to turn off the filter history while this is taking place.

PR #554, opened 2020-04-26
add continuous filter

Jamiras:
Implements #107

Press once to start filtering, press again to stop filtering. While filtering, the current filter will automatically be applied every frame. 

The continuous filter results are applied to the same record in the search history. For Last Value comparisons, the filter _will_ look at the value from the previous frame, but after the next frame that data is no longer available. Going back one step in the history (<< button) will go back to the state before starting the continuous filter.

Note: this has significant overhead when there are more than a couple thousand results. To account for this, if the number of results does not drop to a reasonable level within 60 frames, it will automatically be stopped and the user will be warned to try again with fewer results.

### DIFF
--- a/src/ui/viewmodels/MemorySearchViewModel.cpp
+++ b/src/ui/viewmodels/MemorySearchViewModel.cpp
@@ -3,6 +3,7 @@
 #include "data\ConsoleContext.hh"
 #include "data\EmulatorContext.hh"
 
+#include "services\IClock.hh"
 #include "services\ServiceLocator.hh"
 
 #include "ui\viewmodels\MessageBoxViewModel.hh"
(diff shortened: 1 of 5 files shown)


Keltron3030:
Would it be practical if the continuous filtering rate could change dynamically depending on total addresses, for example maybe above 10,000 results the filter rate is once per 10 frames or higher if needed to restrict demand.

Jamiras:
A reasonable idea. What I've implemented is that N ms must elapse between filterings where N = number of remaining results / 100. So, if there are 10,000 results left, it would only apply the filter if at least 100ms elapsed (10 times per second). If there are 50,000 results left, it would only apply the filter twice per second, etc.

Keltron3030:
Excellent

PR #694, opened 2021-03-27
show tooltip for addresses in ranged code notes

Jamiras:
When an address doesn't have an explicit code note but does fall in the range of another address's code note annotations, the tooltip will now show the containing range (similar to search results).

Here, the address is $c617, which does not have a code note, but $c559 has a code note indicating it covers the 285 bytes between $c559 and $c7dd. Since $c617 falls in that range, that tooltip is displayed, with an annotation that the actual address is 191 bytes into the annotated range.
image

Right+clicking on the address will still take you directly to the address.

### DIFF
--- a/src/data/context/GameContext.cpp
+++ b/src/data/context/GameContext.cpp
@@ -896,6 +896,27 @@ void GameContext::OnCodeNoteChanged(ra::ByteAddress nAddress, const std::wstring
     }
 }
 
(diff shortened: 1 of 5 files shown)


Jamiras:
I've changed my mind on this. Instead of displaying the address of the base note and the index within the size, I think it's cleaner to show the address as-is, and the base address and offset in the annotation.

Instead of "`0x00c559 [191/285]`", the first line of the tooltip now says "`0x00c617 [0x00c559+190]`"

PR #902, opened 2022-06-27
support for indirect code notes

Jamiras:
If the first line of a code note contains the keyword `pointer`, and at least one other line in the code note starts with a plus sign, then every line that starts with a plus sign will be interpreted as an offset. Both decimal and hexadecimal offsets are supported, but hexadecimal offsets must start with `0x`. As long as all offsets parse correctly, virtual code notes will be created using the pointer value and offsets.

image

Here, the first line of the code note indicates that the value is a 24-bit pointer. Note: the size is important for applying the offset. If the full 32-bit pointer is noted, then the offsets will have to be relative to the entire pointer. **This logic does not support 25-bit pointers or multiple levels of pointers.**

The second line says there's a 16-bit value at an offset of 0x1294 from the pointer value. The tooltip shows the subset of the parent code note associated to the offset. Similarly, the code note field in the memory inspector shows the subset of the parent code note, and an hint indicating where the parent code note is. When a subset is displayed, the field is not editable.

Additionally, the virtual code notes will show up in search results:
image

When the pointer changes, the virtual code notes will be relocated, but those changes may not immediately be reflected in the UI.

Hexadecimal offsets that don't start with `0x`, or offsets that aren't at the start of the line or prefixed with a `+` will not be handled and virtual code notes won't be generated from the parent code note. I only expect about 1/3 of existing pointer code notes will be automatically handled with the initial rollout. If desired, they can be manually edited to conform to the dictated design.


pinguupinguu:
thank gosh for this. been waiting so long for this!!!!

redwizard42:
Is the | character needed for the offset label? Or will something like +0xDF4 Monster ID be sufficient?

Jamiras:
> Is the | character needed for the offset label? Or will something like +0xDF4 Monster ID be sufficient?

The pattern it's looking for is:
* A plus (`+`) at the start of the line
* Immediately followed by a decimal or hexadecimal offset (hexadecimal number has `0x` prefix)
* zero or more whitespace characters
* (optional) any single non-alphanumeric character followed by zero or more whitespace characters.

All of these should be equivalent:
```
+0x12 Monster ID
+18 Monster ID
+18=Monster ID
+18: Monster ID
+18 | Monster ID
+18 - Monster ID
+18>     Monster ID
+0x12 ~ Monster ID
```

All text until the next line starting with a plus will be considered part of the virtual note, so you can also do things like:
```
+18 - Monster ID
6 = Orc
7 = Imp
8 = Goblin
+20 - Quantity
A value of 0xFF indicates this slot is not used
````

The example note above was just something pre-existing that I was using to test with.

redwizard42:
> The example note above was just something pre-existing that I was using to test with.

Awesome. That sounds perfect.

PR #905, opened 2022-08-26
switch points field to dropdown

Jamiras:
Matches changes made in https://github.com/RetroAchievements/RAWeb/pull/1054#issuecomment-1207488057

image

If a value outside of the allowed range is provided (legacy support), the achievement will be flagged:

image


src/data/models/AchievementModel.cpp
```diff
@@ -20,6 +20,8 @@ const BoolModelProperty AchievementModel::PauseOnTriggerProperty("AchievementMod
 const IntModelProperty AchievementModel::TriggerProperty("AchievementModel", "Trigger", 0);
 const StringModelProperty AchievementModel::UnlockRichPresenceProperty("AchievementModel", "UnlockRichPresence", L"");
 
+const std::array<int, 10> AchievementModel::s_vValidPoints = {0, 1, 2, 3, 4, 5, 10, 25, 50, 100};
```
luchaos, line 23: When writing the comment for RAWeb i kept thinking; would it make sense to have some kind of configuration response in one of the earliest bootstrap RPC calls where emulator/core/system/console-specific options could be provided to keep the site, db, and emulators in sync?
E.g. one console might allow for the full list of points, another just a subset because of the very different nature? I'm not sure this is a good example for such configuration options; yet, it's came up several times in the past thinking about how to extend the connect/integration API's feature set.

Jamiras: By leaving the validation as a warning, the user can still submit the incorrect values to the server in case that restriction is changed. While the dropdown does prevent them from entering custom values, they can still manually put other values in the XXXX-User.txt file. The server has the final authority on what values are accepted, and was recently changed to error if anything other than approved values are sent.

I think the list of viable points isn't volatile enough to warrant synchronizing every time the client connects. If we decide to change the validation on the server, then we can time it to coincide with a new client.

However, we do already have an API for returning the valid badge range. It's also used for validation in the achievement editor - specifically for when the user manually enters a badge number. That's only fetched once per session, which could result in a potential error if trying to use a new badge uploaded from another client, but I doubt many people ever do enter reuse already uploaded badges. If the points list is an actual concern, we could extend that API to also include the valid point values.

The one configuration thing that I have really considered moving to the server is the cores.json file for RAlibretro. Most RAlibretro releases occur when we roll out a new system so players can get the updated cores.json file. The last rollout for Fairchild Channel F didn't have any corresponding binary changes, so I chose not to release a new version of RAlibretro. Players have to manually edit the file if they want to play Channel F games in RAlibretro as I didn't want to bump the binary version just for a one line change in an external file. But I haven't taken any action on it as it's not something that would happen frequently (historically there has always been some binary changes to publish), RAlibretro is promoted as a developer tool (not a player tool), and we need some way to allow developers to manually enable other systems for testing.

luchaos: > we can time it to coincide with a new client
That's fair. Underlines how the valid points values might not have been the best example 😊 

The valid badge range is a bit of a weak point atm but as you said, rarely an issue. Still, I'd like to revisit that at some point if possible. Points and badge iterator are certainly not a pressing issue.

Providing the cores list however sounds _very_ intriguing. I think it'd fit well into the integration and emulator release management system which could make your life as release manager of those much easier. Fetching the list of cores from libretro is easy enough, providing an interface where we can choose which ones to provide to RAlibretro and to which quality just the same. I'd love to be part of building something like that.

PR #920, opened 2022-10-01
Add warning when serialized size exceeds database limit

Jamiras:
image


luchaos:
Should this mention the `database` constraint, be a local `length` constraint instead?
It'S ultimately a database restriction but I'm not sure how relevant that information is on the client side for creators.

Jamiras:
I've removed 'database' from the messages. It now says "Serialized length exceeds limit"

PR #975, opened 2023-02-04
change wording on modified achievement unlock popup

Jamiras:
implements #968 

Local achievements just say "Local Achievement Unlocked" or "Modified Local Achievement Unlocked".
image

Modified published achievements say "Modified Achievement Unlocked LOCALLY".
image

### DIFF
--- a/src/data/context/GameContext.cpp
+++ b/src/data/context/GameContext.cpp
@@ -516,7 +516,8 @@ void GameContext::AwardAchievement(ra::AchievementID nAchievementId)
     {
         auto sHeader = vmPopup->GetTitle();
         sHeader.insert(0, L"Modified ");
-        sHeader.insert(sHeader.length() - 8, L"NOT ");
+        if (pAchievement->GetCategory() != ra::data::models::AssetCategory::Local)
+            sHeader.append(L" LOCALLY");
         vmPopup->SetTitle(sHeader);
 
         RA_LOG_INFO("Achievement %u not unlocked - %s", pAchievement->GetID(), pAchievement->IsModified() ? "modified" : "unpublished");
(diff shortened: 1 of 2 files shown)


src/data/context/GameContext.cpp
```diff
@@ -528,7 +529,7 @@ void GameContext::AwardAchievement(ra::AchievementID nAchievementId)
     const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
     if (bSubmit && pEmulatorContext.WasMemoryModified())
     {
-        vmPopup->SetTitle(L"Achievement NOT Unlocked");
+        vmPopup->SetTitle(L"Achievement Unlocked LOCALLY");
```
luchaos, line 532: Just a thought, as we tried to reduce ascii art and emphasis via typography in RAWeb before: would '(Local)' be enough? It's the same character count.
FAILED is trickier, i guess putting it at the beginning of the line as "Failed:" would convey the message just as well?

Jamiras: It's important that the player knows the affected unlocks did not occur on the server. The "LOCALLY" for things not sent to the server, and the "FAILED" for things that did get sent but returned an error code. I believe these messages should stand out from standard messages so the user isn't confused why the unlock didn't occur. The messages previously said "Modified Achievement NOT Unlocked" and "Achievement NOT Unlocked" (failure).

I understand the undesirability of using all caps for emphasis, but putting something in parentheses implies subtext and may not actually catch the user's attention. These messages are only on the screen for a few seconds. I'd be fine with just making them Title Case, but I'm not convinced that will be noticed by the player.

Ideally, the player should never see these messages. Modified achievements are usually devs testing a fix. Failed unlocks are something more catastrophic as communication errors are automatically retried and don't generate a failed popup.

luchaos: I like that reasoning. In any case better than the previously mixed in `NOT` 👍

redwizard42: This is definitely an improvement. Had another Jr dev question if the existing message meant it triggered appropriately earlier today. I think this will clear up doubts. I agree with the use of all caps on the modified case. Makes the important part noticeable.

PR #1000, opened 2023-05-20
Feature/add pointer finder to open all

germanamoros:
Added Pointer Finder window to Open All.

### DIFF
--- a/src/ui/viewmodels/IntegrationMenuViewModel.cpp
+++ b/src/ui/viewmodels/IntegrationMenuViewModel.cpp
@@ -351,6 +351,7 @@ void IntegrationMenuViewModel::ShowAllEditors()
         pWindowManager.MemoryInspector.Show();
         pWindowManager.MemoryBookmarks.Show();
         pWindowManager.CodeNotes.Show();
+        pWindowManager.PointerFinder.Show()
     }
     else
     {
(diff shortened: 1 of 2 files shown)


Jamiras:
I disagree with this change.

While technically "Open All" sounds like it should open all windows, the Pointer Finder is something that has no use for the majority of the systems that we support and limited use in others.

Once a developer has found their pointer(s), they should be able to create and test their achievements without ever needing it open again. And if they do, it's readily accessible from the menu.

Automatically opening it so it can just sit there empty most of the time is a waste of screen space. Assuming the user's not going to immediately be using it, they'll likely close it - which is extra effort for the user, and contrary to the convenience intended by having the Open All menu item.

germanamoros:
Understandable.

Yet still I think something should be done with the Pointer Finder option, as it usually an windows that jr-devs tend to miss out at the begining, maybe moving it to it's own section? (shitty mockup attached)

image

That way, it would be outside the scope of Open All, plus it would be visible as something outside the usual dev tools.

What do you think about it?

Jamiras:
I think moving it to a separate section is fine.

I find your argument that jr. dev's "miss out" on the Pointer Finder superfluous. Even if a jr. dev saw the link, it's unlikely they'd know how to use it. Pointers are something that many non-jr. dev's still have troubles understanding.

PR #1009, opened 2023-07-03
add 32-bit BE memory viewer size

Jamiras:
image

vs

image


Alexdatadestroyer:
Oh, this should help a lot for ASCII cases: when letters are informed backwards!

Jamiras:
> Oh, this should help a lot for ASCII cases: when letters are informed backwards!

Not really sure how. It's the exact same order as 8-bit view. It just removes the spaces so the values are clustered four bytes at a time.

Alexdatadestroyer:
> > Oh, this should help a lot for ASCII cases: when letters are informed backwards!
> 
> Not really sure how. It's the exact same order as 8-bit view. It just removes the spaces so the values are clustered four bytes at a time.

Yep, I noticed that. Well... I should had give an example in order to sustain my argumentation...:

I said that based on an example that I saw months ago on [Half-Life 1 (PS2) set page](https://retroachievements.org/game/3058) whereas the Developer [SubliminalSiren](https://retroachievements.org/user/SubliminalSiren) used ASCII Text as Stage ID. All HL 1 stages have same text code when compared to PC version.

[Extracted from the list on this post on Steam](https://steamcommunity.com/app/70/discussions/0/540742667840503958/) and corrected below:

> BLACK MESA INBOUND: map c0a0
> ANOMALOUS MATERIALS: map c1a0
> UNFORSEEN CONSEQUENCES: map c1a1
> OFFICE COMPLEX: map c1a2
> "WE'VE GOT HOSTILES!" map c1a3
> BLAST PIT: map c1a4
> POWER UP: map c2a1
> ON A RAIL: map c2a2
> APPREHENSION: map c2a3
> RESIDUE PROCESSING: map c2a4
> QUESTIONABLE ETHICS: map c2a4d
> SURFACE TENSION: map c2a5
> "FORGET ABOUT FREEMAN": map c3a1
> LAMBDA COMPLEX map c3a2
> XEN: map c4a1
> GONARCH'S LAIR: map c4a2
> INTERLOPER: map c4a2b (Also the ending of Gonarch's Lair)
> NIHILANTH: map c4a3
> ENDING: map c5a1

Looking on some achievements that use Stage ID address, the text used to identify the stage is in backwards. For example, looking on [Office Complex](https://retroachievements.org/achievement/308482) achievement code, you can notice values 0x32613163 and 0x33613163 which identifies, respectively, Office Complex and We've got hostiles stages as well. However, in text format is 2a1c and 3a1c.

With 32-bit Big Endian view, those values can be successfully represented as 0x63316132 and 0x63316133, giving a correct identification of the text as Stage ID.

I didn't see any other relative case yet, but, probably, there are more Playstation 2, Dreamcast or other console games that uses text as Stage ID representation which the value is given backwards due to bytes being inverted as result of 32-bit view on Memory Inspector.

Nowadays, I am trying to create a set for [Unreal Tournament 99 (PS2)](https://retroachievements.org/game/21560) which also has a PC version. So, probably, the Stage ID might be presented as a form of inverted text. I couldn't find it correctly at first glance at time - which rendered me to try other options to identify Stages and Tournament Ladders. However, having knowledge of this HL 1 case now, I will try to check for it, but on inverted text format this time!

Jamiras:
> With 32-bit Big Endian view, those values can be successfully represented as 0x63316132 and 0x63316133, giving a correct identification of the text as Stage ID.

Bit 8-bit view, those values can be successfully represented at 0x63 0x31 0x61 0x32 and 0x63 0x31 0x61 0x33. It sounds like you're referring to the 32-bit BE memory size, which already exists.

PR #1039, opened 2023-11-07
Destroy overlay on new window instead of updating parent

Riku55:
Using RAIntegration with Dolphin, the issue came up that when changing the overlay from the menu window to the game window, using SetParent wasn't working properly, causing the overlay to be relative to the screen and not the new window. Destroying the overlay and then creating a new one seems to work fine, at least for my tests I did with Dolphin and RALibretro.

I also noticed that if the overlay gets destroyed while a popup is still shown, the new overlay would not work, so I added 2 lines to clear all popups before destroying the window.

### DIFF
--- a/src/ui/win32/OverlayWindow.cpp
+++ b/src/ui/win32/OverlayWindow.cpp
@@ -57,14 +57,15 @@ void OverlayWindow::CreateOverlayWindow(HWND hWnd)
 
     if (m_hOverlayWnd && IsWindow(m_hOverlayWnd))
     {
-        // overlay window exists, just update the parent (if necessary)
+        // overlay window already exists, destroy before creating a new one
         if (m_hWnd != hWnd)
         {
-            m_hWnd = hWnd;
-            SetParent(m_hOverlayWnd, hWnd);
-            UpdateOverlayPosition();
+            DestroyOverlayWindow();
         }
-        return;
+        else
+        {
+            return;
+        }
     }
 
     // overlay window does not exist, create it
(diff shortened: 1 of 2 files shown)


CasualPokePlayer:
To be clear, [SetParent](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setparent) is actually working "correctly," the reason it seemed to break is due to the fact that it was changing the parent window from the [desktop window](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdesktopwindow) to the new window passed.

This is of course slightly confusing, as initial overlay window creation does pass the emulator window hwnd over to hWndParent in [CreateWindowEx](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createwindowexw). However, this doesn't actually set the parent window, rather it only sets the owner window, as the overlay window is not a child window, but rather a popup window (which has its parent be the desktop window normally).

The owner window is a separate field from the parent window, and notably [you *cannot* transfer ownership of a window](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#owned-windows) to another window. This would mean that, if the client were to call `RA_UpdateHWnd` with a new window, then destroy the old window, it would just then proceed to destroy the overlay window right after that old window is destroyed! Thus the only sane way to handle `RA_UpdateHWnd` with a new window while still having the overlay window being a popup window would be to destroy the overlay window and create a new one, as this PR does.

The parent change ultimately breaks things as seen in Dolphin more due to the behavior of [MoveWindow](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-movewindow). Per its documentation:
> For a top-level window, the position and dimensions are relative to the upper-left corner of the screen. For a child window, they are relative to the upper-left corner of the parent window's client area.

This is technically right under normal conditions, but also hides against an implementation detail: `MoveWindow` actually is always relative to the parent window. It would just happen that the desktop window is the parent window for top-level windows, normally. By calling `SetParent`, this assumption the documentation makes is broken, and so `MoveWindow` operate more like it does with a typical child window.

Jamiras (approved):
I've verified that this fixes the popups when running Dolphin in Windowed mode, but they don't show up in fullscreen. I've also verified that behavior is not related to this change, and also occurs in RAlibretro. It does not occur in RAVBA or RAMeka.

The easiest way to test is to create a local achievement that triggers every 10 seconds or so (1=1 300x), activate it, and set Keep Active to true.

Riku55:
I also tested the popups in full screen both in Dolphin and RAlibretro. I noticed that they do work in full screen in Dolphin if using any graphics API other than OpenGL.

PR #1051, opened 2024-01-14
add rarity to achievement popups

Jamiras:
If rarity is less than 10%, the popup is modified to say "Rare Achievement Unlocked", and an alternate sound can be played. If the alternate sound (`rareunlock.wav`) doesn't exist, the standard `unlock.wav` will be played instead.

In this video, a rare achievement is unlocked first (whistling sound), then a normal achievement (boop).

https://github.com/RetroAchievements/RAIntegration/assets/32680403/2629db64-589c-45ac-8ed2-0ca426721928

Hardcore rarity only include hardcore unlocks. Non-hardcore rarity includes all unlocks. If an achievement has 7% hardcore unlocks and 15% total unlocks, it would be reported as a rare achievement for a hardcore player, and a normal achievement for a non-hardcore player.

### DIFF
--- a/src/services/AchievementRuntime.cpp
+++ b/src/services/AchievementRuntime.cpp
@@ -1411,8 +1411,38 @@ static void HandleAchievementTriggeredEvent(const rc_client_achievement_t& pAchi
         Expects(!CanSubmitAchievementUnlock(pAchievement.id, nullptr));
 #endif
 
-    ra::services::ServiceLocator::Get<ra::services::IAudioSystem>().PlayAudioFile(bIsError ? L"Overlay\\acherror.wav"
-                                                                                           : L"Overlay\\unlock.wav");
+    const wchar_t* sAudioPath = L"Overlay\\unlock.wav";
+
+    if (bIsError)
+    {
+        sAudioPath = L"Overlay\\acherror.wav";
+    }
+    else if (bSubmit)
+    {
+        const bool bHardcore = pConfiguration.IsFeatureEnabled(ra::services::Feature::Hardcore);
+        const auto nRarity = bHardcore ? pAchievement.rarity_hardcore : pAchievement.rarity;
+
+        if (nRarity >= 10.0)
+        {
+            vmPopup->SetTitle(ra::StringPrintf(L"Achievement Unlocked - %0.2f%%", nRarity));
+        }
+        else if (nRarity > 0.0)
+        {
+            vmPopup->SetTitle(ra::StringPrintf(L"Rare Achievement Unlocked - %0.2f%%", nRarity));
+
+            const wchar_t* sPath = L"Overlay\\rareunlock.wav";
+            const auto& pFileSystem = ra::services::ServiceLocator::Get<ra::services::IFileSystem>();
+            const std::wstring sFullPath = pFileSystem.BaseDirectory() + sPath;
+            if (pFileSystem.GetFileSize(sFullPath) > 0)
+                sAudioPath = sPath;
+        }
+        else
+        {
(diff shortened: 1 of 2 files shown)


Jamiras:
> think it'd be better if it was point tier based instead of rarity based. Doesn't make a ton of sense to get a super epic unlock sound for unlocking a 2 pointer just because it's from a subset, meanwhile there's kaizo hacks that have barely any players so you're getting 100 pointers with 80% earn rate.

This site isn't just for kaizo players. I have nearly 100000 hardcore points, and have yet to earn a single 100 pointer. I do have a few dozen 50 pointers, but 40 / 13000 is approximately 1/3 of 1%, which would make the frequency of seeing the rare achievement too low in my opinion. Also, quite a few of those 50 pointers have a 15% or higher hardcore earn rate, so they're either misscored, or suffer from the same problem you describe where the set is only attracting a specific subset of players.

I'd much rather design this so the average player has the potential to see the popup a few times per game, rather than require they go play something that they'll struggle to complete.

This is quite similar to the Chocolate Donut paradox I use to explain why Retro Ratio is broken. Low earn rates may be due to attrition more than difficulty on popular sets. Rarity is still rarity no matter how you slice it. And, yes, I acknowledge that it would be affected by revisions - but you'd still be amongst the 5% (or whatever) of people who had the achievement.

Also, as this feature is strictly client-side, it's a nice bonus for players that has no real impact on global rankings. A person who cares about their 100 point achievements would be better served with the ability to showcase them for others to see: https://github.com/RetroAchievements/RAWeb/discussions/1207

Blazekickn:
> This site isn't just for kaizo players. I have nearly 100000 hardcore points, and have yet to earn a single 100 pointer. 

I understand that, I'm not saying that the sound should only play when you get a 100 pointer or whatever, but just having it based on rarity alone is gonna result in a lot of sets where 90% of the achievements are considered rare. I think points should play a factor. 

Currently, I'm playing through a shiny and rare drops subset and I have it set to play a loud sound whenever I get an achievement so I actually pay attention to the game that received the shiny and don't reset over it. It'd be nice if I could have the sound play when an achievement is 25 points or more regardless of rarity so the same sound doesn't jumpscare me when I get easy 1-5 point rare drops. 

> I do have a few dozen 50 pointers, but 40 / 13000 is approximately 1/3 of 1%, which would make the frequency of seeing the rare achievement too low in my opinion. 

You could probably have tiers of rarity, like under 25% is uncommon, under 10% is rare, and under 1% is ultra rare or something like that. More customization is nice.

Jamiras:
After years of campaigning to have subsets reflect the player count of the base set to inflate RetroPoints, now you're complaining that the inflated rarity is going to cause all achievements in a subset to be considered rare? If the new sound distracts you, you can always use the same sound for both rare and non-rare achievements.

I agree that RNG subsets make things seem more rare than they should be just because of the current requirement to use a patched ROM. But once players are able to earn RNG subset achievements while working on the main game, those earn rates should normalize. Even at a 1% RNG rate, if a player has a hundred chances to get the achievement in a playthrough, there's a more than 50% chance they'll earn it at some point, which would presumably make the overall earn rate above the 10% threshold.

> just having it based on rarity alone is gonna result in a lot of sets where 90% of the achievements are considered rare.

I think this is highly exaggerated, but there will definitely be some sets where a large portion of the achievements are rare, even though they're not difficult, just do to attrition on the game. Similarly, there will be some difficult achievements that are not rare because the game is only played by people who are good at it.

It's not a perfect system, but a player is more likely to feel rewarded when they do something that very few other people did than when they do something that someone arbitrarily decided was worth a lot of points.

PR #1066, opened 2024-02-05
support for indirect code notes via overflow math

Jamiras:
Instead of masking off the upper bit of GameCube addresses, some clever dev has decided to include the uppermost bit as part of the offset, causing an overflow that results in the uppermost bit being removed.

Normal way to write logic using a masked address:
image

Alternate way to write logic using overflow:
image

This PR updates the achievement editor's tooltip code and memory inspector's code note field to support indirect addresses derived in this manner.


Jamiras:
Also includes a fix for 32-bit BE pointers not being automatically converted to RA addresses (#932).

PR #1083, opened 2024-04-22
Add shift-click feature on memory view to put focus on pointed address

Vancleeff:
Enhanced user experience by providing a convenient way to navigate to specific pointed memory addresses

### DIFF
--- a/src/ui/viewmodels/MemoryViewerViewModel.cpp
+++ b/src/ui/viewmodels/MemoryViewerViewModel.cpp
@@ -4,6 +4,7 @@
 
 #include "data\context\EmulatorContext.hh"
 #include "data\context\GameContext.hh"
+#include "data\context\ConsoleContext.hh"
 
 #include "ui\EditorTheme.hh"
 #include "ui\viewmodels\WindowManager.hh"
@@ -893,6 +894,15 @@ void MemoryViewerViewModel::OnClick(int nX, int nY)
             m_nNeedsRedraw |= REDRAW_MEMORY;
         }
     }
+}
+
+void MemoryViewerViewModel::OnShiftClick(ra::ByteAddress nAddress)
+{
+    const auto& pConsoleContext = ra::services::ServiceLocator::Get<ra::data::context::ConsoleContext>();
+    const auto nConvertedAddress = pConsoleContext.ByteAddressFromRealAddress(nAddress);
+    if (nConvertedAddress != 0xFFFFFFFF)
+        nAddress = nConvertedAddress;
+    SetAddress(nAddress);
 }
 
 void MemoryViewerViewModel::OnResized(int nWidth, int nHeight)
--- a/src/ui/viewmodels/MemoryViewerViewModel.cpp
+++ b/src/ui/viewmodels/MemoryViewerViewModel.cpp
@@ -896,12 +896,18 @@ void MemoryViewerViewModel::OnClick(int nX, int nY)
     }
 }
 
-void MemoryViewerViewModel::OnShiftClick(ra::ByteAddress nAddress)
+void MemoryViewerViewModel::OnShiftClick(int nX, int nY)
 {
+    const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
     const auto& pConsoleContext = ra::services::ServiceLocator::Get<ra::data::context::ConsoleContext>();
+
+    OnClick(nX, nY);
+
+    ra::ByteAddress nAddress = pEmulatorContext.ReadMemory(GetAddress(), GetSize());
     const auto nConvertedAddress = pConsoleContext.ByteAddressFromRealAddress(nAddress);
     if (nConvertedAddress != 0xFFFFFFFF)
         nAddress = nConvertedAddress;
+
     SetAddress(nAddress);
 }
 
--- a/src/ui/viewmodels/MemoryViewerViewModel.cpp
+++ b/src/ui/viewmodels/MemoryViewerViewModel.cpp
@@ -903,12 +903,22 @@ void MemoryViewerViewModel::OnShiftClick(int nX, int nY)
 
     OnClick(nX, nY);
 
-    ra::ByteAddress nAddress = pEmulatorContext.ReadMemory(GetAddress(), GetSize());
+    // We need to create a mask based on the last address of the current console to be able to point to more complexe pointer format
+    // like on Playstation on which pointers are represented as a 24-bit pointer prefixed by 80.
+    std::wstring sMaxAddress = ra::data::MemSizeFormat(pConsoleContext.MaxAddress(), MemSize::ThirtyTwoBit, MemFormat::Hex);
+    std::wstring sMask;
+
+    for (wchar_t ch : sMaxAddress)
+        if (ch != L'0')
+            sMask.push_back(L'F');
+
+    auto nMask = std::stoi(sMask, 0, 16);
+    ra::ByteAddress nAddress = (pEmulatorContext.ReadMemory(GetAddress(), GetSize())) & nMask;
     const auto nConvertedAddress = pConsoleContext.ByteAddressFromRealAddress(nAddress);
     if (nConvertedAddress != 0xFFFFFFFF)
-        nAddress = nConvertedAddress;
-
-    SetAddress(nAddress);
+    {
+        SetAddress(nConvertedAddress);
+    }
 }
 
 void MemoryViewerViewModel::OnResized(int nWidth, int nHeight)
--- a/src/ui/viewmodels/MemoryViewerViewModel.cpp
+++ b/src/ui/viewmodels/MemoryViewerViewModel.cpp
@@ -903,17 +903,7 @@ void MemoryViewerViewModel::OnShiftClick(int nX, int nY)
 
     OnClick(nX, nY);
 
-    // We need to create a mask based on the last address of the current console to be able to point to more complexe pointer format
-    // like on Playstation on which pointers are represented as a 24-bit pointer prefixed by 80.
-    std::wstring sMaxAddress = ra::data::MemSizeFormat(pConsoleContext.MaxAddress(), MemSize::ThirtyTwoBit, MemFormat::Hex);
-    std::wstring sMask;
-
-    for (wchar_t ch : sMaxAddress)
-        if (ch != L'0')
-            sMask.push_back(L'F');
-
-    auto nMask = std::stoi(sMask, 0, 16);
-    ra::ByteAddress nAddress = (pEmulatorContext.ReadMemory(GetAddress(), GetSize())) & nMask;
+    ra::ByteAddress nAddress = (pEmulatorContext.ReadMemory(GetAddress(), GetSize()));
     const auto nConvertedAddress = pConsoleContext.ByteAddressFromRealAddress(nAddress);
     if (nConvertedAddress != 0xFFFFFFFF)
     {
--- a/src/ui/win32/bindings/MemoryViewerControlBinding.cpp
+++ b/src/ui/win32/bindings/MemoryViewerControlBinding.cpp
@@ -16,6 +16,7 @@ constexpr UINT WM_USER_INVALIDATE = WM_USER + 1;
 
 INT_PTR CALLBACK MemoryViewerControlBinding::WndProc(HWND hControl, UINT uMsg, WPARAM wParam, LPARAM lParam)
 {
+    const bool bShiftHeld = (GetKeyState(VK_SHIFT) < 0);
     switch (uMsg)
     {
         case WM_PAINT:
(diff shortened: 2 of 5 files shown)


src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -264,6 +268,15 @@ void MemoryViewerControlBinding::OnClick(POINT point)
     Invalidate();
 }
 
+void MemoryViewerControlBinding::OnShiftClick(POINT point)
+{
+    OnClick(point);
+    const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+    ra::ByteAddress nAddress =
+        pEmulatorContext.ReadMemory(m_pViewModel.GetAddress(), m_pViewModel.GetSize()) & 0xFFFFFF;
+    m_pViewModel.OnShiftClick(nAddress);
```
Jamiras, line 277: You're not shift+clicking `nAddress`, you're just setting the address after converting it. The conversion should be here, then just call `SetAddress`.

Although, for unit testing purposes all of this code should be in the view model. Suggest making `MemoryViewerViewModel::OnShiftClick` accept an x and y, have that call `MemoryViewerViewModel::OnClick`, do the address calculations and call `MemoryViewerViewModel::SetAddress`. Then, please, write unit tests for `MemoryViewerViewModel::OnShiftClick`.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -34,7 +35,10 @@ INT_PTR CALLBACK MemoryViewerControlBinding::WndProc(HWND hControl, UINT uMsg, W
             return FALSE;
 
         case WM_LBUTTONUP:
-            OnClick({ GET_X_LPARAM(lParam) - MEMVIEW_MARGIN, GET_Y_LPARAM(lParam) - MEMVIEW_MARGIN });
+            if (bShiftHeld)
```
Jamiras, line 38: if `bShiftHeld` is only needed for this case, recommend only evaluating it here.
```
if (GetKeyState(VK_SHIFT) < 0)
```

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -264,6 +268,15 @@ void MemoryViewerControlBinding::OnClick(POINT point)
     Invalidate();
 }
 
+void MemoryViewerControlBinding::OnShiftClick(POINT point)
+{
+    OnClick(point);
```
Jamiras, line 273: Note that `OnClick` will abort if the memory viewer has a fixed address (such as when it's used in the pointer finder). This functionality should also be non-functional in the pointer finder.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -264,6 +268,15 @@ void MemoryViewerControlBinding::OnClick(POINT point)
...
+    const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+    ra::ByteAddress nAddress =
+        pEmulatorContext.ReadMemory(m_pViewModel.GetAddress(), m_pViewModel.GetSize()) & 0xFFFFFF;
```
Jamiras, line 276: Why are you masking the value? That should be implicit within `ByteAddressFromRealAddress`

vancleeff:
Thank you for the review @Jamiras, very helpful !
I made some correction and pushed everything, I'm not sure if the tests are enough for this usecase or if I need to add more console context.

tests/ui/viewmodels/MemoryViewerViewModel_Tests.cpp
```diff
@@ -1918,8 +1922,78 @@ TEST_CLASS(MemoryViewerViewModel_Tests)
...
+        viewer.SetAddressFixed(true);
+        viewer.OnShiftClick(10 * CHAR_WIDTH, CHAR_HEIGHT + 4);
+        Assert::AreEqual({0x0U}, viewer.GetAddress());
+
+        // If not fixed address and Shift click on the first word containing 0x20 should lead to address 0x20
+        viewer.SetAddressFixed(false);
+        viewer.OnShiftClick(10 * CHAR_WIDTH, CHAR_HEIGHT + 4);
+        Assert::AreEqual({0x20U}, viewer.GetAddress());
+
+        // Shift click on the second word containing 0x40 should lead to address 0xFF
```
Jamiras, line 1985: Why would 0x40 lead to 0xFF? I think the comment is wrong.

Also, I'd prefer if this second click actually went to a 16-bit address (like 0x0140). The 0xFFFF test case handles the extreme location, but it would be good to validate a 16-bit valid address. You can always call `SetAddress(0)` to reset things for the next click.

src/ui/viewmodels/MemoryViewerViewModel.cpp
```diff
@@ -895,6 +896,21 @@ void MemoryViewerViewModel::OnClick(int nX, int nY)
...
+{
+    const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+    const auto& pConsoleContext = ra::services::ServiceLocator::Get<ra::data::context::ConsoleContext>();
+
+    OnClick(nX, nY);
+
+    ra::ByteAddress nAddress = pEmulatorContext.ReadMemory(GetAddress(), GetSize());
+    const auto nConvertedAddress = pConsoleContext.ByteAddressFromRealAddress(nAddress);
+    if (nConvertedAddress != 0xFFFFFFFF)
+        nAddress = nConvertedAddress;
```
Jamiras, line 909: Please include a test for an inconvertible address.

vancleeff:
Thanks @Jamiras ! I changed the logic a bit by adding a mask for more complex pointer which are not converted into real addresses when using `ByteAddressFromRealAddress`. 

Example : PSX pointer like `0x80012345` are now masked based on the `pConsoleContext.MaxAddress` size and then converted with `ByteAddressFromRealAddress` to point towards `0x12345` address.

I adapted the tests accordingly.

src/ui/viewmodels/MemoryViewerViewModel.cpp
```diff
@@ -895,6 +896,31 @@ void MemoryViewerViewModel::OnClick(int nX, int nY)
...
+    std::wstring sMaxAddress = ra::data::MemSizeFormat(pConsoleContext.MaxAddress(), MemSize::ThirtyTwoBit, MemFormat::Hex);
+    std::wstring sMask;
+
+    for (wchar_t ch : sMaxAddress)
+        if (ch != L'0')
+            sMask.push_back(L'F');
+
+    auto nMask = std::stoi(sMask, 0, 16);
+    ra::ByteAddress nAddress = (pEmulatorContext.ReadMemory(GetAddress(), GetSize())) & nMask;
+    const auto nConvertedAddress = pConsoleContext.ByteAddressFromRealAddress(nAddress);
```
Jamiras, line 917: You completely misunderstood my last comment. What you had was fine. I just wanted a test added for clicking on an address that was not convertible.

You can't convert an address just by masking. It may work for PSX, but it wouldn't work for GBA. The real address "$02001234" converts to a RetroAchievements address of "$00009234".
```
{ 0x008000U, 0x047FFFU, 0x02000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "System RAM" },
```

The conversion is: subtract 0x02000000, then add 0x008000.

The point of calling `ByteAddressFromRealAddress` is to let it do the conversion for you. Just provide the raw data and it'll return an RA Address, or 0xFFFFFFFF. I just wanted you to add a test for a case where it returned 0xFFFFFFFF so we could enshrine the intended behavior. The previous behavior would have just jumped to the max address. The current implementation will not jump at all, which I feel is better.

tests/ui/viewmodels/MemoryViewerViewModel_Tests.cpp
```diff
@@ -1918,8 +1922,120 @@ TEST_CLASS(MemoryViewerViewModel_Tests)
...
+        viewer.OnShiftClick(22 * CHAR_WIDTH, CHAR_HEIGHT + 4);
+        Assert::AreEqual({0x4U}, viewer.GetAddress());
+    }
+
+    TEST_METHOD(TestOnShiftClickThirtyTwoBit)
+    {
+        ra::data::context::mocks::MockConsoleContext mockConsole(PlayStation, L"Playstation");
+
+        MemoryViewerViewModelHarness viewer;
+        viewer.InitializeMemory(0x1FFFFF); // PSX full memory
```
Jamiras, line 1997: Do you really need to allocate all 16MB?

tests/ui/viewmodels/MemoryViewerViewModel_Tests.cpp
```diff
@@ -1918,8 +1922,120 @@ TEST_CLASS(MemoryViewerViewModel_Tests)
...
+        viewer.SetAddress(0);
+        viewer.SetAddressFixed(false);
+        viewer.mockEmulatorContext.WriteMemory(0U, MemSize::ThirtyTwoBit, 0x1234);
+        viewer.OnShiftClick(10 * CHAR_WIDTH, CHAR_HEIGHT + 4);
+        Assert::AreEqual({0x1234U}, viewer.GetAddress());
+
+        // Shift click on the first dword containing 0x80012345 should lead to address 0x00012345 as the conversion keeps
+        // only the same number of nibbles as the Max address from console context (PSX : 1FFFFF)
+        viewer.SetAddress(0);
+        viewer.mockEmulatorContext.WriteMemory(0U, MemSize::ThirtyTwoBit, 0x00012345);
```
Jamiras, line 2019: Your comment says 0x80012345. That's not what you're testing.

src/ui/viewmodels/MemoryViewerViewModel.cpp
Jamiras: I think I see why you're trying to mask now. The `ConsoleContext` is initialized from the rcheevos regions, and the PlayStation regions only define the real memory, not the mirrors:
```
static const rc_memory_region_t _rc_memory_regions_playstation[] = {
    { 0x000000U, 0x00FFFFU, 0x000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "Kernel RAM" },
    { 0x010000U, 0x1FFFFFU, 0x010000U, RC_MEMORY_TYPE_SYSTEM_RAM, "System RAM" }
};
```
There are mirrors at $80000000 and $A0000000 that should also map to $00000000 (real address), which translates to $00000000 (RetroAchievements address).

This logic should be implemented in `ByteAddressFromRealAddress` on a per-console basis. Generic masking is not the correct solution.

Vancleeff: Ok, now I understand how it works, it was a little confusing because of the PSX use-case.
I added these 2 mirrored RAM to consoleinfo.c in rcheevos and now, it works perfectly well on PSX too !
Thanks for your patience @Jamiras !

Vancleeff: To be sure I did it correctly, here's a commit with the change I made to consoleinfo.c : https://github.com/RetroAchievements/rcheevos/commit/59cd2d2a25fa52d57071c1e42a421108c5907366

Jamiras: No, no, no. The regions defined in rcheevos tell emulator developers what memory we expect to be able to find. If you add these mirror regions to the memory map, then they'll become available for writing achievements and appear in search results. See [mirror RAM in NES](https://docs.retroachievements.org/Console-Specific-Tips/#memory-mirroring) and [echo RAM in GameBoy](https://docs.retroachievements.org/Console-Specific-Tips/#memory-mirroring). Not all cores expose the duplicated memory, so achievements relying on it sometimes don't work.

What I meant was something more like this:
```
ra::ByteAddress ConsoleContext::ByteAddressFromRealAddress(ra::ByteAddress nRealAddress) const noexcept
{
    // Map mirrored memory addresses to non-mirrored addresses.
    // Usually, only the non-mirrored addresses appear in m_vRegions.
    switch (m_nId)
    {
        case RC_CONSOLE_PLAYSTATION:
            if ((nRealAddress & 0xFFE00000) == 0x80000000 || // Kernel and User Memory Mirror (2 Meg) Cached
                (nRealAddress & 0xFFE00000) == 0xA0000000)   // Kernel and User Memory Mirror (2 Meg) Uncached
            {
                nRealAddress &= 0x001FFFFF;
            }
            break;

        case RC_CONSOLE_PLAYSTATION_2:
            if ((nRealAddress & 0xFE000000) == 0x20000000 || // Main RAM, uncached
                (nRealAddress & 0xFE000000) == 0x30000000)   // Main RAM, uncached and accelerated
            {
                nRealAddress &= 0x01FFFFFF;
            }
            break;
    }

    for (const auto& pRegion : m_vRegions)
    {
        if (pRegion.RealAddress < nRealAddress)
        {
            const auto nRegionSize = pRegion.EndAddress - pRegion.StartAddress;
            if (nRealAddress < pRegion.RealAddress + nRegionSize)
                return (nRealAddress - pRegion.RealAddress) + pRegion.StartAddress;
        }
    }

    return 0xFFFFFFFF;
}

```

Jamiras:
Are you still working on these? Three of the four PRs have CI failures that need to be addressed, and the fourth has open comments to resolve.

Vancleeff:
Hello @Jamiras !
Yes, I had to finish my second set, it's now on pending review so I will switch back on these PRs !

PR #1084, opened 2024-04-23
Add copy/paste feature to memory view

Vancleeff:
Enhanced user experience by providing a way to copy and paste value directly in the memory view.

### DIFF
--- a/src/ui/win32/bindings/MemoryViewerControlBinding.cpp
+++ b/src/ui/win32/bindings/MemoryViewerControlBinding.cpp
@@ -3,6 +3,7 @@
 #include "ra_fwd.h"
 #include "ra_utility.h"
 
+#include "services\IClipboard.hh"
 #include "ui/EditorTheme.hh"
 #include "ui/drawing/gdi/GDISurface.hh"
 
@@ -231,6 +232,56 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
             }
             return true;
 
+        case 'C':
+            if (bControlHeld)
+            {
+                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+                const auto iValue = pEmulatorContext.ReadMemory(m_pViewModel.GetAddress(), m_pViewModel.GetSize());
+                std::wstring sValue = ra::data::MemSizeFormat(iValue, m_pViewModel.GetSize(), MemFormat::Hex);
+
+                ra::services::ServiceLocator::Get<ra::services::IClipboard>().SetText(ra::Widen(sValue));
+            }
+            return true;
+
+        case 'V':
+            if (bControlHeld)
+            {
+                auto nAddress = m_pViewModel.GetAddress();
+                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+                std::wstring sClipboardText = ra::services::ServiceLocator::Get<ra::services::IClipboard>().GetText();
+                auto n = m_pViewModel.GetSize();
+
+                if (sClipboardText.empty())
+                    return false;
+
+                // Check if the string is a valid hexadecimal value
+                for (wchar_t ch : sClipboardText)
+                    if (!iswxdigit(ch)) return false;
+
+                // Padding zeroes depending if shift is pressed (strict mode) or not (replace mode)
+                if (bShiftHeld)
+                {
+                    const auto nNibblesForSize = ra::data::MemSizeBytes(m_pViewModel.GetSize()) * 2;
+
+                    if (nNibblesForSize < sClipboardText.length())
+                        sClipboardText = sClipboardText.substr(sClipboardText.length() - nNibblesForSize);
+                    else
+                    {
+                        std::wstring sPadding(nNibblesForSize - sClipboardText.length(), L'0');
+                        sClipboardText = (sPadding + sClipboardText);
+                    }
+                }else 
+                    sClipboardText = sClipboardText.length() % 2 == 1 ? (L"0" + sClipboardText) : sClipboardText;
+
+                // Writing every byte separately considerably improves stability and enables long sequences to be pasted
+                for (int i = sClipboardText.length(); i != 0; i-=2)
+                {
+                    std::wstring sValue = sClipboardText.substr(i - 2, 2);
+                    pEmulatorContext.WriteMemoryByte(nAddress++, std::stoi(sValue, 0, 16));
+                }
+            }
+            return true;
+
         default:
             return false;
     }
@@ -308,7 +359,9 @@ void MemoryViewerControlBinding::RenderMemViewer()
 
     const auto& pRenderImage = m_pViewModel.GetRenderImage();
     const auto& pEditorTheme = ra::services::ServiceLocator::Get<ra::ui::EditorTheme>();
-    HBRUSH hBackground = CreateSolidBrush(RGB(pEditorTheme.ColorBackground().Channel.R, pEditorTheme.ColorBackground().Channel.G, pEditorTheme.ColorBackground().Channel.B));
+    HBRUSH hBackground = CreateSolidBrush(RGB(pEditorTheme.ColorBackground().Channel.R,
+                                              pEditorTheme.ColorBackground().Channel.G,
+                                              pEditorTheme.ColorBackground().Channel.B));
 
     // left margin
     RECT rcFill{ rcClient.left, rcClient.top, rcClient.left + MEMVIEW_MARGIN, rcClient.bottom - 1 };
--- a/src/ui/win32/bindings/MemoryViewerControlBinding.cpp
+++ b/src/ui/win32/bindings/MemoryViewerControlBinding.cpp
@@ -249,50 +249,14 @@ bool MemoryViewerControlBinding::HandleShortcut(UINT nChar)
         case 'C':
             if (bControlHeld)
             {
-                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
-                const auto nValue = pEmulatorContext.ReadMemory(m_pViewModel.GetAddress(), m_pViewModel.GetSize());
-                std::wstring sValue = ra::data::MemSizeFormat(nValue, m_pViewModel.GetSize(), MemFormat::Hex);
-
-                ra::services::ServiceLocator::Get<ra::services::IClipboard>().SetText(ra::Widen(sValue));
+                OnCopy();
             }
             return true;
 
         case 'V':
             if (bControlHeld and !m_pViewModel.IsReadOnly())
             {
-                auto nAddress = m_pViewModel.GetAddress();
-                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
-                std::wstring sClipboardText = ra::services::ServiceLocator::Get<ra::services::IClipboard>().GetText();
-                const auto nNibblesForSize = ra::data::MemSizeBytes(m_pViewModel.GetSize());
-                std::vector<std::uint32_t> v_nValues;
-                std::wstring sError;
-
-                if (sClipboardText.empty())
-                    return false;
-
-                // Split the clipboard value into substrings matching the current size and check if they're valid hexadecimal values
-                for (size_t i = 0; i < sClipboardText.size(); i += nNibblesForSize * 2)
-                {
-                    std::wstring sSubString = sClipboardText.substr(i, nNibblesForSize * 2);
-                    unsigned int nValue;
-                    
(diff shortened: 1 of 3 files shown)


src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -231,6 +232,56 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+            }
+            return true;
+
+        case 'V':
+            if (bControlHeld)
+            {
+                auto nAddress = m_pViewModel.GetAddress();
+                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+                std::wstring sClipboardText = ra::services::ServiceLocator::Get<ra::services::IClipboard>().GetText();
+                auto n = m_pViewModel.GetSize();
```
Jamiras, line 252: `n` is not a very descriptive variable name, and I can't tell if it's even being used anywhere.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -231,6 +232,56 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
             }
             return true;
 
+        case 'C':
+            if (bControlHeld)
+            {
+                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+                const auto iValue = pEmulatorContext.ReadMemory(m_pViewModel.GetAddress(), m_pViewModel.GetSize());
```
Jamiras, line 239: please use `nValue` for consistency with the rest of the code.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -231,6 +232,56 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+                std::wstring sValue = ra::data::MemSizeFormat(iValue, m_pViewModel.GetSize(), MemFormat::Hex);
+
+                ra::services::ServiceLocator::Get<ra::services::IClipboard>().SetText(ra::Widen(sValue));
+            }
+            return true;
+
+        case 'V':
+            if (bControlHeld)
```
Jamiras, line 247: Paste shouldn't be allowed in the pointer finder. You'll have to expose `MemoryViewerViewModel::m_bReadOnly` and check that here.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -231,6 +232,56 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+                auto n = m_pViewModel.GetSize();
+
+                if (sClipboardText.empty())
+                    return false;
+
+                // Check if the string is a valid hexadecimal value
+                for (wchar_t ch : sClipboardText)
+                    if (!iswxdigit(ch)) return false;
+
+                // Padding zeroes depending if shift is pressed (strict mode) or not (replace mode)
```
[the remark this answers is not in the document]
vancleeff: The padding was made for the shift mode : in shift mode, the user was able to paste a long hexadecimal sequence and it was pasted and replaced value in memory.
Example : 12345 pasted into these words "1234 0000" resulted in "1234 5000"
 
I changed that in the last commit to be more strict to avoid confusion. 
The shift mode still let the user to paste long sequence but if they are not a multiple of the current size, the residual part will be converted without padding zeroes :
Example : 12345 pasted into these words "1234 0000" result now in "1234 0005"

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -231,6 +232,56 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+                        sClipboardText = sClipboardText.substr(sClipboardText.length() - nNibblesForSize);
+                    else
+                    {
+                        std::wstring sPadding(nNibblesForSize - sClipboardText.length(), L'0');
+                        sClipboardText = (sPadding + sClipboardText);
+                    }
+                }else 
+                    sClipboardText = sClipboardText.length() % 2 == 1 ? (L"0" + sClipboardText) : sClipboardText;
+
+                // Writing every byte separately considerably improves stability and enables long sequences to be pasted
```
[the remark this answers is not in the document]
vancleeff: It was unstable with big nword like "0x80123456" but with the implementation of ParseHex as you suggested, everything is now perfectly fine, thanks ! It was probably due to something wrong with the conversion of my wstring previously.

vancleeff:
Thanks for the review @Jamiras !
I commited these improvements !

src/ui/win32/bindings/ControlBinding.hh
```diff
@@ -5,6 +5,8 @@
 #include "ui\BindingBase.hh"
 #include "ui\win32\DialogBase.hh"
 
+#include "ui\viewmodels\MessageBoxViewModel.hh"
```
Jamiras, line 8: Why was this moved? Please keep unnecessary headers out of header files as they have to be parsed for every file that includes the container header file.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -236,6 +239,68 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+            return true;
+
+        case 'V':
+            if (bControlHeld and !m_pViewModel.IsReadOnly())
+            {
+                auto nAddress = m_pViewModel.GetAddress();
+                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+                std::wstring sClipboardText = ra::services::ServiceLocator::Get<ra::services::IClipboard>().GetText();
+                const auto nNibblesForSize = ra::data::MemSizeBytes(m_pViewModel.GetSize());
+                std::vector<std::uint32_t> v_nValues;
```
Jamiras, line 267: The "x_" prefix defines the scope, and is only for member variables `m_`, statics `s_`, or globals `g_`. This should simply be named `vValues`.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -236,6 +239,68 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+            }
+            return true;
+
+        case 'V':
+            if (bControlHeld and !m_pViewModel.IsReadOnly())
+            {
+                auto nAddress = m_pViewModel.GetAddress();
+                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+                std::wstring sClipboardText = ra::services::ServiceLocator::Get<ra::services::IClipboard>().GetText();
+                const auto nNibblesForSize = ra::data::MemSizeBytes(m_pViewModel.GetSize());
```
Jamiras, line 266: This should be `nBytesForSize`

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -236,6 +239,68 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+                    }
+
+                    v_nValues.push_back(nValue);
+                }
+
+                for (auto nValue : v_nValues)
+                {
+                    //Single mode writes only the first value, multi mode (shift) writes every values
+                    pEmulatorContext.WriteMemory(nAddress, m_pViewModel.GetSize(), nValue);
+                    if (!bShiftHeld)
```
Jamiras, line 292: Recommend moving the shift handling logic into the above loop. No reason to populate the entire array if you're only going to process a single item.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -236,6 +239,68 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+                const auto nValue = pEmulatorContext.ReadMemory(m_pViewModel.GetAddress(), m_pViewModel.GetSize());
+                std::wstring sValue = ra::data::MemSizeFormat(nValue, m_pViewModel.GetSize(), MemFormat::Hex);
+
+                ra::services::ServiceLocator::Get<ra::services::IClipboard>().SetText(ra::Widen(sValue));
+            }
+            return true;
+
+        case 'V':
+            if (bControlHeld and !m_pViewModel.IsReadOnly())
+            {
```
Jamiras, line 262: This logic is sizable enough that I'd split it out into it's own function `OnPaste(bShiftHeld)`. Suggest doing the same for copy as well for consistency.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -236,6 +239,68 @@ bool MemoryViewerControlBinding::HandleNavigation(UINT nChar)
...
+                const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+                std::wstring sClipboardText = ra::services::ServiceLocator::Get<ra::services::IClipboard>().GetText();
+                const auto nNibblesForSize = ra::data::MemSizeBytes(m_pViewModel.GetSize());
+                std::vector<std::uint32_t> v_nValues;
+                std::wstring sError;
+
+                if (sClipboardText.empty())
+                    return false;
+
+                // Split the clipboard value into substrings matching the current size and check if they're valid hexadecimal values
```
Jamiras, line 273: Recommend documenting the shift+paste functionality, as it's something that is not normal and won't be obvious when looking at this code some time in the future.
```
// If the clipboard text is smaller than a single write, it will be treated as a padded value:
//    C => 0C, 000C, or 0000000C (depending on currently selected viewer size)
// If shift is held, and the clipboard string is bigger than a single write, multiple writes will occur. 
// If the last write is not a complete chunk, it will be padded in the same way as described above:
//    12345678C => 12 34 56 78 0C, 1234 5678 000C, or 12345678 0000000C
```

vancleeff:
Thanks @Jamiras, commited !

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -290,6 +319,66 @@ void MemoryViewerControlBinding::OnViewModelIntValueChanged(const IntModelProper
     }
 }
 
+void MemoryViewerControlBinding::OnCopy()
+{
+    const auto& pEmulatorContext = ra::services::ServiceLocator::Get<ra::data::context::EmulatorContext>();
+    const auto nValue = pEmulatorContext.ReadMemory(m_pViewModel.GetAddress(), m_pViewModel.GetSize());
+    std::wstring sValue = ra::data::MemSizeFormat(nValue, m_pViewModel.GetSize(), MemFormat::Hex);
+
+    ra::services::ServiceLocator::Get<ra::services::IClipboard>().SetText(ra::Widen(sValue));
```
Jamiras, line 328: `ra::Widen` is unnecessary. `sValue` is already a `std::wstring`.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -290,6 +319,66 @@ void MemoryViewerControlBinding::OnViewModelIntValueChanged(const IntModelProper
...
+
+        vValues.push_back(nValue);
+    }
+
+    // Multi mode (shift) writes every values
+    for (auto nValue : vValues)
+    {
+        pEmulatorContext.WriteMemory(nAddress, m_pViewModel.GetSize(), nValue);
+        if (!bShiftHeld)
+            break;
```
Jamiras, line 375: This break isn't needed. Processing will be aborted on line 364.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -290,6 +319,66 @@ void MemoryViewerControlBinding::OnViewModelIntValueChanged(const IntModelProper
...
+        }
+
+        // Single mode writes only the first value,
+        if (!bShiftHeld)
+        {
+            pEmulatorContext.WriteMemory(nAddress, m_pViewModel.GetSize(), nValue);
+            return true;
+        }
+
+        vValues.push_back(nValue);
```
Jamiras, line 367: If you don't care about doing a full validation pass before writing, you can avoid using an array entirely and just write as you parse:
```
pEmulatorContext.WriteMemory(nAddress, m_pViewModel.GetSize(), nValue);
if (!bShiftHeld)
    return true;
nAddress += nBytesForSize;
```

Vancleeff:
Fix CI checks

Jamiras:
This has conflicts from the +/- PR.

PR #1086, opened 2024-04-25
Add History to Memory Viewer

Vancleeff:
Enhanced user experience by providing an history on the memory viewer with the last 50 explored addresses (can move backward or forward)

### DIFF
--- a/src/ui/viewmodels/MemoryInspectorViewModel.cpp
+++ b/src/ui/viewmodels/MemoryInspectorViewModel.cpp
@@ -287,6 +287,7 @@ void MemoryInspectorViewModel::OnCurrentAddressChanged(ra::ByteAddress nNewAddre
     SetValue(CurrentAddressValueProperty, nValue);
 
     m_pViewer.SetAddress(nNewAddress);
+    m_pViewer.SaveToMemViewHistory();
 }
 
 void MemoryInspectorViewModel::BookmarkCurrentAddress() const
--- a/src/ui/viewmodels/MemoryInspectorViewModel.cpp
+++ b/src/ui/viewmodels/MemoryInspectorViewModel.cpp
@@ -287,7 +287,6 @@ void MemoryInspectorViewModel::OnCurrentAddressChanged(ra::ByteAddress nNewAddre
     SetValue(CurrentAddressValueProperty, nValue);
 
     m_pViewer.SetAddress(nNewAddress);
-    m_pViewer.SaveToMemViewHistory();
 }
 
 void MemoryInspectorViewModel::BookmarkCurrentAddress() const
--- a/src/ui/viewmodels/MemoryViewerViewModel.cpp
+++ b/src/ui/viewmodels/MemoryViewerViewModel.cpp
@@ -23,11 +23,14 @@ constexpr uint8_t HIGHLIGHTED_COLOR = STALE_COLOR | gsl::narrow_cast<uint8_t>(ra
 constexpr int ADDRESS_COLUMN_WIDTH = 10;
 constexpr int BASE_MEMORY_VIEWER_WIDTH_IN_CHARACTERS =
     (ADDRESS_COLUMN_WIDTH + 16 * 3 - 1); // address column + 16 bytes "XX " - last space
+constexpr int MEMVIEW_HISTORY_MAX_SIZE = 50;
 
 std::unique_ptr<ra::ui::drawing::ISurface> MemoryViewerViewModel::s_pFontSurface;
 std::unique_ptr<ra::ui::drawing::ISurface> MemoryViewerViewModel::s_pFontASCIISurface;
 ra::ui::Size MemoryViewerViewModel::s_szChar;
 int MemoryViewerViewModel::s_nFont = 0;
+std::list<ra::ByteAddress> MemoryViewerViewModel::s_listMemViewHistory = {0};
+std::list<ra::ByteAddress>::iterator MemoryViewerViewModel::s_iterMemViewHistoryIndex = s_listMemViewHistory.begin();
 
 constexpr char FIRST_ASCII_CHAR = 32;
 constexpr char LAST_ASCII_CHAR = 127;
@@ -1312,6 +1315,33 @@ void MemoryViewerViewModel::RenderHeader()
         m_pSurface->WriteText(nX, -2, s_nFont, nColor, sValue);
         nX += s_szChar.Width * (nNibblesPerWord + 1);
     }
+}
+
+void MemoryViewerViewModel::SaveToMemViewHistory()
+{
+    if (*s_iterMemViewHistoryIndex != GetAddress())
+    {
+        s_listMemViewHistory.erase(s_listMemViewHistory.begin(), s_iterMemViewHistoryIndex);
+        if (s_listMemViewHistory.size() > MEMVIEW_HISTORY_MAX_SIZE)
+            s_listMemViewHistory.pop_back();
+        s_listMemViewHistory.push_front(GetAddress());
+        s_iterMemViewHistoryIndex = s_listMemViewHistory.begin();
+    }
+}
+
+void MemoryViewerViewModel::MoveMemViewHistoryForward()
+{
+    if (s_listMemViewHistory.size() >= 1 and s_iterMemViewHistoryIndex != s_listMemViewHistory.begin())
+        --s_iterMemViewHistoryIndex;
+    SetAddress(*s_iterMemViewHistoryIndex);
+}
+
+void MemoryViewerViewModel::MoveMemViewHistoryBackward()
+{
+    int iIndex = std::distance(s_listMemViewHistory.begin(), s_iterMemViewHistoryIndex);
+    if (s_listMemViewHistory.size() > (iIndex + 1))
+        ++s_iterMemViewHistoryIndex;
+    SetAddress(*s_iterMemViewHistoryIndex);
 }
 
 } // namespace viewmodels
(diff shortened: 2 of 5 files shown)


src/ui/viewmodels/MemoryViewerViewModel.hh
```diff
@@ -235,6 +239,8 @@ private:
     static std::unique_ptr<ra::ui::drawing::ISurface> s_pFontSurface;
     static std::unique_ptr<ra::ui::drawing::ISurface> s_pFontASCIISurface;
     static int s_nFont;
+    static std::list<ra::ByteAddress> s_listMemViewHistory;
+    static std::list<ra::ByteAddress>::iterator s_iterMemViewHistoryIndex;
```
Jamiras, line 243: `std::vector` is generally preferred over `std::list` if you're not inserting stuff in the middle. `std::list` uses node pointers to make updates faster at the cost of memory. `ra::ByteAddress` is a 32-bit value, so the overhead of the node is quite expensive.

Also, there's no reason for these to be static. You wouldn't want your history shared across multiple instances - And I'm not really sure how this will interact with the pointer finder. You might be okay since you're managing the addition of items to the history from the inspector.

src/ui/win32/bindings/MemoryViewerControlBinding.cpp
```diff
@@ -57,6 +57,13 @@ INT_PTR CALLBACK MemoryViewerControlBinding::WndProc(HWND hControl, UINT uMsg, W
         case WM_USER_INVALIDATE:
             Invalidate();
             return FALSE;
+
+        case WM_XBUTTONUP:
+            if (GET_XBUTTON_WPARAM(wParam) == XBUTTON1)
+                m_pViewModel.MoveMemViewHistoryBackward();
+            else
```
Jamiras, line 64: Recommend explicitly checking for XBUTTON2 in case other XBUTTONs ever get defined.

src/ui/viewmodels/MemoryViewerViewModel.hh
```diff
@@ -175,6 +175,10 @@ public:
     void AdvanceCursorPage();
     void RetreatCursorPage();
 
+    void SaveToMemViewHistory();
+    void MoveMemViewHistoryBackward();
+    void MoveMemViewHistoryForward();
```
Jamiras, line 180: `MemView` is redundant in these function names as they're part of the `MemoryViewerViewModel`.

src/ui/viewmodels/MemoryViewerViewModel.hh
```diff
@@ -175,6 +175,10 @@ public:
     void AdvanceCursorPage();
     void RetreatCursorPage();
 
+    void SaveToMemViewHistory();
```
Jamiras, line 178: Suggest `AddToHistory`, saving implies something that gets remembered between sessions.

Might also want a `ClearHistory` method so the addresses can be reset when changing games.

src/ui/viewmodels/MemoryViewerViewModel.cpp
```diff
@@ -1314,6 +1317,33 @@ void MemoryViewerViewModel::RenderHeader()
...
 }
 
+void MemoryViewerViewModel::SaveToMemViewHistory()
+{
+    if (*s_iterMemViewHistoryIndex != GetAddress())
+    {
+        s_listMemViewHistory.erase(s_listMemViewHistory.begin(), s_iterMemViewHistoryIndex);
+        if (s_listMemViewHistory.size() > MEMVIEW_HISTORY_MAX_SIZE)
+            s_listMemViewHistory.pop_back();
+        s_listMemViewHistory.push_front(GetAddress());
```
Jamiras, line 1327: Recommend pushing new items on the back to avoid reallocating data until the history is full.

src/ui/viewmodels/MemoryViewerViewModel.cpp
```diff
@@ -1314,6 +1317,33 @@ void MemoryViewerViewModel::RenderHeader()
     }
 }
 
+void MemoryViewerViewModel::SaveToMemViewHistory()
+{
+    if (*s_iterMemViewHistoryIndex != GetAddress())
+    {
+        s_listMemViewHistory.erase(s_listMemViewHistory.begin(), s_iterMemViewHistoryIndex);
+        if (s_listMemViewHistory.size() > MEMVIEW_HISTORY_MAX_SIZE)
+            s_listMemViewHistory.pop_back();
```
Jamiras, line 1326: I think you might have an off-by-one error here. If `size() > MEMVIEW_HISTORY_MAX_SIZE`, that means there's 51 items in the list. Even if you pop the 51st item, you're about to put another 51st back on.

src/ui/viewmodels/MemoryViewerViewModel.hh
Jamiras, line 243: Suggest `s_iterMemViewHistoryIndex` just be an integer. Maintaining a iterator over a dynamically sized array can be difficult. Also, I think it would improve readability.

src/ui/viewmodels/MemoryInspectorViewModel.cpp
```diff
@@ -287,6 +287,7 @@ void MemoryInspectorViewModel::OnCurrentAddressChanged(ra::ByteAddress nNewAddre
     SetValue(CurrentAddressValueProperty, nValue);
 
     m_pViewer.SetAddress(nNewAddress);
+    m_pViewer.SaveToMemViewHistory();
```
Jamiras, line 290: I was originally going to suggest you just capture the new address in `SetAddress`, but putting it here addresses another concern I had about the address being captured when navigating via the keyboard. Recommend having unit tests that ensure the history is not modified by keyboard navigation.

src/ui/viewmodels/MemoryViewerViewModel.hh
vancleeff: I switch to vector and refactor the logic on the last commit. I also tried to switch to integer index instead of iterator but I found it way less readable than the iterator version.

There's probably a better way to do it so I post it that version here and let you judge which version should be implemented and/or what to improve : 

```
void MemoryViewerViewModel::AddToHistory(ra::ByteAddress nAddress)
{
    if (vHistory.at(nHistoryIndex) != nAddress)
    {
        vHistory.erase(std::next(vHistory.begin(), nHistoryIndex + 1), vHistory.end());
        if (vHistory.size() == MEMVIEW_HISTORY_MAX_SIZE)
            vHistory.erase(vHistory.begin());
        vHistory.push_back(nAddress);
        nHistoryIndex = vHistory.size() - 1;
    }
}

void MemoryViewerViewModel::MoveHistoryForward()
{
    int nIndex = std::distance(vHistory.begin(), std::next(vHistory.begin(), nHistoryIndex));
    if (vHistory.size() > (nIndex + 1))
        ++nHistoryIndex;
    SetAddress(vHistory.at(nHistoryIndex));
}

void MemoryViewerViewModel::MoveHistoryBackward()
{
    if (vHistory.size() >= 1 and std::next(vHistory.begin(), nHistoryIndex) != vHistory.begin())
        --nHistoryIndex;
    SetAddress(vHistory.at(nHistoryIndex));
}
```

VS
```
void MemoryViewerViewModel::AddToHistory(ra::ByteAddress nAddress)
{
    if (*iterHistoryIndex != nAddress)
    {
        vHistory.erase(iterHistoryIndex + 1, vHistory.end());
        if (vHistory.size() == MEMVIEW_HISTORY_MAX_SIZE)
            vHistory.erase(vHistory.begin());
        vHistory.push_back(nAddress);
        iterHistoryIndex = vHistory.end() - 1;
    }
}

void MemoryViewerViewModel::MoveHistoryForward()
{
    int nIndex = std::distance(vHistory.begin(), iterHistoryIndex);
    if (vHistory.size() > (nIndex + 1))
        ++iterHistoryIndex;
    SetAddress(*iterHistoryIndex);
}

void MemoryViewerViewModel::MoveHistoryBackward()
{
    if (vHistory.size() >= 1 and iterHistoryIndex != vHistory.begin())
        --iterHistoryIndex;
    SetAddress(*iterHistoryIndex);
}
```

src/ui/viewmodels/MemoryInspectorViewModel.cpp
vancleeff: I moved the AddToHistory directly inside SetAddress() as suggested. 
Concerning capturing the keyboard navigation, I was thinking about that at some point and I found it way more convenient to capture it, it's too confusing if we don't track these movements : 

Let's say we are on 0x0 then we move to 0x123 after which we start to use the keyboard to find something interresting somewhere near 0x123, after some digging we want to come back to 0x123 so we MoveBackward() but we reach 0x0. 
If we want to get back to 0X123 we need to MoveForward() now, which is confusing.

vancleeff:
Committed ! Thanks @Jamiras !

src/ui/viewmodels/MemoryInspectorViewModel.cpp
Jamiras: You don't want to capture keyboard movements. If you jump to address $1234, then press right and down to get to $1245, then type "ABCDEF", you'll be at $1248 and your history will be: [1234->1235->1245->1246->1247->1248]. If I went back in this scenario, I'd expect to jump back to $1234 (the last thing I jumped _to_), not to the last character I typed.

Testing in Visual Studio, it's a little more complicated. It doesn't remember where you jump to, it remembers where you jump from. If you click on row 6, character 8, then row 12, character 10, and press down and type "banana", when you hit back it goes to row 6, character 8. Then forward takes you to the position right after "banana".

src/ui/viewmodels/MemoryViewerViewModel.cpp
```diff
@@ -94,6 +95,8 @@ class MemoryViewerViewModel::MemoryBookmarkMonitor : protected ViewModelCollecti
 private:
     ViewModelCollection<ra::ui::viewmodels::MemoryBookmarksViewModel::MemoryBookmarkViewModel>& m_vBookmarks;
     MemoryViewerViewModel& m_pOwner;
+    std::vector<ra::ByteAddress> vHistory;
+    std::vector<ra::ByteAddress>::iterator iterHistoryIndex;
```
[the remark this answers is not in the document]
Vancleeff: My bad, not needed indeed, thanks !

Jamiras:
This has conflicts from the +/- PR.

PR #1099, opened 2024-05-17
Add Addition and Subtraction Operators

redwizard42:
Adds addition and subtraction as operators for conditions that can modify as support for accumulator features, but available to Add Source, Sub Source, and Add Address.


Jamiras (approved):
Please update the submodule to point at develop now that the dependent code has been merged.

redwizard42:
> Please update the submodule to point at develop now that the dependent code has been merged.
@Jamiras I believe this is done now. The hash when I did the init from remote matches the last commit on develop.

Jamiras:
There's a conflict on the rcheevos hash that should be resolved by merging master and keeping the newer develop tag.

PR #1101, opened 2024-05-21
Add Remember and Recall Support

redwizard42:
Depends on https://github.com/RetroAchievements/rcheevos/pull/343

Adds support for Remember condition type and Recall Operand Type to Asset Editor.


src/ui/viewmodels/TriggerConditionViewModel.cpp
```diff
@@ -219,6 +222,15 @@ void TriggerConditionViewModel::SerializeAppendOperand(std::string& sBuffer, Tri
             sBuffer.push_back('~');
             break;
 
+        case TriggerOperandType::Recall:
+        {
+            std::string recall = "{recall}";
+            for (size_t i = 0; i < recall.length(); i++) {
+                sBuffer.push_back(recall.at(i));
```
redwizard42, line 229: Appending using this method for better clarity when reading the code, but perhaps it would be better to push the individual characters?

Jamiras: Just use append:
```
sBuffer.append("{recall}");
```

src/ui/win32/AssetEditorDialog.cpp
```diff
@@ -302,7 +356,7 @@ class TargetValueColumnBinding : public ValueColumnBinding
 protected:
     bool IsHidden(const ra::ui::ViewModelCollectionBase& vmItems, gsl::index nIndex) const
     {
-        return !vmItems.GetItemValue(nIndex, TriggerConditionViewModel::HasTargetProperty);
+        return !vmItems.GetItemValue(nIndex, TriggerConditionViewModel::HasTargetProperty) || !vmItems.GetItemValue(nIndex, TriggerConditionViewModel::HasTargetValueProperty);
```
Jamiras, line 359: `HasTargetValueProperty` should be added to `DependsOn`

src/ui/viewmodels/TriggerViewModel.cpp
```diff
@@ -35,6 +35,7 @@ TriggerViewModel::TriggerViewModel() noexcept
     m_vConditionTypes.Add(ra::etoi(TriggerConditionType::SubSource), L"Sub Source");
     m_vConditionTypes.Add(ra::etoi(TriggerConditionType::AddHits), L"Add Hits");
     m_vConditionTypes.Add(ra::etoi(TriggerConditionType::SubHits), L"Sub Hits");
+    m_vConditionTypes.Add(ra::etoi(TriggerConditionType::Remember), L"Remember");
```
Jamiras, line 38: I think `Remember` should be lower on the list - maybe between `Or Next` and `Measured`, or all the way down after `Trigger.

src/ui/win32/AssetEditorDialog.cpp
```diff
@@ -252,13 +258,61 @@ class ValueColumnBinding : public ra::ui::win32::bindings::GridTextColumnBinding
...
+    {
+        if (IsHidden(vmItems, nIndex))
+            return L"";
+
+        return ValueColumnBinding::GetText(vmItems, nIndex);
+    }
+
+    bool DependsOn(const ra::ui::BoolModelProperty& pProperty) const override
+    {
+        if (pProperty == TriggerConditionViewModel::HasTargetProperty)
```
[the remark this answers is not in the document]
redwizard42: You are correct and also highlighting the dangers of copy/paste.

src/ui/win32/AssetEditorDialog.cpp
```diff
@@ -159,6 +159,12 @@ class ValueColumnBinding : public ra::ui::win32::bindings::GridTextColumnBinding
                 return true;
             }
 
+            case ra::ui::viewmodels::TriggerOperandType::Recall:
+            {
+                vmItems.SetItemValue(nIndex, *m_pBoundProperty, L"");
+                return true;
+            }
```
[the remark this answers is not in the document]
redwizard42: I mostly mimicked the behavior of when the Target Value ended up hidden by the HasTargetProperty - the cell still allows attempted entries.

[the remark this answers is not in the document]
redwizard42: Overrode CreateInPlaceEditor; as a consequence the target value should no longer accept text when the whole target is hidden either.

src/ui/win32/AssetEditorDialog.cpp
```diff
@@ -252,13 +258,61 @@ class ValueColumnBinding : public ra::ui::win32::bindings::GridTextColumnBinding
...
+        return ValueColumnBinding::GetTooltip(vmItems, nIndex);
+    }
+
+    bool HandleRightClick(const ra::ui::ViewModelCollectionBase& vmItems, gsl::index nIndex) override
+    {
+        if (IsHidden(vmItems, nIndex))
+            return false;
+
+        return ValueColumnBinding::HandleRightClick(vmItems, nIndex);
+    }
```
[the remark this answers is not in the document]
redwizard42: I felt ambitious.

src/ui/win32/AssetEditorDialog.cpp
```diff
@@ -138,6 +138,14 @@ class ValueColumnBinding : public ra::ui::win32::bindings::GridTextColumnBinding
 
     void SetMaximum(unsigned int nValue) noexcept { m_nMaximum = nValue; }
 
+    std::wstring GetText(const ra::ui::ViewModelCollectionBase& vmItems, gsl::index nIndex) const override
+    {
+        if (IsHidden(vmItems, nIndex))
+            return L"";
+
+        return ValueColumnBinding::GetText(vmItems, nIndex);
```
Jamiras, line 146: `ValueColumnBinding` is no longer a base class. This causes a stack overflow.

src/ui/win32/AssetEditorDialog.cpp
```diff
@@ -252,58 +272,96 @@ class ValueColumnBinding : public ra::ui::win32::bindings::GridTextColumnBinding
...
     {
-        if (pProperty == TriggerConditionViewModel::HasTargetProperty)
-            return true;
+        auto* pGridBinding = static_cast<ra::ui::win32::bindings::GridBinding*>(pInfo.pGridBinding);
+        Expects(pGridBinding != nullptr);
+        if (!pGridBinding->GetItems().GetItemValue(pInfo.nItemIndex, TriggerConditionViewModel::HasSourceValueProperty))
+            return nullptr;
 
-        return GridColumnBinding::DependsOn(pProperty);
+        m_nEditingItemIndex = pInfo.nItemIndex;
```
Jamiras, line 312: `m_nEditingItemIndex` doesn't appear to be used anywhere.

src/ui/win32/AssetEditorDialog.cpp
```diff
@@ -252,58 +272,96 @@ class ValueColumnBinding : public ra::ui::win32::bindings::GridTextColumnBinding
...
+    HWND CreateInPlaceEditor(HWND hParent, InPlaceEditorInfo& pInfo) override
+    {
+        auto* pGridBinding = static_cast<ra::ui::win32::bindings::GridBinding*>(pInfo.pGridBinding);
+        Expects(pGridBinding != nullptr);
+        if (!pGridBinding->GetItems().GetItemValue(pInfo.nItemIndex, TriggerConditionViewModel::HasTargetProperty))
+            return nullptr;
+        if (!pGridBinding->GetItems().GetItemValue(pInfo.nItemIndex, TriggerConditionViewModel::HasTargetValueProperty))
+            return nullptr;
+
+        m_nEditingItemIndex = pInfo.nItemIndex;
```
Jamiras, line 353: `m_nEditingItemIndex` doesn't appear to be used anywhere.

src/ui/win32/AssetEditorDialog.cpp
redwizard42: I apparently tested with a test-case that only had hidden values on the source operand, so now I understand how I didn't catch this.

Jamiras (approved):

PR #1112, opened 2024-07-26
Fix bug where modifying conditions improperly hide the target value cell.

redwizard42:
Selecting a modifying condition like Add Source, the Target Value will be hidden when you first change the operator away form 'None', until you change the Target Type.  This was especially noticeable when pasting in something like Add Source Mem 0x1234 / Mem 0x1234, but also evident if you just created a new condition and immediately changed the flag to Add Source, followed by setting the operator.

OnValueChanged handler is updated to re-evaluate the HasTargetValuePropertly when the operator is changed.

### DIFF
--- a/src/ui/viewmodels/TriggerConditionViewModel.cpp
+++ b/src/ui/viewmodels/TriggerConditionViewModel.cpp
@@ -513,9 +513,7 @@ void TriggerConditionViewModel::OnValueChanged(const IntModelProperty::ChangeArg
     }
     else if (args.Property == OperatorProperty)
     {
-        const auto nCurrentType = ra::itoe<TriggerOperandType>(GetValue(TargetTypeProperty));
         SetValue(HasTargetProperty, ra::itoe<TriggerOperatorType>(args.tNewValue) != TriggerOperatorType::None);
-        SetValue(HasTargetValueProperty, !IsParameterlessType(nCurrentType) && GetValue(HasTargetProperty));
         UpdateHasHits();
     }
     else if (args.Property == TypeProperty)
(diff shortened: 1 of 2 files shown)


src/ui/viewmodels/TriggerConditionViewModel.cpp
```diff
@@ -513,7 +513,9 @@ void TriggerConditionViewModel::OnValueChanged(const IntModelProperty::ChangeArg
     }
     else if (args.Property == OperatorProperty)
     {
+        const auto nCurrentType = ra::itoe<TriggerOperandType>(GetValue(TargetTypeProperty));
         SetValue(HasTargetProperty, ra::itoe<TriggerOperatorType>(args.tNewValue) != TriggerOperatorType::None);
+        SetValue(HasTargetValueProperty, !IsParameterlessType(nCurrentType) && GetValue(HasTargetProperty));
```
Jamiras, line 518: The `HasTargetSize` visibility flag is being updated when the `HasTargetProperty` changed. This logic should probably be moved there too:
```
    if (args.Property == HasTargetProperty)
    {
        const auto nTargetType = GetTargetType();
        SetValue(HasTargetSizeProperty, args.tNewValue && IsAddressType(nTargetType));
        SetValue(HasTargetValueProperty, args.tNewValue && !IsParameterlessType(nTargetType));
    }
```

redwizard42: Yes, that's a better location.

src/ui/viewmodels/TriggerConditionViewModel.cpp
```diff
@@ -581,7 +581,11 @@ void TriggerConditionViewModel::UpdateHasHits()
 void TriggerConditionViewModel::OnValueChanged(const BoolModelProperty::ChangeArgs& args)
 {
     if (args.Property == HasTargetProperty)
+    {
+        const auto nTargetType = GetTargetType();
         SetValue(HasTargetSizeProperty, args.tNewValue && IsAddressType(GetTargetType()));
+        SetValue(HasTargetValueProperty, args.tNewValue && !IsParameterlessType(nTargetType));
```
Jamiras, line 587: Line 586 should use the variable too.

PR #1116, opened 2024-08-17
add Pause button to memory bookmarks window

Jamiras:
To make it more obvious that the functionality even exists. I'm not really happy with the label, as it implies behavior very similar to freezing, but the button isn't large enough to fit the entire "Pause on Change" text, and it's consistent with the label shown in the column when the address is flagged to pause on change. "Pause on Change" would also be consistent with the "Pause on Reset" and "Pause on Trigger" options in the achievement editor.

image


wescopeland:
@Jamiras Is "Pause" kind of like setting a breakpoint with debugging tools? Maybe "Break" better conveys what it does. I'm not sure what the limitations of the UI are, but it may help to add an explanatory tooltip on the buttons (if one isn't already present).

I wonder if changing the "Freeze" label might also help, perhaps to something like "Lock"?

Jamiras:
Pause and Freeze are existing behaviors. Pause is currently only accessible by clicking on the field in the grid:
image

This is trying to make the feature more accessible to users. Most users don't even know you can click on the field in the grid. I originally didn't want to add more buttons to the bottom bar because I had plans to add several more things to the dropdown, but it's been several years now, and that still hasn't happened.

The freeze button has been present since the dialog was rewritten in 2020.

Jamiras:
> Is "Pause" kind of like setting a breakpoint with debugging tools? Maybe "Break" better conveys what it does.

Yes. It behaves like a debugger breakpoint. When the value changes, the emulator is paused, and a dialog pops up saying why.
image

My concerns against naming it "break" have to do with the userbase not being familiar with code debuggers. To a non-programmer, "break" usually means "make it not work". I asked for ideas in #devs-chat, and that came up there too, so it's the best alternative currently available.

wescopeland:
Got it. A few other alternatives I can think of:
* "Halt on Chg"
* "Monitor"

These are still a bit distinct from "Freeze", concise, and hopefully don't convey to non-programmers that we're breaking something.

No strong feelings on any of the alternatives, just want to add more ideas to the pool.

Jamiras:
I discounted "Monitor" and "Watch" as that's the general behavior provided by the bookmarks window. I'd go with "Pause on Chg" over "Halt on Chg" for consistency, but I don't like arbitrarily removing letters to make text fit on a widget.

redwizard42:
Is. It possible to just make the cell wider? (I forget the column header... Is it Behavior?)

Could have it do the default is
"Monitor Only"
With "Pause on Chang" and "Freeze Value" and make it a rally explicit at the expense of making the horizontal real estate a bit wider by default 

Having a non-blank default would make it more obvious you can do something with the cell too, though this statement isn't intended as opposition to a button to set behavior.

PR #1129, opened 2024-11-15
Support for multi-level indirect code notes in address tooltips

Jamiras:
Given this code note:
image

This tooltip will now appear:
image

Similarly, for a much deeper code note:
image

This tooltip will appear (note that the address is off because there's a NULL in the pointer chain)
image

For this to work, each pointer must be annotated as a pointer (i.e. `[32-bit pointer]`). Items under a pointer may have any prefix (typically `+`, `.`, or `-`, but more complex things are allowed. The code will find the first "+digit" on the next line and anything preceding that will be considered the prefix. All following lines starting with the prefix are assumed to be nested notes for the pointer.

i.e.
```
  Pointer [32bit]
  +0x428 | Data Pointer (32bit)
  --- +0x24C | Flag
  +0x438 | Graphics Pointer (32bit)
  --- +0x08 | X
  --- +0x0C | Y
```

Note: I've updated the code to use hex offsets even when less than 10 (so the final result differs slightly from the above screenshots).

### DIFF
--- a/src/data/models/CodeNoteModel.cpp
+++ b/src/data/models/CodeNoteModel.cpp
@@ -17,6 +17,7 @@ struct CodeNoteModel::PointerData
     uint32_t RawPointerValue = 0xFFFFFFFF;       // last raw value of pointer captured
     ra::ByteAddress PointerAddress = 0xFFFFFFFF; // raw pointer value converted to RA address
     unsigned int OffsetRange = 0;                // highest offset captured within pointer block
+    bool HasPointers = false;
 
     enum OffsetType
     {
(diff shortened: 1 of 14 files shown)


redwizard42:
Wondering it it'd be reasonable to add parent lines to the note text for clarity.  Such as if you have a pointer to party members which then have pointers to their skills... That way you don't have to note which party member it is in the second layer.

[32-Bit Pointer] Party Members
+0x08 [32-Bit Pointer] Party Member 1
++0x04 Sword Skill [16-Bit]

Party Member 1, Sword Skill [16-bit]

Maybe as an eventual improvement

redwizard42:
> For this to work, each pointer must be annotated as a pointer (i.e. `[32-bit pointer]`). Items under a pointer may have any prefix (typically `+`, `.`, or `-`, but more complex things are allowed. The code will find the first "+digit" on the next line and anything preceding that will be considered the prefix. All following lines starting with the prefix are assumed to be nested notes for the pointer.
> 
> i.e.
> 
> ```
>   Pointer [32bit]
>   +0x428 | Data Pointer (32bit)
>   --- +0x24C | Flag
>   +0x438 | Graphics Pointer (32bit)
>   --- +0x08 | X
>   --- +0x0C | Y
> ```
> 
> Note: I've updated the code to use hex offsets even when less than 10 (so the final result differs slightly from the above screenshots).

I have often used -- to indicate specific values at an offset (example from SaGa 3 being worked on right this moment):
```
Entity 1 (Dune, usually) Stats [24-Bit Pointer] - $0x14b384 holds table count
+0x04=Dune - Max HP << 12 [32-Bit] (Writeable)
Divide this by 4096 (0x1000) to get actual value.
12 Binary Decimals Fixed Point
+0x08=Dune - STR << 12 [32-Bit] (Writeable)
Divide this by 4096 (0x1000) to get actual value.
12 Binary Decimals Fixed Point
+0x0c=Dune - MAG << 12 [32-Bit] (Writeable)
Divide this by 4096 (0x1000) to get actual value.
12 Binary Decimals Fixed Point
+0x10=Dune - AGI << 12 [32-Bit] (Writeable)
Divide this by 4096 (0x1000) to get actual value.
12 Binary Decimals Fixed Point
+0x14=Dune - DEF << 12 [32-Bit] (Writeable)
Divide this by 4096 (0x1000) to get actual value.
12 Binary Decimals Fixed Point
+0xc4=Dune - Status [8-Bit]
-- Bit1=Alive/Conscious
-- Bit4=Dead/K.O.
-- Bit5=Poison
-- Bit6=Sleep
-- Bit7=Paralyze?
+0xc5=Dune - Status [8-Bit]
-- Bit0=Confused?
-- Bit1=Blind
-- Bit2=Curse?
-- Bit3=Stone
-- Bit5=Silence
+0x118=Dune - Current HP << 12 [32-Bit] (Writeable)
Divide this by 4096 (0x1000) to get actual value.
12 Binary Decimals Fixed Point
+0x128=Dune Equipment [24-Bit Pointer]
++0x04=Weapon 1 ID [32-Bit]
-- 0x15=Longsword
++0x0c=Weapon 1 Durability [24-Bit]
-- Bits0-11=Max Durability
-- Bits12-23=Current Durability
```
Will this treat those as nested offsets or will they be treated like part of the note for the prior offset?

Jamiras:
As long as a `+#` is not seen in a line, it won't see it as a nested offset.

>  The code will find the first "+digit" on the next line and anything preceding that will be considered the prefix. 

So:
```
--- +0x1234
++0x1234
.+0x1234
 +0x1234
-+0x1234
indent +0x1234
```

would all represent an indented offset following a line containing "pointer" (though not of the same pointer because they differ from each other). This prefix is captured, and each following line that matches the prefix (up through the `+` of the `+0`) would be treated as sibling notes of the parent pointer.

I know the first three exist in the database. The others are made up and may or may not exist (though I'm pretty confident the last doesn't).

Jamiras:
> Wondering it it'd be reasonable to add parent lines to the note text for clarity. 

I think we'd have to have the entire chain. Sometimes the parent itself isn't enough. Using my example above:
image

"Puzzle 5" is more significant to the value's meaning than "Pointer to Move Digit Array".

While I think it would be reasonable to show the entire note chain (similar to the note structure itself):
```
0x00000004 (indirect 0x00168580)
+0x08 Puzzle Pointer [32-bit]
 +0x5c Move Texture Pointers [32-bit]
  +0x1268 Puzzle 5 - Move Pointer for 1st Digit [32-bit]
   +0x10 Pointer to Move Digit Array [32-bit]
    +0x04
Digit [8-bit]
```
I'm not certain that's any clearer, or how well it conveys the fact that it's a leaf node of the chain, and not the tree structure of a root node.

The individual notes could be garnered by looking at the tooltip of items earlier in the chain.

This is the first step for a couple of other things I want to try to do - like adding a pointer chain value to the bookmark window. In that case, you could just edit the description to add any nuance needed to clarify the value.

PR #1205, opened 2025-07-20
add helper function to modify non-visible virtualized items

Jamiras:
Fixes first part of https://discord.com/channels/310192285306454017/1149693430306447380/1395985705422815283

The multi-edit functionality was only writing to the visible rows.


Jamiras:
Also addresses the second point of the link above. When the virtual list was empty, the ScrollMaximum property wasn't being set to 0 so the list box thought it still had however many items were in the previously selected group.

PR #1274, opened 2026-01-03
move ra::data::context::ConsoleContext to ra::context::IConsoleContext

Jamiras:


src/devkit/data/MemoryRegion.hh
```diff
@@ -0,0 +1,140 @@
...
+        return (nAddress >= m_nStartAddress && nAddress <= m_nEndAddress);
+    }
+
+    /// <summary>
+    /// Gets the type of the region.
+    /// </summary>
+    const Type GetType() const noexcept { return m_nType; }
+
+    /// <summary>
+    /// Gets the type of the region.
```
wescopeland: Typo: should be "Sets".

PR #1276, opened 2026-01-04
Add ARM64 platform support and define RAPIDJSON_ENDIAN for Windows ARM64

jdavidson89:
This PR adds ARM64 platform support for RAIntegration.

Changes included:
- Added ARM64 configuration to the solution and project files
- Updated the VC toolset for the ARM64 configuration to the latest(v145)
- Added RAPIDJSON_ENDIAN=RAPIDJSON_LITTLEENDIAN to fix endianness detection
- Verified successful build on Windows ARM64
Note: This PR does not modify Win32 or x64 configurations. All changes are isolated to the new ARM64 platform.


jdavidson89:
The build is successful and executes correctly when applying my changes to the source tree, when building my fork the build is 1 kb smaller and doesn't have proper versioning info so builds successfully but fails to execute.

src/RA_Integration.vcxproj
```diff
@@ -39,7 +51,7 @@
     <RootNamespace>ra</RootNamespace>
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
```
Jamiras: Just a few lines up from here, please update the logic to set the TargetName to `RA_Integration-ARM64`.

jdavidson89: I will check these changes out when I next get a chance to, I can also revert to v143 that was simply as that was the toolset I had installed I will reupload and yes I have working builds of rasnes9x and rapplewin and am working at porting more over as well, just submitting so you can add a working architecture to your codebase if you so wish to but I will clean it up accordingly.  Thank you for your feedback.

PR #1334, opened 2026-08-20
Support multi-level pointers in pointer finder

Jamiras:
Algorithm will search up to 16 levels deep with offsets up to 8192 bytes (+0x2000). These limits are currently hard-coded.

image


CySlaytor:
I'm very glad to see something like that implemented in the traditional pointer finder.

PR #1348, opened 2026-09-12
feat: add _RA_InstallSearchMemoryBankReader for sparse memory exclusion

Alasonga:
## Problem

Emulators with sparse/lazy memory allocation register large memory banks that are mostly unallocated at runtime. `CaptureMemory` has no way to distinguish an unallocated block from an allocated one — it calls `readBlock` and includes the block regardless.

## Solution

Adds `_RA_InstallSearchMemoryBankReader` — a dedicated per-bank block reader for `CaptureMemory`. Returning `0` excludes the block from the search. The regular block reader remains unchanged for Memory Inspector display.

The search reader takes priority over `_RA_InstallMemoryBankBlockReader` during `CaptureMemory`, complementing #1314 for cases where the block reader cannot return `0` without side effects on other read paths. Emulators that don't register a search reader are unaffected.

## Validation — RPCS3 (PS3), 6 games

Memory map: 256 MB contiguous (Bank 0) + 1024 MB lazy (Banks 1–3). Without a search reader: 1280 MB per search.

| Game | Lazy allocated | Total captured |
|------|----------------|----------------|
| Dark Souls BLES01765 | 186 MB | 442 MB |
| Demon's Souls BLUS30443 | 170 MB | 426 MB |
| NieR BLUS30481 | 129 MB | 385 MB |
| Alice: Madness Returns BLUS30607 | 121 MB | 377 MB |
| Resident Evil 5 BLUS30491 | 178 MB | 434 MB |
| God of War III BCUS98111 | 170 MB | 426 MB |


Alasonga:
Note: AI tools were used to assist in development. All code was reviewed, tested, and verified manually before submission.

Jamiras (commented):
I'm not fond of these changes. They're very specific to an integration _not using_ `rc_client`, which is the complete opposite direction than I would like to see this taken. All future integrations should be using `rc_client` for their normal runtime, and that should be extended with `rc_client_raintegration` for the development tools.

As I [mentioned in the other PR](https://github.com/RetroAchievements/rcheevos/pull/545#issuecomment-5559470290), I think the existing block reader needs to be allowed to return two different "no memory was read" codes. One indicating the address was not recognized, and the other indicating the address is not in use.

>  I wonder if we should allow peek_memory to return 0xFFFFFFFF to indicate the address is valid but not yet available?

If you think this is a better solution, how do you propose [hooking it into the `rc_client_raintegration` framework](https://github.com/RetroAchievements/rcheevos/pull/545#issuecomment-5635975714)?

> We shouldn't rely on this behavior. When the DLL is loaded through rc_client via [rc_client_begin_load_raintegration](https://github.com/RetroAchievements/rcheevos/wiki/rc_client_begin_load_raintegration), generic ReadBlock handlers are registered that wrap the read_memory callback.

[I also said](https://github.com/RetroAchievements/rcheevos/pull/545#issuecomment-5635975714) this wasn't something that we needed to address immediately, and it only took you 29 hours to open a PR completely ignoring the notes that I provided.

Alasonga:
Fair point on the timing. The reason I moved quickly is that the RPCS3 integration is functionally complete. All six points from Wes' roadmap are done:
1. Toolkit support: `rc_client_raintegration` integrated
2. Console ID: PR #535 merged
3. Memory map: PR #545 merged
4. 64-bit pointers: low 32-bit workaround confirmed working with achievements firing across sessions 
5. Address stability: validated cross-session, video proof available
6. Hashing: PR #536 merged

RPCS3's PR #19143 is in Draft pending UX polish and a few remaining features (hardcore guards, mainly), not core functionality. The search space was the last technical issue I wanted to resolve before moving it out of Draft.

Now, to answer your question: the hook point is the `load_callback.` `ResetMemory()` creates the `ReadMemoryBlockN` wrappers automatically but never sets `readBlockSearch`. So after the reset, the emulator calls `_RA_InstallSearchMemoryBankReader` in the same `load_callback` where it already calls `_RA_InstallMemoryBank`.
This is also why the PR isn't specific to non-rc_client integrations, RPCS3 uses `rc_client_begin_load_raintegration`. 

On the sentinel: I looked into `0xFFFFFFFF` before going this route. The problem is that `CaptureMemory` only excludes a block when `nRead == 0`. With `0xFFFFFFFF`, the block gets included and the original problem stays. Additionally, the Memory Inspector path accumulates `nBytesRead` as `uint32_t`, so `+= 0xFFFFFFFF` overflows.

The deeper issue is that lazy PS3 pages need opposite behavior depending on context: Memory Inspector and `rc_client` need zeros, `CaptureMemory` needs to skip the block. One return value can't express both.

